PR #38, opened 2014-09-16
Modified setup.py and control/__init__.py to match numpy.

### LINKED ISSUE #37: Can't install control in Travis-CI
I have a package that depends on `control` but I have trouble testing it in Travis-CI due to the following error when running `pip install control`:

> ImportError: Unable to import git_revision. Try removing control/version.py and the build directory before building.

The full log is here: https://travis-ci.org/kdavies4/ModelicaRes/jobs/35359894.

Do you know how to fix this? It seems that `control.version` can't be imported until the `control` package is installed, which requires `control.version`. A catch-22?

jgoppert:
I tried to avoid using the _CONTROL_SETUP_ switch like numpy
but this was causing some issues for Travis CI. This means for
setup.py test to work we can't do control import * so I had to
change to explicit imports for discrete_test.py. Closes #37.

### DIFF
--- a/control/tests/discrete_test.py
+++ b/control/tests/discrete_test.py
@@ -5,7 +5,15 @@
 
 import unittest
 import numpy as np
-from control import *
+from control.statesp import StateSpace
+from control import matlab
+from control.xferfcn import TransferFunction
+from control.lti import isdtime, timebase, isctime, timebaseEqual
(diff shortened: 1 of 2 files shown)


cwrowley:
Now that setup.py has changed (and #37 fixed), I believe this is superfluous, so closing this PR.

Issue #33, opened 2014-08-12
Upload docs to pythonhosted / pypi


jgoppert:
Pypi now has a place to host docs for python packages. It would be good to upload our docs there.

You can now host documentation at http://pythonhosted.org/control. To upload documentation, prepare a .zip file that is unpacked into this URL. Only static pages are supported. The zip file must have a top-level "index.html".

cwrowley:
Version 0.7.0 is now available at http://python-control.readthedocs.org

Issue #29, opened 2014-08-12
Runtests fails


cwrowley:
At least on my Mac, `python runtests.py` fails, with the messages below:

    Building, see build.log...
    running install
    Checking .pth file support in /Users/clancy/Work/Projects/pycontrol/build/testenv/lib/python2.7/site-packages/
    /usr/local/opt/python/bin/python2.7 -E -c pass
    TEST FAILED: /Users/clancy/Work/Projects/pycontrol/build/testenv/lib/python2.7/site-packages/ does NOT support .pth files
    error: bad install directory or PYTHONPATH
    
    You are attempting to install a package to a directory that is not
    on PYTHONPATH and which Python does not read ".pth" files from.  The
    installation directory you specified (via --install-dir, --prefix, or
    the distutils default setting) was:
    
        /Users/clancy/Work/Projects/pycontrol/build/testenv/lib/python2.7/site-packages/
    
    and your PYTHONPATH environment variable currently contains:
    
        '/Users/clancy/Work/Projects/pycontrol'

This happens for me with python 2.7.8 and 3.4.1.  If I run `python runtests.py -n` to skip the build, then the tests run fine.  So it seems like there might be a problem either with runtests.py or setup.py.  But if I understand correctly from @jgoppert, these files were taken directly from numpy, and if I clone the numpy source and do `python runtests.py` in that, it works fine.  So this is mysterious to me.

PR #41, opened 2014-12-01
Improve docs

cwrowley:
There were some problems with the way the sphinx documentation was configured, in particular the path setting in `doc/conf.py`.  The changes in this pull request fix this, and create "mocks" for numpy and scipy routines so that these packages do not need to be installed for the documentation to build (necessary on readthedocs.org, and a good thing in general).

This branch is based on the 0.6d release, and builds successfully on readthedocs.org (but not on travis.ci, because this release predates that setup).  My plan is to merge this in to master, and then improve the documentation further.


cwrowley:
Finally getting back to this.  I suggest we no longer support Python 3.2, and instead support 3.3 and 3.4 (and 2.7, and possibly 2.6).  That's what Anaconda supports now, and that is probably the main distribution people will be using.  (It's certainly what I suggest to people.)  This has the added benefit of being able to use the conda installer for Travis builds, which is much faster and cleaner than installing everything from source like we do now.

slivingston:
I support the recommendation to not support Python 3.2. Trying to support v3.2 in addition to the other versions of Python listed in [the comment above](https://github.com/python-control/python-control/pull/41#issuecomment-83588705) causes complications with respect to denoting Unicode strings (cf. [PEP 414](https://www.python.org/dev/peps/pep-0414/)).

@cwrowley, I can review and merge this pull request when you are ready. I understand this to be a pull request that addresses issue #33.

cwrowley:
@slivingston, thanks--this is ready to go as far as I am concerned.  Yes, this addresses issue #33 and #29.

murrayrm:
I merged the lsim improvements from crowley/lsim.  This pull request should be ready for merging into master.  Just waiting on Travis CI to build and complete its checks.

murrayrm:
Errors from Travis CI.  @cwrowley, can you take a look?

slivingston:
Note that the build failure may be spurious. GitHub has been under a big DDoS attack. https://status.github.com

slivingston:
I restarted [the build](https://travis-ci.org/python-control/python-control/builds/56316096), and it succeeded.

Issue #19, opened 2014-08-09
Add Project to the Python Package Index [sf#3]


murrayrm:
**Reported by eike on 2011-06-12 13:43 UTC**
The Python Control project should be added to the Python Package Index. 
http://pypi.python.org/pypi

This would ease some administrative work:

* Uploading packages can be easily automated. 
* Each package has a documentation website. Uploading the the generated website can be automated too.
* Users can install the library with ``easy_install``. A user would only need to type ``easy_install python-control`` to download and install the package. 
* It would also give the project additional visibility.

An introduction how to automate build and upload tasks, in conjunction with the Python Package Index, is here:
http://packages.python.org/an_example_pypi_project/

cwrowley:
I just uploaded 0.7.0 to PyPI (and binstar).

PR #52, opened 2015-04-03
Add conda build recipe and improve version numbering

cwrowley:
This pull request adds a "recipe" for building a conda package, and also improves how versions are specified for releases.  Previously, one needed to specify the version number of each release manually in setup.py, and this has led to some confusion, as the version number did not necessarily correspond to tags in the git repository.  With this pull request, version numbers are specified simply by pushing a tag to the git repo, as
```
git tag -a 0.6.6
```
The version numbers are then generated by a script `make_version.py`, which is run automatically by the conda recipe (but not by `setup.py`, so one still needs to remember to run `make_version.py` before uploading to PyPI).  The version numbers are consistent with [PEP 440](https://www.python.org/dev/peps/pep-0440/), so for instance, a version that is 6 commits after the version tagged as 0.6.6 would be version "0.6.6.post6".

The benefits of this approach are that the version numbers are specified only once (as a tag in the git repo), and they are guaranteed to correspond to the git repository.  This avoids some confusion, as happened with v0.6.6 that is currently on PyPI (which I believe was built from @jgoppert's fork and includes PR #38, which has not been merged in, and is probably superfluous now).

This PR also changes the Travis CI build to test the conda recipe, and additionally to install slycot from a binary package I uploaded to binstar.org.  This dramatically speeds up the travis build, which was spending most of its time compiling SLICOT from source.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@ build/
 dist/
 .ropeproject/
 MANIFEST
-control/version.py
+control/_version.py
 build.log
 *.egg-info/
 .coverage
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@ dist/
 .ropeproject/
 MANIFEST
 control/_version.py
+__conda_*.txt
+record.txt
 build.log
 *.egg-info/
 .coverage
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ python:
 before_install:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
-  - sudo apt-get update --fix-missing -qq
-  - sudo apt-get install gfortran liblapack-dev
   # use miniconda to install numpy/scipy, to avoid lengthy build from source
   - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
       wget http://repo.continuum.io/miniconda/Miniconda-3.4.2-Linux-x86_64.sh -O miniconda.sh;
@@ -22,14 +20,15 @@ before_install:
   - conda config --set always_yes yes --set changeps1 no
   - conda update -q conda
   - conda install --yes python=$TRAVIS_PYTHON_VERSION conda-build pip coverage
+  - conda config --add channels http://conda.binstar.org/cwrowley
   - conda info -a
 
 # Install packages
 install:
   - conda build conda-recipe
   - conda install control --use-local
+  - conda install slycot
   - pip install coveralls
-  - pip install slycot
 
 # command to run tests
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ before_install:
   - conda config --set always_yes yes --set changeps1 no
   - conda update -q conda
   - conda info -a
+  - python make_version.py
 
 # Install packages
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,12 +21,13 @@ before_install:
   - hash -r
   - conda config --set always_yes yes --set changeps1 no
   - conda update -q conda
+  - conda install --yes python=$TRAVIS_PYTHON_VERSION conda-build pip coverage
   - conda info -a
-  - python make_version.py
 
 # Install packages
 install:
-  - conda install --yes python=$TRAVIS_PYTHON_VERSION coverage nose numpy scipy matplotlib pip
+  - conda build conda-recipe
+  - conda install control --use-local
   - pip install coveralls
   - pip install slycot
 
--- a/README.rst
+++ b/README.rst
@@ -1,65 +1,96 @@
-Python Control System Library
-=============================
-
 .. image:: https://travis-ci.org/python-control/python-control.svg?branch=master
-    :target: https://travis-ci.org/python-control/python-control
+   :target: https://travis-ci.org/python-control/python-control
 .. image:: https://coveralls.io/repos/python-control/python-control/badge.png
-        :target: https://coveralls.io/r/python-control/python-control
+   :target: https://coveralls.io/r/python-control/python-control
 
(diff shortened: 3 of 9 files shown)


slivingston:
If `control/_version.py` is missing (or otherwise if an `IOError` exception is raised while trying to read from it), then the version string is "dev", which does not conform [with PEP 440](https://www.python.org/dev/peps/pep-0440/#public-version-identifiers).

slivingston:
My comment above applies also if `__version__` fails to be found when executing `control/_version.py`. Cf. [line 7 of setup.py](https://github.com/cwrowley/python-control/blob/5ff196c860909d18707ccdd2f85a89a4b531150c/setup.py#L7).

cwrowley:
That's correct.  And in that case, setup.py gives a UserWarning that 'dev' is an invalid version.  This seems like a reasonable fallback to me, but do you have another suggestion?

slivingston:
Another concern: counting commits since a tag will fail in the case of branching. E.g., a version string of "0.6.6.post6" could occur for two distinct commits that have a common ancestor. While you make the commit hash available via `control.__commit__`, why not include it in the version string?

slivingston:
My basic argument against not including some--possibly incomplete--version number in the source tree is that Git becomes a required dependency of `python-control`. E.g., a snapshot of the repo created via `git archive` will not have any version specifier (except possibly the name of the directory, depending on how `git archive` is called), and it will not be possible to install from that snapshot. This especially matters because GitHub provides a Web service like `git archive`, e.g., https://github.com/python-control/python-control/releases

murrayrm:
While we sort out the version discussion, is it worth separating out the two aspects of this request and re-issuing a separate pull request for all but the versioning part?  Just a thought.

cwrowley:
@slivingston, the main reason I did things the way I did them is that I just copied what was done in binstar!  I agree about the ambiguity of branches, but there shouldn't actually be any branching in the main releases, so that does not bother me.
About the dependency on git, I don't think it's quite accurate to say that Git becomes a required dependency of python-control.  You need git if you want to *release* a version (e.g., upload to PyPI or conda).  But if you are just *installing* with pip or conda, you don't need git, because the version is baked into these distributions.  So the dependency only affects those who are doing the releasing, or are doing something very nonstandard like installing from a git archive.  Also, note that the current setup.py (which was copied from numpy) does try to call `git rev-parse HEAD`, so if anything this change makes us less dependent on Git.
So I agree, it's not perfect, but I think the convenience and fool-proofness of just needing to specify the version once outweigh these drawbacks, which seem to only affect obscure situations.  We do have an actual situation that has come up (namely, that the version currently on pypi does not match any version in the repository), and this solution would fix that and make sure things like that do not happen again.

@murrayrm, I suppose it would be possible to separate this into two pull requests, but I would have to change the conda configuration, and I'd rather not go to extra work to replace that with something that, in my opinion, does not work as well.

slivingston:
As I recall, the original mistake concerning the version on PyPI occurred because someone [uploaded a release](https://github.com/python-control/python-control/issues/19#issuecomment-51960300) without discussing it, from [a fork of the main repo](https://github.com/jgoppert/python-control). Indeed, examine the version number in [setup.py](https://github.com/jgoppert/python-control/blob/38d5c95fd4de6251fc83179fdfc2245874ebcb7e/setup.py) of that fork and compare with [the one on PyPI](https://pypi.python.org/pypi/control/0.6.6). Such could happen again, whether or not the version number is explicitly in the source tree.

While `setup.py` from NumPy will try to call `git`, it will append to the version number ".dev0+Unknown" if that fails. I am not sure that it is appropriately described as being more or less dependent on Git than the changset of this PR.

Nonetheless, I think my concerns about dependence on Git and duplicate version strings for distinct (non-release) commits can be addressed empirically. That is, perhaps we should just merge this PR, and try it for awhile.

By the way, besides `conda` and `pip`, another major venue for getting python-control releases is <https://sourceforge.net/projects/python-control/files/>.

cwrowley:
> While setup.py from NumPy will try to call git, it will append to the version number ".dev0+Unknown" if that fails. I am not sure that it is appropriately described as being more or less dependent on Git than the changset of this PR.

All I meant was that in the new version, `setup.py` does not call git at all.  The only time we call git is when distributing (running `make_version.py`).

Note that the numpy stuff has been unreliable for us (e.g., #37), and the fix using the `__NUMPY_SETUP__` global variable is even called "ugly" and "hackish" in the numpy source.  The binstar approach just seems simpler and cleaner to me, since it clearly separates the "pre-build" (`make_version.py`) from the actual build (`setup.py`), instead of trying to do both of these things in `setup.py`.

> That is, perhaps we should just merge this PR, and try it for awhile.

Great.  So unless anybody has any objections, I'll plan to do the following on Monday:
- Merge this PR (unless somebody else wants to)
- Tag the new version as 0.6.7
- Upload it to PyPI and conda/binstar

I'm not one of the developers on sourceforge, but I can send Richard or Scott the tarball or you could grab it from PyPI and upload to sourceforge.

murrayrm:
All sounds good to me.  When we do a new release on SourceForge, we typically send out an e-mail to python-control-announce describing what is new in the release.  Scott and I can handle that as well, with input from others if we need help in the descriptions.  The e-mail for the last release is here:

http://sourceforge.net/p/python-control/mailman/message/32135649/

cwrowley:
Looking back, I see version 0.6a was released in Nov 2012, more than 2 years ago.  There have been a lot of changes since then (including changing the version numbering scheme from 0.6d, etc, to 0.6.5, etc), so I'm thinking it makes sense to bump the version to 0.7.0.  I'll plan on releasing this as 0.7.0 tomorrow, unless there are any objections.

murrayrm:
Sounds sensible.

PR #54, opened 2015-04-05
Fix #45 - make gain vector optional in root_locus

### LINKED ISSUE #45: rlocus should allow kvect=None as an argument
From Roberto Bucher:
> "rlocus" creates automatically a vector with the k values, if not provided. "root_locus" should have the possibility to pass kvect=None,  but it is not accepted (-> error). It can be simply solved by adding the same lines as in rlocus, before finding a better algorithm to get the k values
>
>    #! TODO: update with a smart calculation of the gains using sys poles/zeros
>    if klist == None:
>        klist = logspace(-3, 3)

cwrowley:
As a consequence of this change, `root_locus` now returns two items, the roots and the gains used (consistent with `matlab.rlocus`).

This PR also fixes some mistakes I found in the documentation.

### DIFF
--- a/control/matlab.py
+++ b/control/matlab.py
@@ -1509,7 +1509,7 @@ def ssdata(sys):
     return (ss.A, ss.B, ss.C, ss.D)
 
 # Return transfer function data as a tuple
-def tfdata(sys, **kw):
+def tfdata(sys):
     '''
     Return transfer function data objects for a system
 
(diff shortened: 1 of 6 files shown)


control/matlab.py
```diff
@@ -1525,17 +1518,12 @@ def tfdata(sys, **kw):
...
-    --------
-    inputs = int; outputs = int
-        For MIMO transfer function, return num, den for given inputs, outputs
-
     Returns
     -------
     (num, den): numerator and denominator arrays
         Transfer function coefficients (SISO only)
     '''
-    tf = _convertToTransferFunction(sys, **kw)
```
cwrowley: The keyword arguments `inputs` and `outputs` to `_convertToTransferFunction` apply only if `sys` is a SISO system, and the result is to return a MIMO transfer function with the specified number of inputs and outputs.  So this is different from what is in the documentation string above, and since this routine just extracts the numerator and denominator of a transfer function, there would never be a need to convert a SISO system to MIMO.

Issue #50, opened 2015-03-20
Interactive plots in IPython notebooks


cwrowley:
I find IPython notebooks a great way to work interactively, and it would be nice if inline plots could be interactive (e.g., letting you zoom, or having tool tips show more information).  With matplotlib, though, inline plots are just images, so they cannot be interactive.

There's a new plotting library called [Bokeh](http://bokeh.pydata.org), an open-source project led by Continuum Analytics (the wonderful folks that make [Anaconda](http://continuum.io/downloads)), that looks really nice for making interactive plots like this.  I'd like to try making an alternate backend one could specify with, say, `control.set_backend("bokeh")` that would make commands like `bode()` and `nyquist()` generate plots using Bokeh.  The default could still be matplotlib, of course.

Any thoughts or suggestions for other libraries that might be useful for making interactive plots?

slivingston:
Outside of IPython notebooks, you can zoom-in, pan, etc. in Matplotlib plots. I just shared a link via issue #49 to [matplotlib.backend_bases.MouseEvent](http://matplotlib.org/api/backend_bases_api.html#matplotlib.backend_bases.MouseEvent) for responding to clicks on points in these plots.

Perhaps there can be two major interactive use-cases: with and without IPython. The former case could be treated with Bokeh, and the latter case with Matplotlib and various GUI hooks that we create.

blinkminster:
Clancy,

I use the following code in my ipython notebooks and it makes the plots
interactive (zooming, panning work; this is in the standard Anaconda
distribution on OSX 10.10):

import matplotlib as mlt
mlt.use('nbagg') # makes plots interactive for zooming.

I haven't tested it with python-control yet but I don't see why there would
be a problem.

- Sawyer

------------------------
Sawyer B. Fuller, Ph.D.
Harvard School of Engineering and Applied Science
Wyss Institute for Biologically Inspired Engineering
http://people.seas.harvard.edu/~minster/

cwrowley:
@blinkminster, that is pretty slick!  I never new about nbagg.  (Incidentally, this also apparently works with the IPython magic `%matplotlib nbagg`.)  I still might give the Bokeh library a try, but this is great to know about.

ilayn:
I am also wondering about this issue for my library and I have been also trying to find a good basis for this issue. So far I've been experimenting with Bokeh. Things are still wonky as an experiment but I just wanted to also get your opinion about the following prototype

[...]

Issue #49, opened 2015-03-20
Bode, Nyquist, and root locus plots are crude


cwrowley:
Bode, Nyquist, and root locus plots are all much nicer in Matlab, both because they make smarter choices for things like frequencies to plot, and because their plots are interactive, so you can click on points to get additional information, etc.
I have some ideas for how to improve this (e.g., more sophisticated ways of generating frequencies for Bode/Nyquist plots, with finer resolution near lightly damped poles or zeros; better gains for root locus plots, with finer resolution near break-out and break-in points, etc), and I'm planning to try implementing those.

slivingston:
Regarding making plots interactive, you could use events from Matplotlib to respond to clicks, e.g., using [matplotlib.backend_bases.MouseEvent](http://matplotlib.org/api/backend_bases_api.html#matplotlib.backend_bases.MouseEvent).

roryyorke:
My 2c: don't forget about Nichols.  It might be worth separating the frequency response and root-locus into separate issues.

For the frequency plots, one could also consider showing low and high frequency asymptotes.
 
One idea I had for this to have the number of sample points be a multiple of the number of poles and zeros, though that is a bit simplistic.  Another idea is to relate the density of the sample points to the curvature of the graph (though I don't know how to turn that into an algorithm...)

repagh:
one thing we also need are damping-omega grids for root-locus. 

for root-locus, having the break-in and break-out points in the gain vector already helps,

Sup3rGeo:
Concerning the damping-omega grids as @repagh said, I have written a simple function that draws the grid for Z-domain (zgrid):

https://gist.github.com/Sup3rGeo/17d0eb2b0f4d60c1ebd24695ac692b5d

Is there any possibility for this to be added to the python-control functions? I have never contributed to open source projects before, but if you tell me if what to do I can perhaps create a pull request for this.

Sup3rGeo:
Created a PR for this (#193)

don4get:
Can we have more of what might be improved from what is already present in the library?

Thanks to all the improvements you done since the creation of this issue, I can't say if it's closed or not.

rxa254:
For the Nyquist plot, I would like to see:

1. Smoother lines. This can be done by sampling the system better in regions where |d(Mag)/df| is large.
1. Better scaling (not sure how to do this). With systems that have a lot of DC gain or if there is a deep bandstop filter, its difficult to see the detail just due to the dynamic range.
[...]

PR #69, opened 2015-08-18
bode-plot

marthoch:
* enhanced the default choice of the frequency range and made it more configurable
* added a manual way to select the frequency range
* discrete systems are now supported and bode-plot delivers an equivalent plot as for continuous systems
* for discrete systems the Nyquist frequency is annotated by a vertical line
* gird for phase set to 45/15deg 
* added an IPython notebook in examples/bode-plot.ipynb to test and show off the bode-plot function.

issue #49: some notes are implemented
issue #50: considered in example


control/freqplot.py
```diff
@@ -60,7 +58,7 @@
 #
 
 # Bode plot
-def bode_plot(syslist, omega=None, dB=None, Hz=None, deg=None,
+def bode_plot(syslist, omega=None, omega_num=None, dB=None, Hz=None, deg=None,
```
roryyorke: if backward compatibility is important, `omega_num` should go just before `*args`

control/freqplot.py
```diff
@@ -72,6 +70,8 @@ def bode_plot(syslist, omega=None, dB=None, Hz=None, deg=None,
         List of linear input/output systems (single system is OK)
     omega : freq_range
         Range of frequencies (list or bounds) in rad/sec
+    omega_num: int
+        number of sample
```
roryyorke: samples (I think)

control/freqplot.py
```diff
@@ -117,65 +117,100 @@ def bode_plot(syslist, omega=None, dB=None, Hz=None, deg=None,
     # If argument was a singleton, turn it into a list
     if (not getattr(syslist, '__iter__', False)):
         syslist = (syslist,)
-
-    mags, phases, omegas = [], [], []
+    
+    if omega is None:
+        # Select a default range if none is provided
+        omega = default_frequency_range(syslist, Hz=Hz)
+    elif (isinstance(omega, tuple) or isinstance(omega, list)) and len(omega) == 2:
```
roryyorke: do you want to also support numpy.ndarray for omega here?  isinstance takes a tuple as second arg, so this could be `isinstance(omega, (tuple,list,np.ndarray))`.  There doesn't seem to be a good general way to check for a sequence (see, e.g., http://stackoverflow.com/a/2937122/1008142 )

control/freqplot.py
```diff
@@ -117,65 +117,100 @@ def bode_plot(syslist, omega=None, dB=None, Hz=None, deg=None,
     # If argument was a singleton, turn it into a list
     if (not getattr(syslist, '__iter__', False)):
         syslist = (syslist,)
-
-    mags, phases, omegas = [], [], []
+    
+    if omega is None:
+        # Select a default range if none is provided
```
roryyorke: Suggest using omega_num in this case, or raising an error if omega_nums is not None

control/freqplot.py
```diff
@@ -357,6 +391,8 @@ def default_frequency_range(syslist):
     -------
     omega : array
         Range of frequencies in rad/sec
+    number_of_samples: int
+        Number of samples to generate
```
roryyorke: args Hz, feature_periphery_decade not documented

control/freqplot.py
```diff
@@ -380,26 +424,55 @@ def default_frequency_range(syslist):
...
+                freq_interesting.append(fn*0.9)
+                p = sys.pole()
+                p = p[p != -1.]
+                features = np.concatenate((features, np.abs(np.log(p) / sys.dt)))
+                z = sys.zero()
+                z = z[(z.imag != 0.0)]  # Get rid of poles and zeros at the origin and real <= 0 & imag==0
+                features = np.concatenate((features, np.abs(np.log(z) / sys.dt)))                
+            else:
+                pass
+                # TODO:
```
roryyorke: warn or raise error here?

control/freqplot.py
```diff
@@ -380,26 +424,55 @@ def default_frequency_range(syslist):
...
         try:
             # Add new features to the list
-            features = np.concatenate((features, np.abs(sys.pole())))
-            features = np.concatenate((features, np.abs(sys.zero())))
+            if sys.isctime():
+                features_ = np.abs(sys.pole())
+                features_ = np.concatenate((features_, np.abs(sys.zero())))                
+                # Get rid of poles and zeros at the origin
+                features_ = features_[features_ != 0.0];
+                features = np.concatenate((features, features_))
```
roryyorke: Would it make sense to remove all poles and zeros on (or very near) the imaginary axis?  It wouldn't make sense to remove such zeros for a Nyquist plot, I guess.

control/freqplot.py
```diff
@@ -380,26 +424,55 @@ def default_frequency_range(syslist):
...
+            elif sys.isdtime(strict=True):
+                fn = np.pi * 1. / sys.dt
+                #TODO: What distance to the Nyquist frequency is appropriate?                
+                freq_interesting.append(fn*0.9)
+                p = sys.pole()
+                p = p[p != -1.]
+                features = np.concatenate((features, np.abs(np.log(p) / sys.dt)))
```
roryyorke: p != -1 lets through poles at 0, which will give an inf in the np.log(p) call.  It also doesn't correspond to the comment below ("Get rid of poles and zeros ...")

control/freqplot.py
```diff
@@ -380,26 +424,55 @@ def default_frequency_range(syslist):
...
+                z = sys.zero()
+                z = z[(z.imag != 0.0)]  # Get rid of poles and zeros at the origin and real <= 0 & imag==0
```
roryyorke: the zero selection here doesn't correspond to the comment

control/freqplot.py
martinup: Thanks for all the valuable comments. Great to get some feedback.

That is how I interpreted the original documentation but the feature was not implemented.
The more I think about the less I like it.
I will implement an additional argument `omega_limits`. That should be more clear.

I agree the frequency-setup stuff should be refactored.

murrayrm:
Looked through changes and these look good to merge.

PR #70, opened 2015-10-17
Signed-off-by: whiplash01 <amirhasandex@gmail.com>

whiplash01:
Updated the gm and wg values

Faulty values were failing the tests, This commit fixes it.

### DIFF
--- a/control/tests/matlab_test.py
+++ b/control/tests/matlab_test.py
@@ -606,9 +606,9 @@ class TestMatlab(unittest.TestCase):
 
         gm, pm, wg, wp = margin(Hol)
         # print("%f %f %f %f" % (gm, pm, wg, wp))
-        self.assertAlmostEqual(gm, 3.32065569155)
+        self.assertAlmostEqual(gm, 1.0263157090822087e-15)
         self.assertAlmostEqual(pm, 46.9740430224)
-        self.assertAlmostEqual(wg, 0.176469728448)
+        self.assertAlmostEqual(wg, 1.6586487848042873e-09)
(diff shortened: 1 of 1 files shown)


roryyorke:
I don't think the patch is right; testCombi01 passes fine on master 49045ca [1], and that's on Python 2.7, 3.3, and 3.4.

The numbers you've used look odd; a GM of 1e-15 at a frequency of 1e-9 seems implausible.

It might be there's a bug lurking in the margin code.  What OS, Python, Numpy, and slycot version did you use to test this?  (Can't recall offhand if slycot code is used in the margin calcs---probably not).

[1] https://travis-ci.org/python-control/python-control/builds/90145794

whiplash01:
@roryyorke Thanks for responding. When I was installing python-control, I was having this error on Gm and Wg values which on terminal appeared as this should be this. I replaced the values and all the tests ran fine. It worked. I am not a control engineering expert, so I can't determine whether the values I have selected are even in the ballpark. Therefore, I concluded it was a bug. 

OS - Ubuntu 14.04 LTS
Python - 2.7
NumPy - 1.8.2
Slycot - Can't determine.

Some help would be appreciated.

roryyorke:
unfortunately I'll be offline for about a week.

I am using Ubuntu 14.04 and Python 2.7 as well.

I can only give general advice now: use virtualenv, try miniconda, and look
at the .travis.xml file for hints on how the dependencies are installed.

whiplash01:
Thanks, I'll check it.

PR #72, opened 2015-11-04
Fix autosummary tables on readthedocs

### LINKED ISSUE #71: Documentation on readthedocs.org does not list functions
In the function reference on http://python-control.readthedocs.org/en/latest/control.html, the functions are not actually listed.  Just the headings show up.

cwrowley:
Fix #71

This PR changes the configuration file used by readthedocs, to use the github version of sphinx instead of the latest release (1.3.1).  This fixes the problem with the autosummary tables not displaying on readthedocs.org.

There was also a problem with the travis build failing, because of a version mismatch in the numpy used to build slycot vs. python-control.  This is also fixed, simply by installing slycot before python-control in the travis build.

There are also a few lines added to the travis script to use their new "container-based infrastructure" and attempt to cache things installed by pip.  Ideally, this should speed up the travis build, but it seems about the same.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,13 @@
+sudo: false
 language: python
+
+cache:
+  apt: true
+  pip: true
+  directories:
+    - $HOME/.cache/pip
+    - $HOME/.local
+
 python:
   - "2.7"
   - "3.3"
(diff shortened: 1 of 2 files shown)


murrayrm:
Reviewed changes and they look good.

PR #78, opened 2016-02-29
First commit of balred2() added to modelsimp.py.

mdclemen:
This pull request contains a function, balred2(), which will do a balanced, truncated realization of an unstable LTI model.

### DIFF
--- a/control/modelsimp.py
+++ b/control/modelsimp.py
@@ -197,6 +197,37 @@ def modred(sys, ELIM, method='matchdc'):
     rsys = StateSpace(Ar,Br,Cr,Dr)
     return rsys
 
+def stabsep(T_schur, Z_schur, sys, ldim, no_in, no_out):
+    """
+    Performs stable/unstabe decomposition of sys after Schur forms have been computed for system matrix.
+
+    Reference: Hsu,C.S., and Hou,D., 1991,
+    Reducing unstable linear control systems via real Schur transformation.
+    Electronics Letters, 27, 984-986.
+
+    """
+    #Author: M. Clement (mdclemen@eng.ucsd.edu) 2016
+    As = np.asmatrix(T_schur)
+    Bs = Z_schur.T*sys.B
+    Cs = sys.C*Z_schur
+    #from ref 1 eq(1) As = [A_ Ac], Bs = [B_], and Cs = [C_ C+]; _ denotes stable subsystem
+    #                      [0  A+]       [B+]
+    A_ = As[0:ldim,0:ldim]
+    Ac = As[0:ldim,ldim::]
+    Ap = As[ldim::,ldim::]
+    
+    B_ = Bs[0:ldim,:]
+    Bp = Bs[ldim::,:]
+    
+    C_ = Cs[:,0:ldim]
+    Cp = Cs[:,ldim::]
+    #do some more tricky math IAW ref 1 eq(3)
+    B_tilde = np.bmat([[B_, Ac]])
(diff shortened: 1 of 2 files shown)


cwrowley:
Looks promising.  I haven't looked at this closely, but I have a couple of remarks at first glance:
* Looks like the Travis tests are failing
* Before this is merged in, it will be important to add unit tests for the new routine
* I personally think it's a bad idea to have two routines `balred()` and `balred2()` with similar functionality and different interfaces.  I'd prefer to see a single routine, for instance with an optional argument to specify which algorithm to use.

roryyorke:
The test failure is the same thing as reported in pr#70; exactly the same gain margins were found by @whiplash01 as Travis produced for Python 2.7 and 3.4 here.

I can't reproduce this on my machine (Ubuntu 14.04, Python 3.4.3) with "own-built" Slycot and python-control.  I have struggled to reproduce the Travis environment with my relatively slow network connection; Conda seems to want to download an awful lot of things, and then change it's mind about Python versions, etc.

Anyway, it looks like there's a subtle bug lurking in the gain margin code.

mdclemen:
I have modified balred() to now just check for whether or not unstable eigenvalues exist, and if so, perform the necessary steps to do the balancing and truncating.

mdclemen:
My first to commits were failing on testBalred(), when orders was passed as a vector(list). I've fixed this error and now the balred() function will return a vector(list) of reduced order models if orders is passed as a vector. I don't know what is going on with the testCombi01().

murrayrm:
Travis build is failing due to older version of conda (fixed in f334b1d3bd69d7036d3e3b80c4a79e33a11099d8).  This will be corrected when code is merged into master.

murrayrm:
@mdclemen: Need unit tests before this can be merged into master.  Also, if you can grab the update to .travis.yml from f334b1d3bd69d7036d3e3b80c4a79e33a11099d8, you can get Travis CI build working.

mdclemen:
Could you please explain what unit tests are or direct me to a reference.

murrayrm:
A good introduction is here:

  https://en.wikipedia.org/wiki/Unit_testing

If you look in control/tests, you'll see a large collection of unit tests for different functions.  If you copy one of those into something called `balred_test.py` and then put in some valid unit tests, we can use that to make sure that everything is working under all of the different platforms.  `margin_test.py` is a reasonably good sample.

cwrowley:
Since the modified routine is in `modelsimp.py`, my suggestion would be to put the new tests in `tests/modelsimp_test.py`, following the [NumPy/SciPy Testing Guidelines](https://github.com/numpy/numpy/blob/master/doc/TESTS.rst.txt).

Also, if you rebase your branch to the current master branch, then the tests will be run automatically by Travis CI when you push more commits to this pull request.  The current version in this PR has a bug in setting things up on Travis CI, which leads to the error showing up here.

(By the way, @murrayrm, it might be good at some point to organize the existing tests better, to follow the NumPy/SciPy guidelines, for instance so that tests for module `foo.py` are in `test_foo.py`.  Many are already like this, but some modules do not have corresponding test modules.)

slivingston:
@mdclemen, will you consider opening a new pull request or reworking commits on this one (and re-opening it)? I can work with you to make unit tests.

mdclemen:
@slivingston, I have made updates to my fork of python-control and it requires the updates I made to slycot. I will wait to get my pull for [python-control/slycot](https://github.com/python-control/Slycot/pull/1) merged first.

PR #89, opened 2016-05-05
step_response returns individual x var response

KronosKoderS:
Added the ability to return individual x variable responses.


KronosKoderS:
I tried building and testing on my machine at home, but ran into issues (I'm assuming this was due to Travis-CI specific stuff).  I see now that there are some test issues.  I'll work on these later and get them working.

Issue #84, opened 2016-04-19
Calculate poles more smarter for state-space systems


cwrowley:
As pointed out by @pmli in #27, the [current implementation](https://github.com/python-control/python-control/blob/0b26e062a6cd82da8a2420ebc3957e5267e6e116/control/statesp.py#L408) involves unnecessarily computing the characteristic polynomial, and then finding its roots.  This is inefficient and numerically unstable.  Much better to use `np.linalg.eigvals`.

murrayrm:
Addressed in PR #91, now merged into master.

Issue #85, opened 2016-04-19
Avoid using inv when not necessary


cwrowley:
As pointed out by @pmli in #27, instead of [computing matrix inverses](https://github.com/python-control/python-control/blob/0b26e062a6cd82da8a2420ebc3957e5267e6e116/control/statesp.py#L458), better to use `np.linalg.solve` when needed.

murrayrm:
Addressed in PR #91, now merged into master.

PR #91, opened 2016-05-20
Addressing #84 and #85

mp4096:
* Compute poles of a state-space system using `numpy.linalg.eigvals` as suggested in #84 
* Replaced matrix inversion in the feedback computation by `numpy.linalg.solve` as suggested in #85 
* Also replaced the singularity check with a more stable one. [`numpy.linalg.matrix_rank`](http://docs.scipy.org/doc/numpy-dev/reference/generated/numpy.linalg.matrix_rank.html) uses a reasonable numerical threshold, so that we don't need to check `abs(det(F))` by hand. Determinant is actually very sensitive to matrix scaling, so better not use it. Both the determinant and the SVD required for `matrix_rank` are cubic, so there should be no slowdown.

### DIFF
--- a/control/statesp.py
+++ b/control/statesp.py
@@ -55,7 +55,7 @@ from numpy import all, angle, any, array, asarray, concatenate, cos, delete, \
     dot, empty, exp, eye, matrix, ones, pi, poly, poly1d, roots, shape, sin, \
     zeros, squeeze
 from numpy.random import rand, randn
-from numpy.linalg import inv, det, solve
+from numpy.linalg import solve, eigvals, matrix_rank
 from numpy.linalg.linalg import LinAlgError
 from scipy.signal import lti, cont2discrete
 # from exceptions import Exception
@@ -405,7 +405,7 @@ but B has %i row(s)\n(output(s))." % (self.inputs, other.outputs))
     def pole(self):
         """Compute the poles of a state space system."""
 
-        return roots(poly(self.A))
+        return eigvals(self.A)
 
     def zero(self):
         """Compute the zeros of a state space system."""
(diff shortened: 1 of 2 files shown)


mp4096:
Travis fails exactly at the [same test](https://travis-ci.org/python-control/python-control/builds/114399599) as `master`, also see #70

pmli:
Kudos for finding `abs(det(F))`. Determinant gives no information about the distance to singularity, therefore it cannot be used to numerically check for singularity. SVD is definitely much better. But, wouldn't it be even better to drop the singularity check and trust [`numpy.linalg.solve`](http://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.linalg.solve.html) to raise `LinAlgError` if singularity is detected? Do you have a use case when `numpy.linalg.matrix_rank` detects singularity, but `numpy.linalg.solve` doesn't?

mp4096:
Thanks for the hint. Indeed, an additional SVD seems to be superfluous. However, it can be really helpful.

Consider this code:
```python
import numpy as np

matrix_size = 3

for spread in [5, 10, 20]:
    A1 = np.random.rand(matrix_size, matrix_size)
    A2 = np.diag(np.logspace(-spread, +spread, matrix_size))
    A3 = np.random.rand(matrix_size, matrix_size)
       
    A = A1@A2@A3   
       
    invA_A = np.linalg.solve(A, A)
    
    message = "Matrix size: {:d}, matrix rank: {:d}"    
    
    print(message.format(matrix_size, np.linalg.matrix_rank(A)))

    # Should be an identity matrix
    plt.figure()
    plt.imshow(np.log10(np.absolute(invA_A)), interpolation="none")
    plt.colorbar()
```

So instead of raising an exception, `numpy.linalg.solve` just returns some numerical garbage. The problem is, `numpy.linalg.solve` relies on LAPACK's [`DGESV`](http://www.netlib.org/lapack/explore-html/d7/d3b/group__double_g_esolve.html#ga5ee879032a8365897c3ba91e3dc8d512), which in turn calls [`DGETRF`](http://www.netlib.org/lapack/explore-html/dd/d9a/group__double_g_ecomputational.html#ga0019443faea08275ca60a734d0593e60) to perform an LUP decomposition. `DGETRF` raises an exception only if a diagonal element is _exactly_ zero:
```
INFO is INTEGER
= 0:  successful exit
< 0:  if INFO = -i, the i-th argument had an illegal value
> 0:  if INFO = i, U(i,i) is exactly zero. The factorization
    has been completed, but the factor U is exactly
    singular, and division by zero will occur if it is used
    to solve a system of equations.
```

Hence, numerical rank loss cannot be detected in `numpy.linalg.solve`

pmli:
Thanks for the detailed explanation. I assumed NumPy/SciPy would emulate Matlab with its warnings about large condition numbers, using [`DGESVX`](http://www.netlib.org/lapack/explore-html/d7/d3b/group__double_g_esolve.html#ga9d90ccf6e340cacd08b7bbbb502ceb21). At least, there is an open [NumPy issue](https://github.com/numpy/numpy/issues/3755) concerning exactly this, although it is quite old. Until it is closed, I don't see a way to cheaply estimate the condition number in NumPy which could replace SVD.

mp4096:
Oh, I didn't knew about this issue, thanks! But it doesn't seem they're going to do anything about this behaviour.

Anyway, most technical systems have very few (in terms of linear algebra :wink:) inputs/outputs. I think this additional SVD-based rank check will makes things slower only if there are at least 1000 inputs/outputs (we're inverting `D` here).

mp4096:
Thank you! I think one can close #84 and #85 now.

Issue #96, opened 2016-06-07
Issue with control.phase_plot


aramakri:
In the python-control documentation:
http://python-control.readthedocs.io/en/latest/generated/control.phase_plot.html

It says that the phase_plot function "Produces a vector field or stream line plot for a planar system."

It also shows how to plot only streamlines:
"phase_plot(func. X0=(...), T=Tmax, ...)"

I followed this direction with this code:
"def d1(x1x2,t):
    x1,x2 = x1x2
    return np.array([x2, x2 - 2*x1])

x1x2_0 = np.array([[-1.,1.],[-1.,-1.],[1.,1.],[1.,-1.],[-1.,0.],[1.,0.],[0.,-1.],[0.,1.],[0.,0.]])

plt.figure(1)
control.phase_plot(d1,X0=x1x2_0,T=100)"

However, I get a TypeError:
"TypeError                                 Traceback (most recent call last)
<ipython-input-13-e779973f3045> in <module>()
     11 
     12 plt.figure(1)
---> 13 control.phase_plot(d1,X0=x1x2_0,T=100)
     14 
     15 def d2(x1x2,t):

/home/aramakri/cds/local/lib/python2.7/site-packages/control/phaseplot.pyc in phase_plot(odefun, X, Y, scale, X0, T, lingrid, lintime, logtime, timepts, parms, verbose)
    129         #! TODO: Add sanity checks
    130         (x1, x2) = np.meshgrid(
--> 131             frange(X[0], X[1], float(X[1]-X[0])/X[2]),
    132             frange(Y[0], Y[1], float(Y[1]-Y[0])/Y[2]));
    133 

TypeError: 'NoneType' object has no attribute '**getitem**'"

It assumes that there is a meshgrid to plot arrows and returns an error when there is none.

I am trying to plot the streamlines without the arrows because when arrows are plotted, they are so thick that the actual graph is obscured.

murrayrm:
I've verified the bug and set up a branch to fix this.

slivingston:
fixed in commit 0104e4bd036a40e45da70524a13e8356a4ad935e.

PR #97, opened 2016-06-13
Phaseplot typeerror

murrayrm:
Addresses issue #96 where a type error is generate for phase plots without the arrows.  Updated the code to handle this case correctly and also added a unit test to cover this case.

### DIFF
--- a/control/phaseplot.py
+++ b/control/phaseplot.py
@@ -120,7 +120,10 @@ def phase_plot(odefun, X=None, Y=None, scale=1, X0=None, T=None,
     # Figure out ranges for phase plot (argument processing)
     #
     #! TODO: need to add error checking to arguments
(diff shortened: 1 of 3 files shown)


murrayrm:
@slivingston: if you are OK with this change, can you go ahead and merge.  I could do it, but I think it is good practice to have a second set of eyes.  If you don't have time, unassigned and maybe someone else can pick it up.

slivingston:
@murrayrm I missed your request for action. I am reviewing this now.

slivingston:
In commit c498d3d27a5d0c8d1707bb2d67e199b036aedb2c I add a reference to #96 in the comments of the test case introduced in 0104e4bd036a40e45da70524a13e8356a4ad935e (from this pull request).

PR #98, opened 2016-06-23
Do not overwrite np = numpy with an array size.

ochurlaud:
The function failed to be called with python3

### DIFF
--- a/control/robust.py
+++ b/control/robust.py
@@ -90,8 +90,8 @@ def h2syn(P,nmeas,ncon):
 
     n = np.size(P.A,0)
     m = np.size(P.B,1)
-    np = np.size(P.C,0)
-    out = sb10hd(n,m,np,ncon,nmeas,P.A,P.B,P.C,P.D)
+    mp = np.size(P.C,0)
+    out = sb10hd(n,m,mp,ncon,nmeas,P.A,P.B,P.C,P.D)
     Ak = out[0]
     Bk = out[1]
     Ck = out[2]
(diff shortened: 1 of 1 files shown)


slivingston:
I would not expect this to only affect Python 3.

slivingston:
Can you provide an example that demonstrates the error?

Since `np` as an identifier for `numpy` is not used within the function scope after where `np` is assigned to an array size, it is not an error. I agree that changing the name would improve the style, but it is not a matter of correctness.

ochurlaud:
Hum, I don't remember: it was some time ago. Comparing to other modules, t should be call _np I think instead of what I wrote. Anyway: as it's a small fix, without huge repercussions, I think it should be somehow done. 

Even if you don't use my PR, just do the correction somehow?

roryyorke:
I fixed this on #100.  It *is* a bug; if, on that branch, I revert from np_ to np,  I get the error below.  I'd have to read up on how Python handles globals and locals to explain this (did that change in Python 3?).

```
Test h2syn
----------------------------------------------------------------------
  File "robust_test.py", line 34, in testH2syn
    k = control.robust.h2syn(p, 1, 1)
  File "/home/rory/src/python-control/control/robust.py", line 91, in h2syn
    n = np.size(P.A,0)
UnboundLocalError: local variable 'np' referenced before assignment
```

ochurlaud:
Ok, then maybe merging #100 would be better than mine.

slivingston:
Changes from pull request #100 were merged, which subsumes changes from here, [as noted above](https://github.com/python-control/python-control/pull/98#issuecomment-238972383). In particular, cf. lines [93](https://github.com/python-control/python-control/blob/d4e015598cc88d0b08104754b899e1f6156eb68a/control/robust.py#L93) and [157 of control/robust.py](https://github.com/python-control/python-control/blob/d4e015598cc88d0b08104754b899e1f6156eb68a/control/robust.py#L157) as of commit d4e015598cc88d0b08104754b899e1f6156eb68a (current tip of `master` branch).

@ochurlaud Nonetheless, thanks for this pull request. I should have been faster to review it.

PR #101, opened 2016-07-03
Refactored matrix inversions (see #85 and #91)

mp4096:
Continuing work on the issue raised in #85 and partially done in #91:
- Replaced most of the `.I`s and `inv`s with `numpy.linalg.solve` or `numpy.linalg.lstsq`. The remaining inversions which I didn't touch are in `yottalab.py`. Is it still used?
- Where applicable, replaced `A\B`, `A\C` by solving `A\[B, C]` and taking views for the corresponding results. This avoids redundant cubic LU decompositions.
- Added singularity checks before matrix inversions.
- Added two unit tests for the reachable canonical form.
- Refactored an [old TODO](https://github.com/python-control/python-control/blob/master/control/statesp.py#L127).
- Fixed a minor typo: [`compatability`](https://en.wiktionary.org/wiki/compatability) -> [`compatability`](https://en.wiktionary.org/wiki/compatibility#English)

~~I commented out the matrix singularity check in [`modelsimp.py`](https://github.com/mp4096/python-control/blob/master/control/modelsimp.py#L183) since it correctly raises an error during [`testModred()`](https://github.com/python-control/python-control/blob/master/control/tests/matlab_test.py#L364) in `matlab_test.py`. Matrix `A22` is indeed numerically singular in this case (second column is something like `[[0.0, 3.0e-16]]`). Should we change the test case?~~ See comment for bf6c446f0e5beb502d2139fa2c65c87d17a83bf8

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -148,7 +148,7 @@
 2012-11-03  Richard Murray  <murray@altura.local>
 
 	* src/rlocus.py (_RLSortRoots): convert output of range() to
-	explicit list for python 3 compatability
+	explicit list for python 3 compatibility
 
 	* tests/modelsimp_test.py, tests/slycot_convert_test.py,
 	tests/mateqn_test.py, tests/statefbk_test.py: updated test suites to
@@ -604,7 +604,7 @@
 	initial_response, impulse_response and step_response.
 
 	* src/rlocus.py: changed RootLocus to root_locus for better
-	compatability with PEP 8.  Also updated unit tests and examples.
+	compatibility with PEP 8.  Also updated unit tests and examples.
 
 2011-07-25  Richard Murray  <murray@malabar.local>
 
@@ -994,7 +994,7 @@
 2010-09-02  Richard Murray  <murray@sumatra.local>
 
 	* src/statefbk.py (place): Use np.size() instead of len() for
-	finding length of placed_eigs for better compatability with
+	finding length of placed_eigs for better compatibility with
 	different python versions [courtesy of Roberto Bucher]
 
 	* src/delay.py (pade): New file for delay-based computations +
--- a/control/canonical.py
+++ b/control/canonical.py
@@ -7,7 +7,7 @@ from .statesp import StateSpace
 from .statefbk import ctrb, obsv
 
 from numpy import zeros, shape, poly
-from numpy.linalg import inv
+from numpy.linalg import solve
 
 __all__ = ['canonical_form', 'reachable_form', 'observable_form']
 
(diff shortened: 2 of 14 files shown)


robertobucher:
The original "yottalab" project doesn't exixt anymore. It has been 
splitted into two different modules:

- ctrl_repl.py: replacement or implementation of particular functions, 
containing
    - d2c
    - dlqr
    - place (using a wrapper to fortran functions derived from the 
scicoslab project

and

- ctrl_utils with design utilities like:
   - full_obs    - full order observer
   - red_obs     - reduced order observer
   - comp_form   - state feedback controller+observer in compact form
   - comp_form_i - state feedback controller+observer+integ in compact form
   - set_aw      - introduce anti-windup into controller
   - grstep      - graphical step response

All the other functions of the original yottalab project have migrated 
in the control toolbox.

Best regards

Roberto

BTW: For the Swiss Control Association I've prepared a little 
description about Python for control purposes.
The link is 
http://robertobucher.dti.supsi.ch/bucherdl/BookPythonForControl.pdf

mp4096:
1fe639c938a00412969483fb25dda3be277f0351: After some consideration I've also replaced the rank check with both `numpy.linalg.matrix_rank` _and_ `numpy.linalg.det` with a single `numpy.linalg.matrix_rank` check. Determinant is really unsuitable for real-life matrix rank computation.

Case in point:
```python
import numpy as np

A = np.matrix("1.0e-3, 0.0; 0.0, 1.0e-3")
print(np.linalg.det(A))
print(np.linalg.matrix_rank(A))
```

This change _does_ affect unit tests: Fewer systems are skipped in `testAcker`.

mp4096:
bf6c446f0e5beb502d2139fa2c65c87d17a83bf8: Added singularity check before matrix inversion for DC matching / residualization. I took the liberty to change the unit test slightly, since it does not really matter anyway (see the docstring, it just calls MATLAB-like methods and does not check the results). Instead of eliminating the 2nd and the 3rd state variable, it now eliminates the 1st and the 2nd.

2c25b736701024fec8d3582b7860e05c0932e44a: Adds a unit test to check residualization of unstable systems (should thrown a `ValueError` exception).

32323d4a8607930679ca198480ea0966e1680382: Made code more pythonic, in this case list comprehensions instead of for-loops. Also call `np.linalg.eigvals` instead of `np.linalg.eig`, for we do not need the eigenvector matrix, only the eigenvalues.

mp4096:
See https://travis-ci.org/mp4096/python-control/builds/144536541 for CI results with the correct configuration from #102.

slivingston:
I will review this tonight.

slivingston:
For posterity, the "old TODO" that is referenced in the opening post of this pull request is in [control/statesp.py at line 127](https://github.com/python-control/python-control/blob/c498d3d27a5d0c8d1707bb2d67e199b036aedb2c/control/statesp.py#L127) as of commit c498d3d27a5d0c8d1707bb2d67e199b036aedb2c (current tip of `master` branch).

slivingston:
The change in your first commit (<https://github.com/python-control/python-control/pull/101/commits/4d70d00255864df2651f8375935d31dd5fc0e7a0>) refactors some code to use `map`, whereas changes from PR #110 refactor that code using a list comprehension.

I compared the two alternatives using %timeit of Jupyter for matrices that would correspond to LTI systems with 10 state dimensions, 2 inputs, 2 outputs (so, the A matrix has shape (10, 10), B matrix  shape (10,2), etc.), and with 100 state dimensions, 20 inputs, 20 outputs, all using matrices from numpy.random.random((r,c)).

The difference in timing performance does not appear significant. However, the case of `map` was slightly  (approximately 500 to 900 ns less in duration), so I recommend that we use it.

slivingston:
@mp4096 Thanks for these improvements of numerical reliability.

As future work, I think that it would be interesting to add more test cases that would fail to be detected if `numpy.linalg.inv` was used instead of `numpy.linalg.solve`.

PR #102, opened 2016-07-06
BuildFix: change Travis build to handle conda changes

roryyorke:
Changes:
  - do build and test inside a conda environment
  - specify Python version at conda build stage

These changes triggered by builds failing *without* changes in HEAD; I assume this is due to Conda changes, rather than Travis changes.

See https://travis-ci.org/roryyorke/python-control/builds/142789873 for an example of master HEAD failing; first noticed in the Python 3.3 travis failure in #101 (haven't tested that on top of this fix, but given that it passes on Python 2.7 and 3.4, it seems like it would pass).

Both .travis.yml and conda-recipe/meta-yaml could do with a review by an expert (I'm not one); e.g., why is "nose" a build requirement, rather than a test requirement?  Shouldn't we follow standard (0.7.1+stuff) versioning?  Why does the version end up as 0.6d.stuff on my Travis builds, not 0.7.stuff?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,19 +28,24 @@ before_install:
   - hash -r
   - conda config --set always_yes yes --set changeps1 no
   - conda update -q conda
-  - conda install --yes python=$TRAVIS_PYTHON_VERSION conda-build pip coverage
+  # conda-build must be installed in the conda root environment
+  - conda install conda-build
+  # TODO: point at conda.anaconda.org?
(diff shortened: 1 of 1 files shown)


mp4096:
Many thanks for the repaired CI config!

Works like a charm for #101: https://travis-ci.org/mp4096/python-control/builds/142872267

I would actually suggest you add Python 3.5 as a fourth test environment, it seems to gain momentum steadily...

slivingston:
I will review this today.

PR #103, opened 2016-08-07
Add function to convert a system into observable canonical form

yuina822:
Only for SISO systems

### DIFF
--- a/control/canonical.py
+++ b/control/canonical.py
@@ -4,12 +4,12 @@
 from .exception import ControlNotImplemented
 from .lti import issiso
 from .statesp import StateSpace
-from .statefbk import ctrb
+from .statefbk import ctrb, obsv
 
 from numpy import zeros, shape, poly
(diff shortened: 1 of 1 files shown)


murrayrm:
Can we put in some unit tests to confirm that this is working.  (And while we are at it, perhaps add some unit tests for reachable canonical form?)

Also, there seems to be a a problem with the python 3.3 build (looks like a conda problem/version conflict) that we should sort out.

mp4096:
I actually added two tests for the reachable canonical form in #101, see [here](https://github.com/python-control/python-control/pull/101/files#diff-2ac1f4a4c19e1c07d2339d182b5842de). One can easily adapt them for the OCF.

yuina822:
I have write a observable test unit 5f6139c. Thank you @mp4096 . Unobservable test didn't done well because exception handling for unobservable systems isn't added. I think we could  add exception handling for unreachable/unobservable systems.

PR #105, opened 2016-08-10
UnSkipTest: these tests pass on Python 2.7, 3.3, 3.4, and 3.5.

roryyorke:


slivingston:
Does anyone know why these were being skipped originally? The `unittest.skip` decorators were introduced in 6d9da6eacd81887589fdfae25f3076353a013690, which was a big commit that mixes many different kinds of changes, and the commit message does not offer relevant explanation.

slivingston:
Otherwise, I agree with the changes and will merge after some comment about the original motivation is made.

Perhaps @jgoppert knows.

murrayrm:
I took a quick look and couldn't see any reason why these were skipped.

slivingston:
Before merging this, I found that all tests pass on a Debian GNU/Linux x86_64 box with Python versions 2.7.9, 3.3.6, 3.4.2, 3.5.2

murrayrm:
Decided not to track this against milestone 0.8.0 since it is primarily an internal change (so not visible to users).

PR #107, opened 2016-08-11
Allow static gain StateSpace objects to be straightforwardly created.

roryyorke:
Lets the (fairly?) obvious StateSpace([],[],[],D) work.  One could previously create these by, e.g., converting a static gain TransferFunction to a StateSpace object.

In this case A will be 0-by-0, B 0-by-inputs, and C outputs-by-0.  My intuition is that this is a generally correct approach (consider the simplification of _remove_useless_states, for example), but it may require some thought during review.

### DIFF
--- a/control/statesp.py
+++ b/control/statesp.py
@@ -146,7 +146,7 @@ a StateSpace object.  Recived %s." % type(args[0]))
         if self.states != B.shape[0]:
             raise ValueError("A and B must have the same number of rows.")
         if self.states != C.shape[1]:
-            raise ValueError("A and C C must have the same number of columns.")
+            raise ValueError("A and C must have the same number of columns.")
         if self.inputs != B.shape[1]:
             raise ValueError("B and D must have the same number of columns.")
         if self.outputs != C.shape[0]:
--- a/control/statesp.py
+++ b/control/statesp.py
@@ -122,34 +122,35 @@ a StateSpace object.  Recived %s." % type(args[0]))
         else:
             raise ValueError("Needs 1 or 4 arguments; received %i." % len(args))
 
-        # Here we're going to convert inputs to matrices, if the user gave a
-        # non-matrix type.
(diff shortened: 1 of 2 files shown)


control/statesp.py
```diff
@@ -122,34 +122,35 @@ def __init__(self, *args):
...
         if self.states != B.shape[0]:
-            raise ValueError("B must have the same row size as A.")
+            raise ValueError("A and B must have the same number of rows.")
         if self.states != C.shape[1]:
-            raise ValueError("C must have the same column size as A.")
-        if self.inputs != D.shape[1]:
-            raise ValueError("D must have the same column size as B.")
-        if self.outputs != D.shape[0]:
-            raise ValueError("D must have the same row size as C.")
+            raise ValueError("A and C C must have the same number of columns.")
```
mp4096: I think it's one `C` too many here. :smile_cat:

roryyorke:
D'oh!  Thanks, will fix.

cwrowley:
I like this syntax for creating static gain objects.  This PR looks good to me.

roryyorke:
One case not covered by this PR is the "empty" system, `StateSpace([],[],[],[])`.  My first attempt at handling this broke `StateSpace.__getitem__`; I haven't yet come up with a tidy fix.  The source of the hassle, IMO, is that `numpy.matrix([])` returns a 1x0 matrix, rather than 0x0. My current thinking is to have a `numpy.matrix` wrapper, `statesp._matrix`, for which `_matrix([]` is 0x0.

Wanting 0-input, 0-output systems may seem like fussiness, but it turns out to be necessary (or useful, at least), in a prototype of `augw` I've written. (`augw` ends up being part of `mixsyn`; see Octave's docs for `augw` at [1]).

[1] http://octave.sourceforge.net/control/function/augw.html

slivingston:
Pull request #110 subsumes this one.

PR #108, opened 2016-08-12
Allow minreal to work with static gain StateSpace objects

roryyorke:
Finding this bug is what provoked #107, which this builds on.

This takes the simple option of doing nothing for a static gain.  I didn't investigate the slycot.tb01pd call at all; perhaps it can take a static gain, and there's some other bug lurking in minreal code.


slivingston:
Pull request #110 subsumes this one.

PR #109, opened 2016-08-12
Enable Travis builds for Python 3.5

roryyorke:
Following suggestion of @mp4096, add Python 3.5 to .travis.yml.  Sorry @slivingston, you merged #102 before I could push this there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ python:
   - "2.7"
   - "3.3"
   - "3.4"
+  - "3.5"
 
 # install required system libraries
 before_install:


slivingston:
Thanks for adding Python 3.5 as a target. This is also motivated by #105. We should next add
```
Programming Language :: Python :: 3.5
```
to setup.py.

slivingston:
re https://github.com/python-control/python-control/pull/109#issuecomment-239573286
In 58a3ff2036aab49b63d07684c60a35d59cb395f3, I add 3.5 to the list of supported Python versions. For reference, the classifiers that are currently recognized on PyPI are <https://pypi.python.org/pypi?%3Aaction=list_classifiers>

PR #110, opened 2016-08-31
Rory ss static pole

roryyorke:
This is a mod on top of #108, rebased on cdd3e73.  If #107 and #108 and this mod are all OK, I guess one could just take this PR and delete the others (which I haven't rebased yet).

The mod lets StateSpace.pole work with static gains; it returns np.array([]), which is both reasonable, and the behaviour as TransferFunction.pole for static gains.  _Without_ this mod, the eigval call in StateSpace.pole results in an exception when A is an empty matrix.

Regression test added.

I'm pretty sure the Python 3.3 failure on Travis CI [1] hasn't got anything to do with this.

Prompted by modelsimp.minreal bugging out when minreal()'ing a statespace object was reduced to a static gain.

[1] https://travis-ci.org/roryyorke/python-control/jobs/156620757

### DIFF
--- a/control/tests/statesp_test.py
+++ b/control/tests/statesp_test.py
@@ -254,10 +254,14 @@ class TestStateSpace(unittest.TestCase):
         np.testing.assert_array_equal(np.diag([2,3]),g6.D)
 
     def test_matrixStaticGain(self):
-        """Regression: can we create a scalar static gain?"""
+        """Regression: can we create matrix static gains?"""
         d1 = np.matrix([[1,2,3],[4,5,6]])
         d2 = np.matrix([[7,8],[9,10],[11,12]])
         g1=StateSpace([],[],[],d1)
+
+        # _remove_useless_states was making A = [[0]]
+        self.assertEqual((0,0), g1.A.shape)
+
         g2=StateSpace([],[],[],d2)
         g3=StateSpace([],[],[],d2.T)
 
@@ -269,6 +273,19 @@ class TestStateSpace(unittest.TestCase):
         np.testing.assert_array_almost_equal(solve(np.eye(2)+d1*d2,d1), h3.D)
         h4 = g1.append(g2)
         np.testing.assert_array_equal(block_diag(d1,d2),h4.D)
+
+
+    def test_remove_useless_states(self):
+        """Regression: _remove_useless_states gives correct ABC sizes"""
+        g1 = StateSpace(np.zeros((3,3)),
+                        np.zeros((3,4)),
+                        np.zeros((5,3)),
+                        np.zeros((5,4)))
+        self.assertEqual((0,0), g1.A.shape)
+        self.assertEqual((0,4), g1.B.shape)
+        self.assertEqual((5,0), g1.C.shape)
+        self.assertEqual((5,4), g1.D.shape)
+        self.assertEqual(0, g1.states)
 
 
     def test_BadEmptyMatrices(self):
--- a/control/tests/statesp_test.py
+++ b/control/tests/statesp_test.py
@@ -378,6 +378,12 @@ class TestDrss(unittest.TestCase):
                         self.assertTrue(abs(z) < 1)
 
 
+    def testPoleStatic(self):
+        """Regression: pole() of static gain is empty array"""
+        np.testing.assert_array_equal(np.array([]),
+                                      StateSpace([],[],[],[[1]]).pole())
+
+
 def suite():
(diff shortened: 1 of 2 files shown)


roryyorke:
Reran this on Travis today, everything passes [1].

[1] https://travis-ci.org/roryyorke/python-control/builds/156620754

slivingston:
I will review it within 12 hours.

slivingston:
However, changing to use `super` is orthogonal to this pull request, so we can wait until later to do so.

control/tests/statesp_test.py
```diff
@@ -5,9 +5,10 @@
 
 import unittest
 import numpy as np
-from scipy.linalg import eigvals
+from numpy.linalg import solve
+from scipy.linalg import eigvals, block_diag
 from control import matlab
-from control.statesp import StateSpace, _convertToStateSpace
+from control.statesp import StateSpace, _convertToStateSpace,tf2ss
```
slivingston: Though `tf2ss` is added here, it is not used in this module (statesp_test.py). Am I missing something?

slivingston:
@roryyorke Did you write "BugFix?" in the first commit to indicate that the change might be thought of as adding a new feature, instead of fixing a bug?

slivingston:
Regarding the behavior of `numpy.matrix([])`, I did not find any written justification. It might be worthwhile for someone else to also search.

I studied the constructor code for `matrix` objects, which shows that `numpy.matrix([])` returns a value with shape `(1,0)` because

1. none of the `isinstance` calls match for `[]` (type `list`);
2. `[]` is passed through `numpy.array`, which returns an `ndarray` of shape `(0,)` [1];
3. `np.array([]).ndim == 1`, so the `matrix` object is given `shape = (1, np.array([]).shape[0])` [2].

Thus, I guess that one design motivation is to map 1-dimensional arrays to row matrices, without specially treating empty arrays.

[1] line 270 of numpy/matrixlib/defmatrix.py, <https://github.com/numpy/numpy/blob/2498b74420074dffa3a062ab7fc69dfeb88a5050/numpy/matrixlib/defmatrix.py#L270>
[2] lines 277-278 of numpy/matrixlib/defmatrix.py, <https://github.com/numpy/numpy/blob/2498b74420074dffa3a062ab7fc69dfeb88a5050/numpy/matrixlib/defmatrix.py#L277-L278>

slivingston:
My recommendations:

* Assert `self.states == 0` in `test_remove_useless_states` of statesp_test.py because StateSpace._remove_useless_states contains the assignment `self.states = self.A.shape[0]`, and one consequence of changes from your pull request is that `self.states = 0` when all states are useless (instead of `self.states = 1` from before).

* Remove the new import of `tf2ss` in statesp_test.py. If you can quickly rebase commits so as to entirely avoid the noise (i.e., accidental addition, then removal), please do. Otherwise, OK to just add a correction commit.

Once these are addressed, this PR is ready to be merged.

roryyorke:
Thanks for the thorough review! I'll put all my responses here.

TODO for super: I think we agree this change doesn't belong in this PR; I suppose the Right Thing would have been to open a new issue.

"BugFix?": indeed, whether this is a bugfix or enhancement is not very clear. I can edit the log (I guess? never done that with git) if you like.

On empty SS objects: I have implemented this, unfortunately didn't push it to the PR. It follows the proposed solution.  See [1].

tf2ss: I don't remember why that's there; some sort of exploratory debugging, I guess. Will remove. I'll see if I can do the rebase you suggest.

I'll add the recommended assertion, thanks.

OK, so for now I will:
  - rebase this on c498d3d
  - remove the ? from the "BugFix?" log message
  - remove the unnecessary tf2ss import
  - add the extra assertion

The empty SS object extension of [1] is on the same theme as the rest of the PR, but would require additional review.  I could either add it here, or open a new PR for it.

[1] https://github.com/roryyorke/python-control/commit/ba75ae535413e9edfb1dc787d65451ea0c238d8a

roryyorke:
I squashed the "messagefix" commit in the rebase, otherwise changes as in previous comment.

roryyorke:
I squashed the messagefix commit, otherwise rebase is as described in previous comment.

slivingston:
Thanks; the rebase is good. To better organize discussion, I decided to leave consideration of a wrapper of `numpy.matrix([])` (as in <https://github.com/python-control/python-control/pull/110#issuecomment-269400683> and <https://github.com/roryyorke/python-control/commit/ba75ae535413e9edfb1dc787d65451ea0c238d8a>) to another issue or PR.

PR #118, opened 2016-10-21
Updates to balred() and gram()

mdclemen:
balred() now supports "matchdc" and will handle unstable systems and will accept a list of ORDERS, which will return a list of reduced order systems. gram() now does Cholesky factored gramians if desired. Requires my updates to [slycot](https://github.com/python-control/Slycot/pull/2).

### DIFF
--- a/control/modelsimp.py
+++ b/control/modelsimp.py
@@ -203,7 +203,7 @@ def balred(sys, orders, method='truncate', alpha=None):
     States are eliminated based on Hankel singular value.
     If sys has unstable modes, they are removed, the
     balanced realization is done on the stable part, then
-    reinserted IAW reference below.
+    reinserted in accordance with the reference below.
 
     Reference: Hsu,C.S., and Hou,D., 1991,
     Reducing unstable linear control systems via real Schur transformation.
@@ -219,12 +219,12 @@ def balred(sys, orders, method='truncate', alpha=None):
     method: string
         Method of removing states, either ``'truncate'`` or ``'matchdc'``.
     alpha: float
-        Specifies the alpha-stability boundary for the eigenvalues
-        of the state dynamics matrix A. For a continuous-time
-        system, alpha <= 0 is the boundary value for the real parts
-        of eigenvalues, while for a discrete-time system, 0 <= alpha <= 1
-        represents the boundary value for the moduli of eigenvalues.
-        The alpha-stability domain does not include the boundary.
+        Redefines the stability boundary for eigenvalues of the system matrix A.
+        By default for continuous-time systems, alpha <= 0 defines the stability
+        boundary for the real part of A's eigenvalues and for discrete-time
+        systems, 0 <= alpha <= 1 defines the stability boundary for the modulus
+        of A's eigenvalues.See SLICOT routines AB09MD and AB09ND for more
+        information.
 
     Returns
     -------
@@ -250,14 +250,14 @@ def balred(sys, orders, method='truncate', alpha=None):
         raise ValueError("supported methods are 'truncate' or 'matchdc'")
     elif method=='truncate':
         try:
-            from slycot import ab09md
+            from slycot import ab09md, ab09ad
         except ImportError:
-            raise ControlSlycot("can't find slycot subroutine ab09md") 
+            raise ControlSlycot("can't find slycot subroutine ab09md or ab09ad")
     elif method=='matchdc':
         try:
             from slycot import ab09nd
         except ImportError:
-            raise ControlSlycot("can't find slycot subroutine ab09nd") 
+            raise ControlSlycot("can't find slycot subroutine ab09nd")
 
     #Check for ss system object, need a utility for this?
 
@@ -289,7 +289,13 @@ def balred(sys, orders, method='truncate', alpha=None):
         m = np.size(sys.B,1)
         p = np.size(sys.C,0)
         if method == 'truncate':
-            Nr, Ar, Br, Cr, Ns, hsv = ab09md(dico,job,equil,n,m,p,sys.A,sys.B,sys.C,alpha=alpha,nr=i,tol=0.0)
+            #check system stability
+            if np.any(np.linalg.eigvals(sys.A).real >= 0.0):
+                #unstable branch
+                Nr, Ar, Br, Cr, Ns, hsv = ab09md(dico,job,equil,n,m,p,sys.A,sys.B,sys.C,alpha=alpha,nr=i,tol=0.0)
+            else:
+                #stable branch
+                Nr, Ar, Br, Cr, hsv = ab09ad(dico,job,equil,n,m,p,sys.A,sys.B,sys.C,nr=i,tol=0.0)
             rsys.append(StateSpace(Ar, Br, Cr, sys.D))
 
         elif method == 'matchdc':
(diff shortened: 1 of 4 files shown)


control/statefbk.py
```diff
@@ -347,6 +347,9 @@ def gram(sys,type):
     #Check for ss system object
     if not isinstance(sys,statesp.StateSpace):
         raise ValueError("System must be StateSpace!")
+    gramtypes = ['c', 'o', 'cf', 'of']
```
slivingston: Why do you declare `gramtypes` as a variable? It only appears to be used in the conditional expression on the next line, so the motivation for allocating a variable with function-level scope is not apparent to me.

control/modelsimp.py
```diff
@@ -50,7 +50,7 @@
 from .statesp import StateSpace
 from .statefbk import gram
 
-__all__ = ['hsvd', 'balred', 'modred', 'era', 'markov', 'minreal']
+__all__ = ['hsvd', 'balred', 'modred', 'era', 'markov', 'minreal',]
```
slivingston: Why did you add a trailing `,` in this list expression? It is not a syntax error, but it is not consistent with the current style of the `control` package.

mdclemen: I forgot to delete the trailing ','.

control/statefbk.py
mdclemen: I don't really remember, I wrote that function a few months ago. If you think there is a better way to do things, then by all means change it.

slivingston: For the reason that I gave above, moving the list into the conditional expression improves code quality, and thus, I recommend the change. However, it is not critical for accepting this pull request.

Just to be clear, I am belatedly providing review of your proposed changes now.

control/modelsimp.py
slivingston: Can you delete it, then? I am writing additional comments now, so you might want to wait before rebasing until I send them all.

slivingston:
In the first commit (<https://github.com/python-control/python-control/pull/118/commits/cc4b5c75e676b86dd3ba16850c6a7361b0f30f0b>) of this pull request, the message and author are incorrectly copied from cdd3e73838cea0b26a7155d0b2084aa698e5eefd. When I skim the diff, it appears that you applied changes as part of your own work, but then, those were combined with the patch corresponding to commit cdd3e73838cea0b26a7155d0b2084aa698e5eefd. Can you clean it up, in particular remove the re-applied pieces of commit cdd3e73838cea0b26a7155d0b2084aa698e5eefd?

slivingston:
It appears that substantial amounts of code are added in the first commit and then deleted or modified in the second. Can you add details in the commit messages to explain this transition? Alternatively, can you squash changes into a single commit?

Those are all of my questions and requests for now. I will wait until you respond or rebase before continuing to review.

mdclemen:
I made changes to the code that you recommended, but my git skills might not be good enough for the other things you want me to do.

slivingston:
@mdclemen Can I have your permission to create a commit that has your changes from this PR and lists you as author (and me as committer)? This way I can make the rebasing adjustments that I requested above. If I do this, how do you prefer your name to be written? E.g., I can use "clementm" as in your commit <https://github.com/python-control/python-control/pull/118/commits/f74e66552aa30645166f62fa4e5c115a196f85ca>, or your name as it appears in your publications.

Alternatively, we can arrange to meet in IRC to work through the `rebase` process together.

control/modelsimp.py
```diff
@@ -211,26 +218,46 @@ def balred(sys, orders, method='truncate'):
         of systems)
     method: string
         Method of removing states, either ``'truncate'`` or ``'matchdc'``.
+    alpha: float
+        Specifies the alpha-stability boundary for the eigenvalues
```
slivingston: This text description of `alpha` appears to be copied from AB09MD.f of the SLICOT source code. Is that true? SLICOT is under the GNU GPL, whereas we want to keep this `control` package under [the BSD license](https://github.com/python-control/python-control/blob/c1f5a9e06fa1cce06a859edcfb41b257ab772372/LICENSE).

slivingston: If the text was indeed copied, can you create a new description of the `alpha` parameter? An easy solution for now is to write a brief summary and indicate where more information can be found (the documentation of SLICOT).

slivingston: The original text from SLICOT to which I am referring is in [lines 89-96 of AB09MD.f](https://github.com/python-control/Slycot/blob/b01430c0f5117e1d6b6fd1717e04fe765a37ebf2/slycot/src/AB09MD.f#L89-L96)

control/modelsimp.py
```diff
@@ -197,10 +197,17 @@ def modred(sys, ELIM, method='matchdc'):
...
     return rsys
 
-def balred(sys, orders, method='truncate'):
+def balred(sys, orders, method='truncate', alpha=None):
     """
     Balanced reduced order model of sys of a given order.
     States are eliminated based on Hankel singular value.
+    If sys has unstable modes, they are removed, the
+    balanced realization is done on the stable part, then
+    reinserted IAW reference below.
```
slivingston: What is "IAW"?

mdclemen:
In Accordance With

control/modelsimp.py
slivingston: can you write "in accordance with" explicitly? The abbreviation might not be widely recognized.

I can also make the change after merging.

slivingston:
I resolved the merge conflicts, but before I apply them here, can you first address my question about the description of the `alpha` parameter?

control/modelsimp.py
mdclemen: I did in fact copy the description of `alpha` from AB09MD.f.

slivingston:
Should we keep the original call to `AB09AD` when the system is stable? Otherwise (if `method='truncate'` and the system is not stable), your newly introduced call to `AB09MD` is used.

The Fortran implementations of these two subroutines are different, so I am motivated to understand them more before removing the one dedicated to the case of stable systems.

slivingston:
Once we decide actions regarding `AB09AD` and the description of the `alpha` parameter, then this PR will be ready for merging. Thanks for your good work on the `gram` and `balred` functions!

mdclemen:
I don't know why the SLICOT authors decided to include a user defined boundary for what constitutes unstable eigenvalues but essentially `alpha` allows the user to redefine the boundary for a stable eigenvalue.

mdclemen:
@slivingston let me know what you think of my updates.

control/modelsimp.py
```diff
@@ -219,12 +219,12 @@ def balred(sys, orders, method='truncate', alpha=None):
...
-        of the state dynamics matrix A. For a continuous-time
-        system, alpha <= 0 is the boundary value for the real parts
-        of eigenvalues, while for a discrete-time system, 0 <= alpha <= 1
-        represents the boundary value for the moduli of eigenvalues.
-        The alpha-stability domain does not include the boundary.
+        Redefines the stability boundary for eigenvalues of the system matrix A.
+        By default for continuous-time systems, alpha <= 0 defines the stability
+        boundary for the real part of A's eigenvalues and for discrete-time
+        systems, 0 <= alpha <= 1 defines the stability boundary for the modulus
+        of A's eigenvalues.See SLICOT routines AB09MD and AB09ND for more
```
slivingston: While resolving the merge conflict, I added a space here after the `.`. So, now it reads "...eigenvalues. See SLICOT...".

slivingston:
Your updates are good. Thanks for addressing my comments.

PR #122, opened 2016-12-26
Allow unit tests to run without slycot

murrayrm:
This PR provides updates to allow unit tests to run correctly when `slycot` is not installed.  There is also a change in `xferfcn.py` to allow proper creation of transfer functions when num or den is a 0d ndarray.  This can occur when `slycot` is not installed due to a difference in the way that transfer functions are created from state space matrices in `xferfcn._convertToTransferFunction()` when `slycot` is not available.

### DIFF
--- a/control/tests/convert_test.py
+++ b/control/tests/convert_test.py
@@ -162,6 +162,29 @@ class TestConvert(unittest.TestCase):
                             np.testing.assert_array_almost_equal( \
                                 ssorig_imag, tfxfrm_imag)
 
+    def testConvertMIMO(self):
+        """Test state space to transfer function conversion."""
+        verbose = self.debug
+
+        # Do a MIMO conversation and make sure that it is processed
+        # correctly both with and without slycot
+        #
+        # Example from issue #120, jgoppert
+        import control
+
+        # Set up a transfer function (should always work)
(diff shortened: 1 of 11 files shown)


control/tests/convert_test.py
```diff
@@ -162,6 +162,29 @@ def testConvert(self):
...
+
+        # Do a MIMO conversation and make sure that it is processed
+        # correctly both with and without slycot
+        #
+        # Example from issue #120, jgoppert
+        import control
+
+        # Set up a transfer function (should always work)
+        tfcn = control.tf([[[-235, 1.146e4],
+                          [-235, 1.146E4],
```
slivingston: Indentation here and on the next 4 lines seems to be slightly off. I recommend
```python
        tfcn = control.tf([[[-235, 1.146e4],
                            [-235, 1.146E4],
                            [-235, 1.146E4, 0]]],
                          [[[1, 48.78, 0],
                            [1, 48.78, 0, 0],
                            [0.008, 1.39, 48.78]]])
```

slivingston:
For future work, I think that it would be useful to either:

* notify the user when a test case is modified according to whether `slycot` is found; or
* separate instances of the above into uniquely named tests, so that a test case is never hot-swapped but rather skipped or not.

PR #123, opened 2016-12-27
Small documentation updates 

murrayrm:
This PR includes some small updates to the python-control documentation, including updating the path for Slycot, adding some entries for missing functions, and adding a new "conventions" section.


slivingston:
From commit <https://github.com/python-control/python-control/pull/123/commits/1f156781f3eb239ddd458c91ec49d7defa2747e0>, names from `control.robust` are copied into the top-level (i.e., `control`) of the package. This is the only set of changes of this PR that involves more than documentation.

I think that the change is reasonable because there have not been collisions among desired names of function or classes (at least, not that I am aware of). Furthermore, the change is consistent with the current pattern of organization.

murrayrm:
The change in `__init.py__` in commit 1f15678 was actually required in order to get those functions to show up properly in the documentation.  So while it has impact beyond documentation, there is a reason that it was included in this PR.

doc/intro.rst
```diff
@@ -35,14 +35,38 @@ some thing to keep in mind:
...
+correctly by running the command::
+
+  python -c "import slycot"
+
+and verifying that no error message appears.  It may be necessary to install
+`slycot` from source, which requires a working FORTRAN compiler and the
+`lapack` library.  More information on the slycot package can be obtained
+from the `slycot project page <https://github.com/python-control/Slycot>`_.
+
+For users with a working the Anaconda distribution of Python, the following
```
slivingston: Should this be "...with the Anaconda distribution..."?

@murrayrm If you agree, I can merge and then delete the words "a working".

slivingston:
I just noticed another commit involving more than documentation: <https://github.com/python-control/python-control/pull/123/commits/93b50b13ecb9e04039958098e40beed5d4590a19>. [Your comment above](https://github.com/python-control/python-control/pull/123#issuecomment-269843123) applies here as well.

OK to include the change. I mention it here only as correction to [my earlier comment](https://github.com/python-control/python-control/pull/123#issuecomment-269842565) where I describe <https://github.com/python-control/python-control/pull/123/commits/1f156781f3eb239ddd458c91ec49d7defa2747e0> as the only such commit.

doc/intro.rst
slivingston: The change is trivial, if the text should indeed be different, and can be applied directly to `master` branch.

slivingston:
There are several instances of trailing whitespace, but they are small and only in documentation, so I decided to merge and avoid further delay.

PR #125, opened 2016-12-31
DOC: fixed additional small typo in installation docs

murrayrm:
Forgot to push a local documentation change.  This PR fixes a small error in the conda installation instructions and also adds a command to make sure numpy, scipy and matplotlib are installed (all pre-requisites).

### DIFF
--- a/doc/intro.rst
+++ b/doc/intro.rst
@@ -59,9 +59,10 @@ and verifying that no error message appears.  It may be necessary to install
 from the `slycot project page <https://github.com/python-control/Slycot>`_.
 
 For users with a working the Anaconda distribution of Python, the following
-command can be used::
+commands can be used::
 
-  conda install -c python-control -c cyclus slycot python-control
+  conda install numpy scipy matplotlib    # if not yet installed
+  conda install -c python-control -c cyclus slycot control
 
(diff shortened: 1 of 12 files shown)


slivingston:
Good catch of the error in the `conda install` command! I missed it while reviewing #122.

If it is a working Anaconda installation, it should already have `numpy`, `scipy`, and `matplotlib`, as listed at <https://docs.continuum.io/anaconda/pkg-docs.html>. Nonetheless an extra `conda install numpy scipy matplotlib` should not cause trouble.

doc/intro.rst
```diff
@@ -59,9 +59,10 @@ and verifying that no error message appears.  It may be necessary to install
 from the `slycot project page <https://github.com/python-control/Slycot>`_.
 
 For users with a working the Anaconda distribution of Python, the following
```
slivingston: Should this be "...with the Anaconda distribution..."?

(copied from <https://github.com/python-control/python-control/pull/123#discussion_r94277363>; still relevant here)

slivingston:
The CI test failures are caused by 500 response codes from requests to repo.continuum.io. I get the same response when I attempt `conda install -c python-control -c cyclus slycot control` from my own local host, so likely an error on the (Continuum) server side.

Issue #114, opened 2016-09-22
dcgain on discrete-time systems is wrong


roryyorke:
dcgain should distinguish between continuous- and discrete-time systems when determining DC gain.

``` python
import control
gct=control.tf([1],[1,1])
gcs=control.ss(gct)
gdt=control.sample_system(gct,5e-2,method='zoh')
gds=control.ss(gdt)
print('gct.dcgain()',gct.dcgain())
print('gcs.dcgain()',gcs.dcgain())
print('gdt.dcgain()',gdt.dcgain())
print('gds.dcgain()',gds.dcgain())
print('gdt.horner(1)',gdt.horner(1))
print('gds.horner(1)',gds.horner(1))
```

gives

```
gct.dcgain() 1.0
gcs.dcgain() 1.0
gdt.dcgain() -0.051271096376024033
gds.dcgain() -0.051271096376024033
gdt.horner(1) [[ 1.+0.j]]
gds.horner(1) [[ 1.]]
```

slivingston:
addressed in commit 310f5805f0fd51aa9e1191aaf3083d483421695d, introduced into `master` branch via merge of PR #126

PR #126, opened 2017-01-01
BugFix: DC gain for discrete-time systems

roryyorke:
Fix for #114.

An alternative design would be to have `LTI.dcgain` call either `self(0)` (ctime) or `self(1)` (dtime); `TransferFunction.__call__` could special case for s=0, so that `g(0)` and `g.dcgain(0)` return the same thing.  In this case `__call__` would need to be implemented for `StateSpace`.  In my opinion `evalfr` and `freqresp` should both use `__call__`, and that `horner` should be deprecated.

### DIFF
--- a/control/statesp.py
+++ b/control/statesp.py
@@ -612,11 +612,14 @@ inputs/outputs for feedback.")
             at the origin
         """
         try:
-            gain = np.asarray(self.D -
-                              self.C.dot(np.linalg.solve(self.A, self.B)))
+            if self.isctime():
+                gain = np.asarray(self.D -
+                                    self.C.dot(np.linalg.solve(self.A, self.B)))
+            else:
+                gain = self.horner(1)
         except LinAlgError:
-            # zero eigenvalue: singular matrix
-            return np.nan
+            # eigenvalue at DC
+            gain = np.tile(np.nan,(self.outputs,self.inputs))
         return np.squeeze(gain)
 
 
--- a/control/tests/statesp_test.py
+++ b/control/tests/statesp_test.py
@@ -228,7 +228,8 @@ class TestStateSpace(unittest.TestCase):
 
         assert sys1.dt == sys1_11.dt
 
-    def test_dcgain(self):
+    def test_dcgain_cont(self):
+        """Test DC gain for continuous-time state-space systems"""
         sys = StateSpace(-2.,6.,5.,0)
         np.testing.assert_equal(sys.dcgain(), 15.)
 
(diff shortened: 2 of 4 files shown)


roryyorke:
rebased on 32f13bc.  Wonder why coveralls giving these "changes unknown" results?

slivingston:
regarding the mysterious "changes unknown" from Coveralls, it might be a bug in Coveralls server side.
Similar behavior is reported in https://github.com/lemurheavy/coveralls-public/issues/351

control/statesp.py
```diff
@@ -612,14 +612,20 @@ def dcgain(self):
...
-            return np.nan
+            # eigenvalue at DC
+            # TODO: should this be of shape (outputs,inputs) ?
+            # TODO: inf rather than nan?
+            # TODO: could one not do a bit better, maybe by finding a least squares solution?
+            #       consider G(s) = diag(1/s, 1/(s+1)); DC gain is [[inf, 0],[0,1]]
+            gain = np.nan
         return np.squeeze(gain)
 
-
```
slivingston: blank line deletion here does not seem to be relevant to the other changes in this commit.

control/statesp.py
```diff
@@ -612,14 +612,20 @@ def dcgain(self):
...
+                                    self.C.dot(np.linalg.solve(self.A, self.B)))
+            else:
+                gain = self.horner(1)
         except LinAlgError:
-            # zero eigenvalue: singular matrix
-            return np.nan
+            # eigenvalue at DC
+            # TODO: should this be of shape (outputs,inputs) ?
+            # TODO: inf rather than nan?
+            # TODO: could one not do a bit better, maybe by finding a least squares solution?
```
slivingston: I think that these questions are good for discussion, but doing so in an issue or on the mailing list would be better than new TODO comments here.

roryyorke: Fair enough.  On reflection, I think the shape remark is correct, and the other two require discussion.  Shall I change this to `np.tile(np.nan,(self.outputs,self.inputs))` (with unit test), and open an issue on whether we can do better?  FWIW, after a bit of thought and some experimentation, I'm not sure how to do better than evaluating DC gain on a `minreal`ed version of each input-output pair.

slivingston:
It seems that Coveralls is still broken. I measured change in test coverage on Python 3.5.2. Results:

* Before your patch: 76.74% (687 missed statements; 2954 total)
* After your patch: 77.47% (667 missed statements; 2960 total)

as reported from

    nosetests --with-coverage --cover-html --cover-package=control

slivingston:
Your proposed change `StateSpace.dcgain` to return a matrix of `numpy.nan` values that has the same shape as the gain matrix is good. The documentation of `StateSpace.dcgain` should be updated for this.

This also constitutes API change, in particular change to the return type in certain cases. I am not aware of existing applications that would break. Furthermore, NaN spreads readily in a computation.

Therefore, I think that once my comment about documentation is addressed, this PR is good to go.

roryyorke:
I had planned to fix the StateSpace.dcgain docstring and the whitespace - I guess it's too late on this PR?

slivingston:
I moved forward with the merge because my remaining comments can be addressed separately with no more effort than addressing them here (in this PR) would require. Also I noticed that there are several other issues of indentation style in the statesp.py, so the issue here and in the other locations can be handled separately soon.

slivingston:
I updated the docstring in ee75d9aab7f36f565fb3ddb3c9699be6c2084173 on `master` branch.

Issue #81, opened 2016-03-21
tf2ss() does not seem to work for static gains


MauriceRienks:
Creating a state space object of a static gain using tf2ss results in the ValueError: A must be square.

For example: tf2ss([1.],[1.])

cwrowley:
I just verified this for `tf2ss` (using slycot)--definitely looks like a bug.  Note that there are two cases that may happen when you call `tf2ss`, depending on whether slycot is installed.  In fixing this bug, it would be good to add tests that test both versions (with and without slycot).

slivingston:
The case of static gains when Slycot is installed is now handled, given changes from <https://github.com/python-control/Slycot/pull/5>, which will be in the next release of Slycot.

slivingston:
fixed in PR #129, changes from which are now in `master` branch and will appear in the next release.

PR #129, opened 2017-01-04
BugFix: tf2ss now handles static gains

roryyorke:
Fix for #81.

### DIFF
--- a/control/statesp.py
+++ b/control/statesp.py
@@ -10,6 +10,7 @@ python-control library.
 
 # Python 3 compatibility (needs to go here)
 from __future__ import print_function
+from __future__ import division # for _convertToStateSpace
 
 """Copyright (c) 2010 by California Institute of Technology
 All rights reserved.
@@ -647,6 +648,7 @@ def _convertToStateSpace(sys, **kw):
     """
 
     from .xferfcn import TransferFunction
+    import itertools
     if isinstance(sys, StateSpace):
         if len(kw):
             raise TypeError("If sys is a StateSpace, _convertToStateSpace \
@@ -679,16 +681,27 @@ cannot take keywords.")
                 ssout[3][:sys.outputs, :states],
                 ssout[4], sys.dt)
         except ImportError:
-            # If slycot is not available, use signal.lti (SISO only)
-            if (sys.inputs != 1 or sys.outputs != 1):
-                raise TypeError("No support for MIMO without slycot")
+            # No Slycot.  Scipy tf->ss can't handle MIMO, but static
+            # MIMO is an easy special case we can check for here
+            maxn = max(max(len(n) for n in nrow)
+                       for nrow in sys.num)
+            maxd = max(max(len(d) for d in drow)
+                       for drow in sys.den)
+            if 1==maxn and 1==maxd:
+                D = empty((sys.outputs,sys.inputs),dtype=float)
+                for i,j in itertools.product(range(sys.outputs),range(sys.inputs)):
+                    D[i,j] = sys.num[i][j][0] / sys.den[i][j][0]
+                return StateSpace([], [], [], D, sys.dt)
+            else:
+                if (sys.inputs != 1 or sys.outputs != 1):
+                    raise TypeError("No support for MIMO without slycot")
 
-            # TODO: do we want to squeeze first and check dimenations?
-            # I think this will fail if num and den aren't 1-D after
-            # the squeeze
-            lti_sys = lti(squeeze(sys.num), squeeze(sys.den))
-            return StateSpace(lti_sys.A, lti_sys.B, lti_sys.C, lti_sys.D,
-                              sys.dt)
+                # TODO: do we want to squeeze first and check dimenations?
+                # I think this will fail if num and den aren't 1-D after
+                # the squeeze
+                lti_sys = lti(squeeze(sys.num), squeeze(sys.den))
+                return StateSpace(lti_sys.A, lti_sys.B, lti_sys.C, lti_sys.D,
+                                  sys.dt)
 
     elif isinstance(sys, (int, float, complex)):
         if "inputs" in kw:
--- a/control/tests/convert_test.py
+++ b/control/tests/convert_test.py
@@ -184,6 +184,28 @@ class TestConvert(unittest.TestCase):
         if (not slycot_check()):
             self.assertRaises(TypeError, control.tf2ss, tfcn)
 
+    def testTf2ssStaticSiso(self):
+        """Regression: tf2ss for SISO static gain"""
+        import control
+        gsiso = control.tf2ss(control.tf(23, 46))
+        self.assertEqual(0, gsiso.states)
+        self.assertEqual(1, gsiso.inputs)
+        self.assertEqual(1, gsiso.outputs)
+        # in all cases ratios are exactly representable, so assert_array_equal is fine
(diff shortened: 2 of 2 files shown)


slivingston:
Could you split `testTf2ssStatic` from https://github.com/python-control/python-control/pull/129/commits/a44305b1c05633fc7c26f86ea7b639f18185d55a#diff-34ce1aa9250b93553151bda70aa98a07 into two separate tests? Use `slycot_check` as the splitting boundary.

Following [my comment](https://github.com/python-control/python-control/pull/122#issuecomment-269883256) on PR #122, tests should not partially change when Slycot is present or not. Otherwise, debugging efficiency is negatively impacted.

roryyorke:
You mean using `@unittest.skipIf(not slycot_check(), "slycot not installed")` as elsewhere in the test suite?

Sure, will do.

roryyorke:
It looks like this *is* actually a bug in Slycot, as speculated in [1]; I can get SLICOT's `TD04AD` to convert a static TF.  I'll see if I can wrap my mind around f2py's syntax and fix this in Slycot.  In that case we can keep the unit tests in this PR, but the change to `statesp.py` will be different, and the code will be much simpler.

[1] https://github.com/roryyorke/python-control/blob/5c839c205dce8e072909509545266a466d4a2404/control/statesp.py#L669-L671

slivingston:
Because changes from <https://github.com/python-control/Slycot/pull/5> are now in Slycot, the only remaining case before we can close #81 is when Slycot is not installed.

@roryyorke Some of the changes from your original commit in this PR might be able to address the case of no `slycot`. In particular,
```python
D = empty((sys.outputs,sys.inputs),dtype=float)
for i,j in itertools.product(range(sys.outputs),range(sys.inputs)):
    D[i,j] = sys.num[i][j][0] / sys.den[i][j][0]
return StateSpace([], [], [], D, sys.dt)
```

roryyorke:
Is your idea to have three cases: (1) static gain (SISO or MIMO), (2) with-Slycot (any dynamic), (3) no-Slycot (SISO dynamic only) ?

We could add a case (0), namely empty systems (or this could be handled in (1)); this would let us sidestep the Slycot issues with empty systems you identified in [in the related Slycot PR](https://github.com/python-control/Slycot/pull/5#issuecomment-275310567).

slivingston:
Unless I am missing something, with Slycot, static gain does not need to be treated as a special case, thanks to your patches to Slycot.

However, when Slycot is not installed, static gain would be treated separately.

slivingston:
so, the cases are:
1. with Slycot
2. static gains, no Slycot
3. SISO, no Slycot

Before creating a special case to handle empty systems (as noticed during review of <https://github.com/python-control/Slycot/pull/5>), we might wait until there is a motivating application.

roryyorke:
CI won't pass without updated Slycot; it passes on my Ubuntu 14.04 x64 system, with Conda Python 3.5.2, both with Slycot [9a098d12](https://github.com/python-control/Slycot/commit/9a098d124cb31c083d26feae746a1f98e26d1a90), and without Slycot.

slivingston:
I also demonstrated success on your tests `testTf2ssStaticSiso` and `testTf2ssStaticMimo` on a GNU/Linux system using Python versions 3.5.2 and 2.7.9, each with and without the current tip of `master` branch Slycot, [001b45d437ae7f9e9f3ff0d82db339897d548cab](https://github.com/python-control/Slycot/commit/001b45d437ae7f9e9f3ff0d82db339897d548cab)

Tomorrow I will work on releasing the next version of Slycot.

control/tests/convert_test.py
```diff
@@ -184,6 +184,28 @@ def testConvertMIMO(self):
...
+        self.assertEqual(1, gsiso.inputs)
+        self.assertEqual(1, gsiso.outputs)
+        # in all cases ratios are exactly representable, so assert_array_equal is fine
+        np.testing.assert_array_equal([[0.5]], gsiso.D)
+
+    def testTf2ssStaticMimo(self):
+        """Regression: tf2ss for MIMO static gain"""
+        import control
+        # 2x3 TFM
+        gmimo = control.tf2ss(control.tf([[ [23],   [3],  [5] ], [ [-1],  [0.125],  [101.3] ]],
```
slivingston: How did you select these coefficients?

control/statesp.py
```diff
@@ -679,16 +681,26 @@ def _convertToStateSpace(sys, **kw):
...
-            if (sys.inputs != 1 or sys.outputs != 1):
-                raise TypeError("No support for MIMO without slycot")
-
-            # TODO: do we want to squeeze first and check dimenations?
-            # I think this will fail if num and den aren't 1-D after
-            # the squeeze
-            lti_sys = lti(squeeze(sys.num), squeeze(sys.den))
-            return StateSpace(lti_sys.A, lti_sys.B, lti_sys.C, lti_sys.D,
-                              sys.dt)
+            # No Slycot.  Scipy tf->ss can't handle MIMO, but static MIMO is an easy special case we can check for here
```
slivingston: Comment lines should have length no more than 79 char. [PEP 8 recommends 72](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) for the case of comments, but in my experience a uniform max line length is easier to follow.

control/statesp.py
```diff
@@ -679,16 +681,26 @@ def _convertToStateSpace(sys, **kw):
...
+            if 1==maxn and 1==maxd:
+                D = empty((sys.outputs,sys.inputs),dtype=float)
+                for i,j in itertools.product(range(sys.outputs),range(sys.inputs)):
+                    D[i,j] = sys.num[i][j][0] / sys.den[i][j][0]
+                return StateSpace([], [], [], D, sys.dt)
+            else:
+                if (sys.inputs != 1 or sys.outputs != 1):
+                    raise TypeError("No support for MIMO without slycot")
+
+                # TODO: do we want to squeeze first and check dimenations?
```
slivingston: Should "dimenations" should be "dimensions"?

slivingston:
As you might have noticed, a current Slycot release was made, and now the CI jobs for this PR are passing.

Aside from my request about line length and question about the static gains MIMO test case, changes here are good to go.

control/statesp.py
roryyorke: OK, will fix.  Sorry, it's pretty obvious now that I look at it.  Emacs fills this to 68 columns, which in this case doesn't waste an extra line:
```
            # No Slycot.  Scipy tf->ss can't handle MIMO, but static
            # MIMO is an easy special case we can check for here
```

control/tests/convert_test.py
roryyorke: It's an arbitrary selection of numbers over a (small) range of magnitudes and signs; in all cases the ratios are exactly representable with low precision (see d just below), so I could use np.assert_array_equal().

control/statesp.py
roryyorke: I can't take the credit for this one, the comment was there already ;).  Not sure it's applicable anymore: the check that sys.inputs and sys.outputs are both 1 implies that squeeze *will* bring num and den to 1-D?

slivingston: Yes, I agree. The comment is obsolete after the addition in commit c90a399a38babc7b496fe2f848d3ac9593b954f0 of the check that sys.inputs = sys.outputs = 1.

I am planning to review and perform maintenance on documentation and code comments, so I can delete this comment at that time.

PR #131, opened 2017-02-10
Removed matplotlib holds.

jed-frey:
I found this testing with matplotlib-2.0.

‘hold’ functionality deprecated.

The ‘hold’ keyword argument and all functions and methods related to it are deprecated, along with the ‘axes.hold’ rcParams entry. The behavior will remain consistent with the default hold=True state that has long been in place. Instead of using a function or keyword argument (hold=False) to change that behavior, explicitly clear the axes or figure as needed prior to subsequent plotting commands.

http://matplotlib.org/api/api_changes.html#hold-functionality-deprecated


slivingston:
thanks

PR #132, opened 2017-02-16
Add smart selection of gains for root locus plot. Calculation of break…

gonmolina:
…points in real axis.

### DIFF
--- a/control/rlocus.py
+++ b/control/rlocus.py
@@ -57,25 +57,24 @@ import numpy as np
 import pylab  # plotting routines
 import scipy.signal  # signal processing toolbox
 from scipy import array, poly1d, row_stack, zeros_like, real, imag
-
-from . import xferfcn
+from .xferfcn import _convertToTransferFunction
 from .exception import ControlMIMONotImplemented
 
 __all__ = ['root_locus', 'rlocus']
 
 
 # Main function: compute a root locus diagram
-def root_locus(dinsys, kvect=None, xlim=None, ylim=None, plotstr='-', plot=True,
-               print_gain=True):
+def root_locus(sys, kvect=None, xlim=None, ylim=None, plotstr='-', Plot=True,
+               PrintGain=True):
     """Root locus plot
 
     Calculate the root locus by finding the roots of 1+k*TF(s)
     where TF is self.num(s)/self.den(s) and each k is an element
-    of gvect.
+    of kvect.
 
     Parameters
     ----------
-    dinsys : LTI object
+    sys : LTI object
         Linear input/output systems (SISO only, for now)
     kvect : list or ndarray, optional
         List of gains to use in computing diagram
@@ -83,9 +82,9 @@ def root_locus(dinsys, kvect=None, xlim=None, ylim=None, plotstr='-', plot=True,
         control of x-axis range, normally with tuple (see matplotlib.axes)
     ylim : tuple or list, optional
         control of y-axis range
-    plot : boolean, optional (default = True)
+    Plot : boolean, optional (default = True)
         If True, plot magnitude and phase
-    print_gain: boolean (default = True)
+    PrintGain: boolean (default = True)
         If True, report mouse clicks when close to the root-locus branches,
         calculate gain, damping and print
     plotstr: string that declare of the rlocus (see matplotlib)
@@ -99,7 +98,7 @@ def root_locus(dinsys, kvect=None, xlim=None, ylim=None, plotstr='-', plot=True,
     """
 
     # Convert numerator and denominator to polynomials if they aren't
-    (nump, denp) = _systopoly1d(dinsys)
+    (nump, denp) = _systopoly1d(sys)
 
     if kvect is None:
         gvect = _default_gains(nump, denp)
@@ -107,7 +106,7 @@ def root_locus(dinsys, kvect=None, xlim=None, ylim=None, plotstr='-', plot=True,
         gvect = np.asarray(kvect)
 
     # Compute out the loci
-    mymat = _find_roots(dinsys, gvect)
+    mymat = _find_roots(sys, gvect)
     mymat = _sort_roots(mymat)
 
     # set smoothing tolerance
@@ -147,7 +146,7 @@ def root_locus(dinsys, kvect=None, xlim=None, ylim=None, plotstr='-', plot=True,
             # isolate poles in p1, p2
             if np.max(np.abs(p2 - p1)) > smtol:
                 newg = np.linspace(g1, g2, 5)
-                newmymat = _find_roots(dinsys, newg)
+                newmymat = _find_roots(sys, newg)
                 gvect = np.insert(gvect, i1 + 1, newg[1:4])
                 mymat = np.insert(mymat, i1 + 1, newmymat[1:4], axis=0)
                 mymat = _sort_roots(mymat)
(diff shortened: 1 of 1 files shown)


slivingston:
Is there any way to make your proposed features without breaking the public API?

As a matter of style, I agree with some of the changes, e.g., avoiding camelCase in `root_locus`, but I think that they are better done as part of a batch improvement of style that goes into a single version increment.

slivingston:
Regarding the comments near the top of rlocus.py where you list changes and the date, it is much better to place those notes about changes in your commit message because:
1. file history is studied through commit history;
2. changelogs in the files can become difficult to search compared to root CHANGELOG.

Some of the files have author, date, and change notes, but they are mostly vestigial and not consistently added. They might be deleted soon.

control/rlocus.py
```diff
@@ -43,43 +43,52 @@
...
-def root_locus(sys, kvect=None, xlim=None, ylim=None, plotstr='-', Plot=True,
-               PrintGain=True):
+def root_locus(dinsys, kvect=None, xlim=None, ylim=None, plotstr='-', plot=True,
+               print_gain=True):
     """Root locus plot
 
     Calculate the root locus by finding the roots of 1+k*TF(s)
     where TF is self.num(s)/self.den(s) and each k is an element
-    of kvect.
+    of gvect.
```
murrayrm: This should still be kvect, based on the call signature.

control/rlocus.py
```diff
@@ -43,43 +43,52 @@
...
 
     Parameters
     ----------
-    sys : LTI object
+    dinsys : LTI object
```
murrayrm: Is there a reason to use dinsys?  Most (all?) other functions use `sys` for the system argument.  What does `dinsys` stand for??

control/rlocus.py
```diff
@@ -179,36 +290,37 @@ def _RLFindRoots(sys, kvect):
...
-    if abs(K.real) > 1e-8 and abs(K.imag/K.real) < 0.04:
+    k = -1. / sys.horner(s)
+    if abs(k.real) > 1e-8 and abs(k.imag / k.real) < 0.04:
         print("Clicked at %10.4g%+10.4gj gain %10.4g damp %10.4g" %
-              (s.real, s.imag, K.real, -1*s.real/abs(s)))
+              (s.real, s.imag, k.real, -1 * s.real / abs(s)))
+
 
-rlocus = root_locus
+rlocus = root_locus
```
murrayrm: File should end in a new line, no?

control/rlocus.py
gonmolina: ok!

control/rlocus.py
gonmolina: I had problems while I was testing the code (maybe an import sys in the worng place). I will  rename it to sys and test again.

control/rlocus.py
gonmolina: ok

gonmolina:
@slivingston 
I have taken into account all comments you post and I have made the corrections. I am not programmer so probably there are several mistakes related to collaborate in a project (like changing the API). This is because it is the first time I try to participate in this kind of work. 
There are any guide to collaborate with this particular project ?   There are some automatic check that to do before the pull request?
I have plot several root locus and lines look nice using default gains calculation.

@murrayrm I made  suggested corrections.


Thanks a lot.

slivingston:
@gonmolina I will review your recent changes tomorrow.

Regarding some guide for collaboration on this project, there is not an explicit one. The general style guides for Python apply: [PEP 8](https://www.python.org/dev/peps/pep-0008) and [PEP 257](https://www.python.org/dev/peps/pep-0257). Many practices used for the NumPy project we also aspire to have here, so consider consulting [the NumPy contributing guide](https://github.com/numpy/numpy/blob/9bba99d822ea5dd589dfc936cb03597880e074d8/CONTRIBUTING.md). Some details, e.g., about NumPy project governance, are not relevant here. Another guide that is good to try to follow is [that of TuLiP](https://github.com/tulip-control/tulip-control/blob/238866b45a86117d89ddb32343e42f5739ab64ae/CONTRIBUTING.md).

Regarding automatic checks on the pull request (PR), tests are performed using Travis CI. You can find results from jobs for PRs at https://travis-ci.org/python-control/python-control/pull_requests

control/rlocus.py
```diff
@@ -43,43 +43,52 @@
...
 #
+# GDM, 16 February 2017: add smart selection of gains based on axis.
+#   * Add gains up to a tolerance is achieved
+#   * Change some variables and functions names ir order to improve "code style"
+#
+#
 # $Id$
 
 # Packages used by this module
+from functools import partial
```
slivingston: Why did you move the import of `partial` to here? Similarly, why were the imports of `scipy.signal` and `pylab` moved?

Are you changing the code to follow the style that groups imports according to standard library, other packages, etc.? If so, can you move such style-only changes to a dedicated commit? History is easier to understand that way.

control/rlocus.py
```diff
@@ -80,6 +88,7 @@ def root_locus(sys, kvect=None, xlim=None, ylim=None, plotstr='-', Plot=True,
     PrintGain: boolean (default = True)
         If True, report mouse clicks when close to the root-locus branches,
         calculate gain, damping and print
+    plotstr: string that declare of the rlocus (see matplotlib)
```
slivingston: Can you add details to this description? E.g.,

    format string for rendering the root loci. Interpretation of the string is defined by matplotlib.pyplot.plot.

control/rlocus.py
```diff
@@ -93,52 +102,164 @@ def root_locus(sys, kvect=None, xlim=None, ylim=None, plotstr='-', Plot=True,
...
+    if Plot:
         f = pylab.figure()
         if PrintGain:
             f.canvas.mpl_connect(
-                'button_release_event', partial(_RLFeedbackClicks, sys=sys))
+                'button_release_event', partial(_feedback_clicks, sys=sys))
         ax = pylab.axes()
 
-        # plot open loop poles
+        # Plot open loop poles
```
slivingston: I might be missing something, but it is not clear why you changed these "plot" words to uppercase "Plot". Is it merely a style change? If so, can you place it in a separate commit and indicate that it is a trivial change and only affects comments?

slivingston:
Several of the changes that I am requesting involve rebasing and modifying commits in this PR. You might already know how, but in case not, two useful beginning references are:
* https://help.github.com/articles/about-git-rebase/
* https://sethrobertson.github.io/GitBestPractices/

Alternatively, if you add commits such that the total diff of all combined commits in this PR is good, then I can just squash them into a single commit and merge that.

slivingston:
Is the code in this PR ported from Octave ?

slivingston:
I ask because Octave is under the GPL, and I do not want to GPL-pollute the `python-control` sourcetree.

E.g., compare with <https://sourceforge.net/p/octave/control/ci/tip/tree/inst/rlocus.m>

slivingston:
After you address my question about originality, can you provide a nontrivial demonstration of the changes? I played some with your `_default_gains`, but I would enjoy your recommendation about an example.

gonmolina:
@slivingston 
I have ported from Octave a part of this PR. I have also used some part of the code of evans.sci (siclab) in this PR. However, for simplicity, most of the pionts calculated would be the same that calculated in Octave. 

I just fix some problems I found in Octave. In Octave rlocus paths doesn't finished in zeros  (I add some extra points) and I change the way that sort the roots (I use the same that It was used in rlocus.py). Other differences are axis limits, and  asymptotes. 

About the license problems. I think right now **it is not original**, or it is hard to defend the originality. I didn't try to hide it (if you compare you will see that even the comments in part of the code are the same). 

I have some simple ideas to improve the algorithm, adding points in different positions that could result in fewer calculated points. The same for the finishing and adding points criteria.  However I don't know when the work begins to be original. I mean, the main idea would be the same: add points to the rlocus up to a criteria is satisfied (evans.sci and rlocus.m use the same finishing criteria, but the way they add points is quite different between one and another). My idea is to use the same way to add points but changing the creterias. Could be that considered original enough?
There is any rule to define when it is original or when is a copy?

slivingston:
Thanks for your explanation. It might have been better if I asked you whether it is derivative work, since that is the crucial question for copyright. Surely there is some originality in the process of porting.

The general guide about whether something is a derivative work is whether you depended critically on referring to most or all details of the original. E.g., did you copy-and-paste and then make necessary changes? Or, did you have to repeatedly compare with the original Octave implementation, line-by-line?

The copyright applies to the materials, not the ideas. So, if you provide a new implementation of an existing algorithm, all without detailed copying or porting of another existing implementation, then there should be no doubt about your new implementation being new (not derived).

gonmolina:
@slivingston 
Ok. Thanks for the comments. In that way, I think this PR have to be rejected! I will rewrite the code taking into consideration all comments you made and I will make a new PR. Thank you very much.

slivingston:
@gonmolina thanks. I will review your new PR soon.

slivingston:
for ease of future reference, note that the replacement PR is #138

Issue #95, opened 2016-06-01
Organize unit tests to follow the NumPy/SciPy guidelines


murrayrm:
Mentioned as part of PR #78: By the way, @murrayrm, it might be good at some point to organize the existing tests better, to follow the NumPy/SciPy guidelines, for instance so that tests for module foo.py are in test_foo.py. Many are already like this, but some modules do not have corresponding test modules.

slivingston:
For reference, the NumPy/SciPy Testing Guidelines are in the numpy repository at [doc/TESTS.rst.txt](https://github.com/numpy/numpy/blob/175476fec763984573e60485a447978cbcb381c0/doc/TESTS.rst.txt).

bnavigator:
The NumPy testing guidelines are now at https://numpy.org/doc/stable/reference/testing.html and replaced nose with pytest. (Nose is deprecated in general). Note that they primarily refer to NumPy/SciPy modules, which python-control is not.

With #438 this is mostly resolved.

PR #133, opened 2017-02-17
Add run of tests without Slycot on Travis CI

slivingston:
Because tests and parts of `control` that are used depend on whether Slycot is installed, it is worthwhile to exercise both cases as part of CI testing. This pull request does so. Test coverage is only for the case with Slycot installed (as from before this pull request).

The second commit disables tests that depend on Slycot if it is not found. Affected tests were introduced in PR #118.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ before_install:
   - conda install conda-build
   - conda config --add channels python-control
   - conda info -a
-  - conda create -q -n test-environment python="$TRAVIS_PYTHON_VERSION" pip coverage slycot
+  - conda create -q -n test-environment python="$TRAVIS_PYTHON_VERSION" pip coverage
   - source activate test-environment
   # coveralls not in conda repos
   - pip install coveralls
@@ -45,6 +45,11 @@ install:
 
 # command to run tests
 script:
+  # Before installing Slycot
+  - python setup.py test
+
+  # Now, get and use Slycot
+  - conda install slycot
   - coverage run setup.py test
 
 after_success:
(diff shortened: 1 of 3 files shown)


slivingston:
Changes here increase runtime for CI builds by approximately 90 seconds (as a sum across all configurations: Python 2.7, 3.3, 3.4, 3.5).

There is at least one negative consequence of changes here. Tests that are independent of Slycot are performed twice without potential benefit. However, the marginal gain in computation time and load is small.

roryyorke:
Looks good.  I assume you saw all the [deprecation warnings from scipy](https://travis-ci.org/python-control/python-control/jobs/202678373#L770).

FWIW, an alternative is to use Travis CI's matrix capability; I imagine "with Slycot" and "without Slycot" configurations.  Using a matrix setup opens other possibilities: 
  1) a "no-conda" config---I don't know how useful this is, but I gather from https://github.com/python-control/Slycot/issues/3 that this could be desired.
  2) Mac OS testing.  I haven't explored this (don't have a Mac), but I think Travis CI does support it.

murrayrm:
This PR looks good to me.  Will merge in 24 hours if there are no further comments.

murrayrm:
@slivingston: leaving it to you to delete the branch (just in case there is some reason to keep it).

slivingston:
@roryyorke and @murrayrm thanks to both of you for reviews.

re <https://github.com/python-control/python-control/pull/133#issuecomment-280749408>:

1. Yes, I noticed the deprecation warnings, but I think that it is outside the scope of this pull request. My plan is to add more tests while addressing issue #95 before making changes to avoid deprecated external API.

2. Before opening this PR, I tried testing on Mac OS as part of a matrix setup on Travis CI, e.g., <https://travis-ci.org/python-control/python-control/builds/201547383>. Ultimately I did not succeed. Because Slycot is the main difficulty in terms of cross-platform support, I decided to devote more effort to configuring Mac OS on Travis CI for Slycot instead of `control`.

PR #137, opened 2017-03-07
Update bdalg.py

Marxlp:
Extend the series and parallel to support more parameters

### DIFF
--- a/control/bdalg.py
+++ b/control/bdalg.py
@@ -60,14 +60,16 @@ from . import frdata as frd
 
 __all__ = ['series', 'parallel', 'negate', 'feedback', 'append', 'connect']
 
-def series(sys1, sys2):
-    """Return the series connection sys2 * sys1 for --> sys1 --> sys2 -->.
+def series(sys1, sys2,*sysn):
+    """Return the series connection (... * sys3 *) sys2 * sys1 
+    for (... sys3 -->) sys1 --> sys2 -->.
 
     Parameters
     ----------
     sys1: scalar, StateSpace, TransferFunction, or FRD
     sys2: scalar, StateSpace, TransferFunction, or FRD
-
+    sysn: scalar, StateSpace, TransferFunction, or FRD
+    
     Returns
     -------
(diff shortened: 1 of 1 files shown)


control/bdalg.py
```diff
@@ -60,14 +60,16 @@
 
 __all__ = ['series', 'parallel', 'negate', 'feedback', 'append', 'connect']
 
-def series(sys1, sys2):
-    """Return the series connection sys2 * sys1 for --> sys1 --> sys2 -->.
+def series(sys1, sys2,*sysn):
+    """Return the series connection (... * sys3 *) sys2 * sys1
```
roryyorke: How about
```
def series(sys1, *sysn):
    from functools import reduce
    return reduce(lambda x, y: y*x, sysn, sys1)
```
and similarly for parallel?

One downfall there is that `series(s1)` would return the `s1` instance, and not a copy of it.

Marxlp: Yes, it's  a good idea. But, using functional programming would increase the complexity of the code or decrease the elegance of the code.

slivingston:
@Marxlp are you going to add unit tests?

slivingston:
regarding code style, you must use 4-space indentation units.

control/bdalg.py
slivingston: Depending on one's experience, the use of `reduce` and `lambda` is elegant. Given the simplicity of the operation, the more important question here is performance. Sometimes functional idioms like the one suggested by @roryyorke can be better optimized by the Python compiler than a for-loop.

cwrowley: +1 for `reduce` and `lambda`

murrayrm:
Submitted [pull request](https://github.com/Marxlp/python-control/pull/1) to @Marxlp fixing the comments above.  Will wait a few days for update.  If not, will submit separate PR and merge to master.

Marxlp:
I opened a new pull request.

murrayrm:
For posterity: new pull request is #178.

PR #138, opened 2017-03-21
add unsupervised calculation on gains for root locus plot.

gonmolina:
Add gains up to a tolerance is achived.
@slivingston I wait for reviews. I have a file with some examples that are an exersice taken from Franklyn book. Let me know if you need it to test the algorithm. Do I have to put in a particular place? Where? Thanks in advance.

### DIFF
--- a/control/rlocus.py
+++ b/control/rlocus.py
@@ -227,10 +227,8 @@ def _k_max(num, den, real_break_points, k_break_points):
         kmax_asymp = -den(farthest_points) / num(farthest_points)
     else:
         kmax_asymp = [den.coeffs[0] / num.coeffs[0] * 3]
-    if np.max(kmax_asymp) == 0:
-        kmax = np.max(np.concatenate((np.real(kmax_asymp), k_break_points), axis=0))
-    else:
-        kmax = kmax_asymp
+
+    kmax = np.max(np.concatenate((np.real(kmax_asymp), k_break_points), axis=0))
     return kmax
 
 
--- a/control/rlocus.py
+++ b/control/rlocus.py
@@ -191,6 +191,7 @@ def _default_gains(num, den, xlim, ylim):
     else:
         y_tolerance = 0.05 * (ylim[1] - ylim[0])
 
+
     tolerance = np.max([x_tolerance, y_tolerance])
     distance_points = np.abs(np.diff(mymat, axis=0))
     indexes_too_far = np.where(distance_points > tolerance)
@@ -206,7 +207,7 @@ def _default_gains(num, den, xlim, ylim):
         distance_points = np.abs(np.diff(mymat, axis=0)) > tolerance  # distance between points
         indexes_too_far = np.where(distance_points)
 
-    new_gains = np.hstack((np.linspace(kvect[-1], kvect[-1]*200, 10)))
+    new_gains = np.hstack((np.linspace(kvect[-1], kvect[-1]*1e10, 10)))
     new_points = _RLFindRoots(num, den, new_gains[1:10])
     kvect = np.append(kvect, new_gains[1:10])
     mymat = np.concatenate((mymat, new_points), axis=0)
--- a/control/rlocus.py
+++ b/control/rlocus.py
@@ -172,7 +172,7 @@ def _default_gains(num, den, xlim, ylim):
         mymat = _RLSortRoots(mymat)
         distance_points = np.abs(np.diff(mymat, axis=0))
         indexes_too_far = np.where(distance_points > tolerance)
-    new_gains = np.hstack((np.logspace(np.log10(kvect[-1]), np.log10(kvect[-1]*200), 10)))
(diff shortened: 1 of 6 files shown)


slivingston:
@gonmolina can you add a citation in the docstring to the book that you used?

slivingston:
I am sorry for my delay; I lost track of my GitHub notifications feed. I anticipate finishing reviewing this within 2 days.

gonmolina:
I added a book as a reference. 
I added possibility to draw a grid on s-plane similar to matlab (argument grid = True).
I have tested a lot this function and works fine. 
I have wrote a script with several tests with a lot of cases and root locus figures seem to be fine. If you think this script can help you to test the algorithm I can add it to the repository (please tell me where is the right place to add it). 
I have added support of 0 degrees root locus (positive feedback or  k<0) in the algorithm. This works fine in all cases that I have tested, but I couldn't solve when the num.order=den.order. To solve this problem a complete new solution have to be written, mainly when lines are plotted and when roots are sorted. I raise an error when a root locus of this kind is demanded. 

 @slivingston I wait for your review.
Thanks in advance.
Gonzalo

slivingston:
@gonmolina thanks working on it now.

control/rlocus.py
```diff
@@ -127,14 +137,137 @@ def root_locus(sys, kvect=None, xlim=None, ylim=None, plotstr='-', Plot=True,
...
+    if axmax != axmin:
+        deltax = (axmax - axmin) * 0.02
+    else:
+        deltax = np.max([1., axmax / 2])
+    axlim = [axmin - deltax, axmax + deltax]
+    return axlim
+
+
+def _k_max(num, den, real_break_points, k_break_points):
+    """" Calculation the maximum gain for the root locus shown in tne figure"""
```
slivingston: misprint: "tne" should be "the". Also, the first word should some verb. E.g., "Calculate"

control/rlocus.py
```diff
@@ -157,29 +290,33 @@ def _systopoly1d(sys):
...
 
     # Convert numerator and denominator to polynomials if they aren't
-    (nump, denp) = _systopoly1d(sys)
-
     roots = []
     for k in kvect:
         curpoly = denp + k * nump
         curroots = curpoly.r
+        if len(curroots) < denp.order:
+            curroots = np.insert(curroots, len(curroots), np.inf)  # if i have less poles than open loop is because  i
```
slivingston: The comment here overflows into the next line. For ease of reading, it should be moved to the line before. This style for block comments is given [in PEP 8](https://www.python.org/dev/peps/pep-0008/#block-comments). E.g., after I edit the comment text a little,
```python
if len(curroots) < denp.order:
    # If I have fewer poles than open loop, it is because I
    # have one at infinity.
    curroots = np.insert(curroots, len(curroots), np.inf)
```

slivingston:
@gonmolina thanks. The changes are good.

slivingston:
Actually considering the failures on CI testing, there may have been a regression. I am trying to reproduce locally.

gonmolina:
Is there any way to do the test locally in my machine without doing a push? 
I think I can fix it....

slivingston:
There are several new changes outside the scope of the previous review, so I will review these when you tell me that they are ready.

Several of the commit messages mention fixing of bugs. If these existed before this PR, can you add regression tests for them, i.e., tests that demonstrate the bugs?

gonmolina:
I made 2 changes:
- I added line 268 to ensure kmax> 0 when all the singular points are in 0. Problems getting kvect when deal with systems of the form 1/s^n. I do not think this is the problem while building.
- I greatly increased the gain after the kmax in line 209 so that the lines of the  roots locus reach the zeros (in the figure) for all cases. I think this can be the problem since several groups of equal roots turn out to be about the same for too large gains. Solution: fewer points after kmax, smaller maximum gain, and separate them logarithmically. This problem was clear when I was trying to solve 0 degrees root locus with for systems with equal number of zeros and poles.

A regression test for the first problem is:

import control as ctrl
sys1=ctrl.tf([1],[1,0,0])
r,k = ctrl.rlocus(sys1, grid=True)

slivingston:
@gonmolina it looks like you are permitting me to push changes to the branch associated with this pull request. Can I do so, to add the regression test that [you suggest above](https://github.com/python-control/python-control/pull/138#issuecomment-312406694)?

gonmolina:
yes, you can do it.

murrayrm:
@slivingston I'm trying to clean up some of the PRs over the holiday break.  Let me know if you are still working on this one.

murrayrm:
I've run Travis CI on a local branch and all tests are OK: [results](https://travis-ci.org/murrayrm/python-control/builds/324465228).

@gonmolina, @slivingston Are we ready to merge or is more testing required?  Seems like this has to be better than what was there before (placeholder).

PR #140, opened 2017-03-26
Fix minreal not returning a discrete TF

rssalessio:
In case we want to use minreal on a discrete TF the returning TF is continuous and not discrete. This should fix the problem

### DIFF
--- a/control/xferfcn.py
+++ b/control/xferfcn.py
@@ -685,7 +685,7 @@ only implemented for SISO functions.")
                 den[i][j] = np.atleast_1d(real(poly(poles)))
 
         # end result
-        return TransferFunction(num, den, None if self.dt == None else self.dt)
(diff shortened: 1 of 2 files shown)


control/xferfcn.py
```diff
@@ -685,7 +685,7 @@ def minreal(self, tol=None):
                 den[i][j] = np.atleast_1d(real(poly(poles)))
 
         # end result
-        return TransferFunction(num, den)
+        return TransferFunction(num, den, None if self.dt == None else self.dt)
```
roryyorke: Couldn't you use `self.dt` instead of `None if self.dt == None else self.dt`?

rssalessio: Yea sure :D. Don't know what I was thinking in that moment

roryyorke:
Looks good to me, thanks.

Issue #142, opened 2017-04-10
Step response output matrix not consistent between continuous-time and discrete-time


gpannocchia:
Dear Python-control developers,

When I use the step response function (control.step) on a continuous-time multivariable system the output response matrix has dimension (nsteps, noutputs), whereas the same function used on discrete-time multivariable system returns an output response matrix of dimension (noutputs, nsteps). 

Here is a sample code, which highlights the behavior. 

*********************** STARTS HERE
import control as cnt
import numpy as np

# Continuous-time system matrices
A = [[-1., 0.5],[0,-2.]]
B = np.eye(2)
C = np.eye(2)
D = np.zeros([2,2])
sysc = cnt.ss(A,B,C,D)

# Discretization
ts = 0.1
sysd = cnt.sample_system(sysc,ts)

# Step responses
tfin = 10.
Time = np.linspace(0,10,int(tfin/ts)+1)
YoutC, TimeC = cnt.step(sysc, T=Time)
YoutD, TimeD = cnt.step(sysd, T=Time)

print "Version of python-control", cnt.__version__
print "Output response matrix dimension (Continuous-Time) =", YoutC.shape
print "Output response matrix dimension (Discrete-Time) =", YoutD.shape
*********************** ENDS HERE

It returns (I am using python 2.7 under mac-osx):

Version of python-control 0.7.0
Output response matrix dimension (Continuous-Time) = (101, 2)
Output response matrix dimension (Discrete-Time) = (2, 101)

Thank you. Best regards,

	Gabriele

murrayrm:
I tried to recreate this error in the master branch and note the following differences:

- In the current version of `python-control`, the `step` function is actually `step_response`
- In the current version of `python-control`, the `step_response` function only handles SIMO systems (and warning message is generated and the second input is ignored)

The behavior above can be recreated using the `MATLAB` compatibility module, `control.matlab`, and the discrepancy in return values is as described above.

murrayrm:
Tracked down the problem: in `control/timeresp.py`, line 324, a call is made to the `scipy.signal` library for discrete time simulation.  Here is the code around that point:

    # Discrete time simulation using signal processing toolbox
    dsys = (A, B, C, D, sys.dt)
    tout, yout, xout = sp.signal.dlsim(dsys, U, T, X0)

[...]

PR #143, opened 2017-04-15
Make output response matrices consistent between continuous and discrete time

murrayrm:
This PR addresses issue #142 where the output response matrices were not consistent for continuous and discrete time simulations.  The error was in the discrete time simulation, where the conventions used by `spicy.signal` are different than those used by `python-control`.  Also added a unit test that catches this error.


murrayrm:
Thanks for the review, @roryyorke.  Merging into master.

PR #145, opened 2017-06-09
BUG: allow state-space to transfer function conversion of static gains

roryyorke:


slivingston:
is there an issue associated with this PR? If not, it is not necessary to create one now. I only ask to ensure that relevant issues are noted here.

roryyorke:
There's no associated issue.

PR #146, opened 2017-06-10
Updates to margin calculation

repagh:
Which margin to consider in margin function is a tricky issue. Have revisited this with the help of 

        # "A note on the Gain and Phase Margin Concepts
        # Journal of Control and Systems Engineering, Yazdan Bavafi-Toosi,
        # Dec 2015, vol 3 iss 1, pp 51-59

a very illustrative paper, by the way. Most results now correspond to Matlab, edge cases (gain 1, but not crossing that, e.g.) give slightly different results, but not unreasonable; for a type 0 sys with max gain 1, Matlab will give you phase margin -180, at w=0, and python-control now gives inf / nan.

### DIFF
--- a/control/margins.py
+++ b/control/margins.py
@@ -87,7 +87,7 @@ def _polysqr(pol):
 # RvP, July 8, 2015, augmented to calculate all phase/gain crossings with
 #                    frd data. Correct to return smallest phase
 #                    margin, smallest gain margin and their frequencies
-def stability_margins(sysdata, returnall=False, epsw=1e-8):
+def stability_margins(sysdata, returnall=False, epsw=0.0):
     """Calculate stability margins and associated crossover frequencies.
 
     Parameters
@@ -157,16 +157,15 @@ def stability_margins(sysdata, returnall=False, epsw=1e-8):
 
         # first remove imaginary and negative frequencies, epsw removes the
         # "0" frequency for type-2 systems
-        w_180 = np.real(w_180[(np.imag(w_180) == 0) * (w_180 >= epsw)])
+        w_180 = np.real(w_180[(np.imag(w_180) == 0) * (w_180 > epsw)])
         #print ('2:w_180', w_180)
 
         # evaluate response at remaining frequencies, to test for phase 180 vs 0
         resp_w_180 = np.real(np.polyval(sys.num[0][0], 1.j*w_180) /
                              np.polyval(sys.den[0][0], 1.j*w_180))
-        #print ('resp_w_180', resp_w_180)                     
-
+        
         # only keep frequencies where the negative real axis is crossed
-        w_180 = w_180[np.real(resp_w_180) < 0.0]
+        w_180 = w_180[np.real(resp_w_180) <= 0.0]
 
         # and sort
         w_180.sort()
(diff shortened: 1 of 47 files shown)


slivingston:
@repagh Because you requested review from @cwrowley but then merged without receiving review, was merging an accident?

repagh:
Dunno exactly how that happened. I had this very small & easy fix to the discrete impulse response that I made a second merge request for and really did not think it was a problem merging myself. So I merged that, and in some way both requests got merged.

slivingston:
one handy switch for `git push` is `-n` or `--dry-run`, which shows what would happen if the changes were pushed but does not perform the changes. I often use it as a quick preflight check.

control/margins.py
```diff
@@ -253,20 +265,30 @@ def dstab(w):
...
+    
     if returnall:
         return GM, PM, SM, w_180, wc, wstab
     else:
+        if GM.shape[0] and not np.isinf(GM).all():
+            with np.errstate(all='ignore'):
+                gmidx = np.where(np.abs(np.log(GM)) == 
+                                 np.min(np.abs(np.log(GM))))
+        else:
+            gmidx = -1
```
roryyorke: I'd have used "None" here; the reader could interpret -1 as "last element".

control/margins.py
```diff
@@ -253,20 +265,30 @@ def dstab(w):
...
+        if PM.shape[0]:
+            pmidx = np.where(np.abs(PM) == np.amin(np.abs(PM)))[0]
         return (
-            (GM.shape[0] or None) and np.amin(GM),
-            (PM.shape[0] or None) and np.amin(PM),
-            (SM.shape[0] or None) and np.amin(SM),
-            (w_180.shape[0] or None) and w_180[GM==np.amin(GM)][0],
-            (wc.shape[0] or None) and wc[PM==np.amin(PM)][0],
-            (wstab.shape[0] or None) and wstab[SM==np.amin(SM)][0])
+            (not gmidx != -1 and float('inf')) or GM[gmidx][0],
```
roryyorke: How about ```GM[gmidx][0] if gmidx!=-1 else np.inf```? (etc.)

control/tests/margin_test.py
```diff
@@ -11,6 +11,22 @@
 from control.statesp import StateSpace
 from control.margins import *
 
+def assert_array_almost_equal(x, y, ndigit=4):
```
roryyorke: A comment indicating why you need this would be good.  I thought perhaps it was because nan's compare unequal, but np.testing.assert_array_almost_equal seems to handle nans fine (i.e., considers them to be equal).

control/tests/margin_test.py
```diff
@@ -11,6 +11,22 @@
...
+
+    x = np.array(x)
```
roryyorke: nitpick: ```np.asarray```?

control/tests/margin_test.py
```diff
@@ -11,6 +11,22 @@
...
+    y = np.array(y)
+    try:
+        if np.isfinite(x).any() and \
+           np.equal(np.isfinite(x), np.isfinite(y)).all() and \
+           np.equal(np.isnan(x), np.isnan(y)).all():
+            np.testing.assert_array_almost_equal(
+                x[np.isfinite(x)], y[np.isfinite(y)], ndigit)
```
roryyorke: this would, if I read correctly, consider np.inf and -np.inf to be equal, as long as one element of the arrays was finite.  Is that the intent?

control/tests/margin_test.py
```diff
@@ -153,17 +267,48 @@ def test_nocross(self):
         h2 = 3*(10+s)/(2+s)
         h3 = 0.01*(10-s)/(2+s)/(1+s)
         gm, pm, wm, wg, wp, ws = stability_margins(h1)
-        self.assertEqual(gm, None)
-        self.assertEqual(wg, None)
+        assert_array_almost_equal(
+            [gm, pm, wg, wp],
+            [float('Inf'), float('Inf'), float('NaN'), float('NaN')])
```
roryyorke: ```np.inf```, ```np.nan``` ?

control/tests/margin_test.py
```diff
@@ -153,17 +267,48 @@ def test_nocross(self):
...
         gm, pm, wm, wg, wp, ws = stability_margins(h3)
-        self.assertEqual(pm, None)
+        self.assertTrue(np.isnan(wp))
         omega = np.logspace(-2,2, 100)
         out1b = stability_margins(FRD(h1, omega))
         out2b = stability_margins(FRD(h2, omega))
         out3b = stability_margins(FRD(h3, omega))
 
+    def test_zmore_margin(self):
+        print("""
```
roryyorke: do we really want to print out info like this during tests?

roryyorke:
If I understand the paper correctly, it defines GM and PM as the smallest absolute change (log-scaled, for GM) margin that results in a stability change.  GM > 0 means gain should be increased for stabilty change, GM < 0 means it should be decreased, and similarly for PM.  These definitions sound reasonable to me, but I'm not an academic.

Do we expect ```stability_margins``` to give margins for *unstable* systems (that is, change in gain or phase that would make the system stable) ?

repagh:
seems useful to me, and Matlab's margin does it too, so why not.

PR #151, opened 2017-07-01
H-infinity mixed-synthesis

roryyorke:
This adds methods "augw" and "mixsyn", with interfaces similar to the Matlab and Octave functions of the same name.  They are for so-called mixed-sensitivity design, in which the sensitivity, complementary sensitivity, and "input sensitivity" (K*S, or reference to input mapping) can be weighted and stacked for H-infinity synthesis.

As part of this, I've fixed what I think is a bug, namely that empty StateSpace objects could not be created.  Such empty objects are convenient in the augw implementation.

I've added two examples for mixsyn, one SISO and one MIMO.

augw is the major part of the new code.  It may be that there is some better way to implement it, but I have tried to write it in a way that no uncontrollable or unobservable modes are introduced, and so that is still (somewhat) clear.

### DIFF
--- a/control/robust.py
+++ b/control/robust.py
@@ -171,3 +171,146 @@ def hinfsyn(P,nmeas,ncon):
     CL = StateSpace(Ac, Bc, Cc, Dc)
 
     return K, CL, gam, rcond
+
+
+def _size_as_needed(w,wname,n):
+    """_size_as_needed(w,wname,n) -> w2
+    If w is scalar, reshape to nxn; otherwise check size is OK.  Return w as StateSpace object.
+    If w is None, return empty StateSpace object.
+    """
+    from . import append, ss
+    if w is not None:
+        if not isinstance(w,StateSpace):
+            w = ss(w)
+        if 1==w.inputs and 1==w.outputs:
+            w = append(*(w,)*n)
+        else:
+            if w.inputs != n:
+                msg=("{}: weighting function has {} inputs, expected {}".
+                      format(wname,w.inputs,n))
+                raise ValueError(msg)
+    else:
+        w = ss([],[],[],[])
+    return w
+
+
+def augw(g,w1=None,w2=None,w3=None):
+    """augw(g,w1=None,w2=None,w3=None) -> p
+    Augment plant for mixed sensitivity problem
+     g - LTI object, ny-by-nu
+     w1 - weighting on S; None, scalar, or k1-by-ny LTI object
+     w2 - weighting on KS; None, scalar, or k2-by-nu LTI object
+     w3 - weighting on T; None, scalar, or k3-by-ny LTI object
+     p - augmented plant; StateSpace object
+    If a weighting is None, no augmentation is done for it.  At least
+    one weighting must not be None.
+    If a weighting w is scalar, it will be replaced by I*w, where I is
+    ny-by-ny for w1 and w3, and nu-by-nu for w2.
+
+    See also hinfsyn, mixsyn
+    """
+    from . import append, ss, connect
+
+    if w1 is None and w2 is None and w3 is None:
+        raise ValueError("At least one weighting must not be None")
+    ny = g.outputs
+    nu = g.inputs
+
+    w1,w2,w3 = [_size_as_needed(w,wname,n)
+                for w,wname,n in zip((w1,w2,w3),
+                                     ('w1','w2','w3'),
+                                     (ny,nu,ny))]
+
+    if not isinstance(g,StateSpace):
+        g = ss(g)
+
+    #       w         u
+    #  z1 [ w1   |   -w1*g  ]
+    #  z2 [ 0    |    w2    ]
+    #  z3 [ 0    |    w3*g  ]
+    #     [------+--------- ]
+    #  v  [ I    |    -g    ]
+
+    # error summer: inputs are -y and r=w
+    Ie = ss([],[],[],np.eye(ny))
+    # control: needed to "distribute" control input
+    Iu = ss([],[],[],np.eye(nu))
+
+    sysall = append(w1,w2,w3,Ie,g,Iu)
+
+    niw1 = w1.inputs
+    niw2 = w2.inputs
+    niw3 = w3.inputs
+
+    now1 = w1.outputs
+    now2 = w2.outputs
+    now3 = w3.outputs
+
+    q = np.zeros((niw1+niw2+niw3+ny+nu,2))
+    q[:,0] = np.arange(1,q.shape[0]+1)
+
+    # Ie -> w1
+    q[:niw1,1] = np.arange(1+now1+now2+now3,
+                           1+now1+now2+now3+niw1)
+
+    # Iu -> w2
+    q[niw1:niw1+niw2,1] = np.arange(1+now1+now2+now3+2*ny,
+                                    1+now1+now2+now3+2*ny+niw2)
+
+    # y -> w3
+    q[niw1+niw2:niw1+niw2+niw3,1] = np.arange(1+now1+now2+now3+ny,
+                                              1+now1+now2+now3+ny+niw3)
+
+    # -y -> Iy; note the leading -
+    q[niw1+niw2+niw3:niw1+niw2+niw3+ny,1] = -np.arange(1+now1+now2+now3+ny,
+                                                       1+now1+now2+now3+2*ny)
+
+    # Iu -> G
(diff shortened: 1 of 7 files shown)


murrayrm:
@roryyorke There's a small conflict with master that it would be good to fix.  Will try to do a more thorough review in the coming days...

roryyorke:
Will have a look at the conflict.

control/robust.py
```diff
@@ -171,3 +171,146 @@ def hinfsyn(P,nmeas,ncon):
     CL = StateSpace(Ac, Bc, Cc, Dc)
 
     return K, CL, gam, rcond
+
+
+def _size_as_needed(w,wname,n):
+    """_size_as_needed(w,wname,n) -> w2
```
murrayrm: Nonstandard docstring.  I don't think we use this type of docstring (fcn(args) -> output) anyplace else.  In addition, parameters, returns, etc are documented in a different format.  Would prefer to use a consistent style, as done in `hinfsyn` in this file.

control/robust.py
```diff
@@ -171,3 +171,146 @@ def hinfsyn(P,nmeas,ncon):
...
+                msg=("{}: weighting function has {} inputs, expected {}".
+                      format(wname,w.inputs,n))
+                raise ValueError(msg)
+    else:
+        w = ss([],[],[],[])
+    return w
+
+
+def augw(g,w1=None,w2=None,w3=None):
+    """augw(g,w1=None,w2=None,w3=None) -> p
```
murrayrm: Nonstandard docstring.

control/robust.py
```diff
@@ -171,3 +171,146 @@ def hinfsyn(P,nmeas,ncon):
...
+    g: LTI; the plant for which controller must be synthesized
+    w1: weighting on s = (1+g*k)**-1; None, or scalar or k1-by-ny LTI
+    w2: weighting on k*s; None, or scalar or k2-by-nu LTI
+    w3: weighting on t = g*k*(1+g*k)**-1; None, or scalar or k3-by-ny LTI
+    At least one of w1, w2, and w3 must not be None.
+    k: synthesized controller; StateSpace object
+    cl: closed system mapping evaluation inputs to evaluation outputs; if p is the augmented plant, with
+        [z] = [p11 p12] [w], then cl is the system from w->z with u=-k*y.  StateSpace object.
+        [y]   [p21   g] [u]
+    info: namedtuple with fields, in order,
```
murrayrm: This is the first use of a `namedtuple` in `python-control`.  Is there a more standard way to return this sort of information in `numpy` or `scipy`?  If so, perhaps we use that format?

control/statesp.py
```diff
@@ -54,7 +54,7 @@
 import math
 import numpy as np
 from numpy import all, angle, any, array, asarray, concatenate, cos, delete, \
-    dot, empty, exp, eye, matrix, ones, poly, poly1d, roots, shape, sin, \
+    dot, empty, exp, eye, matrix, ones, pi, poly, poly1d, roots, shape, sin, \
```
murrayrm: We should avoid using `numpy.pi` since this can generate errors when `numpy` is incorporated as a `mock` function.  See [PR 171, commit b881b7](https://github.com/python-control/python-control/pull/171/files/b881b77e485be3c7d90a56a0db8d190b56cdf8c1).  Replace with `math.pi`.

control/statesp.py
```diff
@@ -69,6 +69,19 @@
 
 __all__ = ['StateSpace', 'ss', 'rss', 'drss', 'tf2ss', 'ssdata']
 
+
+def _matrix(a):
+    """_matrix(a) -> numpy.matrix
+    a - passed to numpy.matrix
+    Wrapper around numpy.matrix; unlike that function,  _matrix([]) will be 0x0
```
murrayrm: Not clear to my why this is useful?   Can you say a bit here (or in a comment in the code)?

control/robust.py
roryyorke: Will fix here and elsewhere (augw, _matrix, ...).  I see I also used this in xferfcn.py in 310f5805 in `_dc_gain_cont`.

control/robust.py
roryyorke: `namedtuple` has been around since Python 2.6, so it's hardly non-standard ;).  I couldn't find namedtuple use in numpy, but it is used in scipy (stats/stats.py, for example).

I can change this to a plain tuple; if we do keep the namedtuple, I'd like to remove the leading underscore from the name.

control/statesp.py
roryyorke: unqualified `pi` no longer used following your aa6e0dfc anyway.  Will remove.

control/statesp.py
roryyorke: I'll expand the commentary; it's just a helper to make it easier to get 0x0 matrices, which are needed for empty state space systems.

roryyorke:
I thought I'd check the docstrings via sphinx, but when I run `make html` in the `doc` sub-directory, I get
```
  File "/home/rory/miniconda3/envs/control-python27/lib/python2.7/re.py", line 155, in sub
    return _compile(pattern, flags).sub(repl, string, count)
TypeError: expected string or buffer
```

I get the same error on master.

I see `doc-requirements.txt` references Sphinx from github, but that is somewhat older (log talks about Sphinx 1.3.1, I've got 1.6 installed from conda).

This is my conda environment (2.7-based, and AFAIK up-to-date; don't know why I don't have Scipy 1.0):
```
alabaster                 0.7.10                   py27_0  
babel                     2.5.0                    py27_0  
cairo                     1.14.8                        0  
certifi                   2016.2.28                py27_0  
cycler                    0.10.0                   py27_0  
dbus                      1.10.20                       0  
docutils                  0.14                     py27_0  
expat                     2.1.0                         0  
fontconfig                2.12.1                        3  
freetype                  2.5.5                         2  
funcsigs                  1.0.2                    py27_0  
functools32               3.2.3.2                  py27_0  
glib                      2.50.2                        1  
gst-plugins-base          1.8.0                         0  
gstreamer                 1.8.0                         0  
icu                       54.1                          0  
imagesize                 0.7.1                    py27_0  
jinja2                    2.9.6                    py27_0  
jpeg                      9b                            0  
libffi                    3.2.1                         1  
libgcc                    5.2.0                         0  
libgfortran               3.0.0                         1  
libiconv                  1.14                          0  
libpng                    1.6.30                        1  
libxcb                    1.12                          1  
libxml2                   2.9.4                         0  
markupsafe                1.0                      py27_0  
matplotlib                2.0.2               np113py27_0  
mkl                       2017.0.3                      0  
mock                      2.0.0                    py27_0  
nose                      1.3.7                    py27_1  
numpy                     1.13.1                   py27_0  
numpydoc                  0.7.0                    py27_0  
openssl                   1.0.2l                        0  
pbr                       1.10.0                   py27_0  
pcre                      8.39                          1  
pip                       9.0.1                    py27_1  
pixman                    0.34.0                        0  
pycairo                   1.10.0                   py27_0  
pygments                  2.2.0                    py27_0  
pyparsing                 2.2.0                    py27_0  
pyqt                      5.6.0                    py27_2  
python                    2.7.13                        0  
python-dateutil           2.6.1                    py27_0  
pytz                      2017.2                   py27_0  
qt                        5.6.2                         5  
readline                  6.2                           2  
requests                  2.14.2                   py27_0  
scipy                     0.19.1              np113py27_0  
setuptools                36.4.0                   py27_1  
sip                       4.18                     py27_0  
six                       1.10.0                   py27_0  
snowballstemmer           1.2.1                    py27_0  
sphinx                    1.6.3                    py27_0  
sphinxcontrib             1.0                      py27_0  
sphinxcontrib-websupport  1.0.1                    py27_0  
sqlite                    3.13.0                        0  
subprocess32              3.2.7                    py27_0  
tk                        8.5.18                        0  
typing                    3.6.2                    py27_0  
wheel                     0.29.0                   py27_0  
zlib                      1.2.11                        0  
```

murrayrm:
@roryyorke I had the same problem with sphinx on my local machine.  For some reason it seems to work on readthedocs.org.  I think the issue may be with the way that version string is being reported (somehow comes in as a float).  We should probably chase this down, but perhaps via a separate issue.

I'll try to make sure things look OK on readthedocs when I get a chance (probably tonight or this weekend).

murrayrm:
Send @roryyorke a PR that creates the documentation on readthedocs.

PR #158, opened 2017-10-22
SISO tf() may not work with numpy arrays with numpy.int elements.

kangwonlee:
SISO tf() function may not work if an input argument is numpy.int.
(@sunhapos could identify this)

Please check the test for cases.

Should this be a numpy issue?  If so, is a quick patch such as 3fe4067 desirable?

Thanks,

KW

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ before_install:
   - conda config --set always_yes yes --set changeps1 no
   - conda update -q conda
   # conda-build must be installed in the conda root environment
-  - conda install conda-build
+  - conda install conda-build=2.1.17
   - conda config --add channels python-control
   - conda info -a
   - conda create -q -n test-environment python="$TRAVIS_PYTHON_VERSION" pip coverage
--- a/control/tests/tf_input_element_int.py
+++ b/control/tests/tf_input_element_int.py
@@ -1,0 +1,46 @@
+import unittest
+
+import numpy as np
+
+import control as ctl
+
+
+class TestTfInputIntElement(unittest.TestCase):
+    # currently these do not pass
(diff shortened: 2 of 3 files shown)


kangwonlee:
Does this discussion seem related? https://github.com/numpy/numpy/issues/2951

roryyorke:
Have you looked at PR #135 ? I see @jed-frey handles all sorts of types in the proposed `_cleanup` function [here](https://github.com/python-control/python-control/pull/135/files#diff-c8d41d88d4a282029b1aff9a65dae888R1333)

kangwonlee:
@roryyorke Thanks for your comments.  @jed-frey Hope this helps. Also found this [QnA](https://stackoverflow.com/questions/38155039/what-is-the-difference-between-native-int-type-and-the-numpy-int-types).

murrayrm:
A few additional comments on this:

- The same changes made here for `xferfcn.py` also need to be made in `statesp.py` and probably any other code that checks for data types `int` or `float` (need to add `numpy` types).

- A quicker way to check against all of the `numpy` data types is to check against `numpy.number`.

- Some related (and conflicting) changes are present in PR #170, though this PR goes beyond that in cleaning things up.

I'll try to do an integration of all of this in the coming days.

murrayrm:
Different parts of this PR are covered by other PR's:

* PR #170 has a fix for the originally identified issue of `numpy.int` data types.  Note that the fix there is more general (handles any `numpy.number` type).

* PR #158 fixes the issue with conda that was also addressed here (by using an earlier version of conda).

This is fixed PR #170 and also PR #158.  Closing this PR since the issue will be addressed when the other two are merged.

murrayrm:
The changes in this PR have been incorporated into PR #135.

PR #163, opened 2017-11-07
statespace constructor fix

BMaxV:
...other changes related, the deleted setup.py wasn't used anywhere as far as I could see.

### DIFF
--- a/control/statesp.py
+++ b/control/statesp.py
@@ -91,38 +91,21 @@ class StateSpace(LTI):
 
     """
 
-    def __init__(self, *args):
+    def __init__(self,A,B,C,D,dt=None):
         """Construct a state space object.
 
         The default constructor is StateSpace(A, B, C, D), where A, B, C, D are
         matrices or equivalent objects.  To call the copy constructor, call
         StateSpace(sys), where sys is a StateSpace object.
+        
+        Matrix sizes need to be consistent:
+        A must be square.
+        A and B must have the same number of rows.
+        A and C must have the same number of columns.
+        B and D must have the same number of columns.
+        C and D must have the same number of rows.
 
         """
-
-        if len(args) == 4:
-            # The user provided A, B, C, and D matrices.
-            (A, B, C, D) = args
-            dt = None;
-        elif len(args) == 5:
-            # Discrete time system
-            (A, B, C, D, dt) = args
-        elif len(args) == 1:
-            # Use the copy constructor.
-            if not isinstance(args[0], StateSpace):
-                raise TypeError("The one-argument constructor can only take in \
-a StateSpace object.  Recived %s." % type(args[0]))
-            A = args[0].A
-            B = args[0].B
-            C = args[0].C
-            D = args[0].D
-            try:
-                dt = args[0].dt
-            except NameError:
-                dt = None;
-        else:
-            raise ValueError("Needs 1 or 4 arguments; received %i." % len(args))
-
         A, B, C, D = map(matrix, [A, B, C, D])
 
         # TODO: use super here?
@@ -131,6 +114,7 @@ a StateSpace object.  Recived %s." % type(args[0]))
         self.B = B
         self.C = C
         self.D = D
+        self.dt=dt
 
         self.states = A.shape[1]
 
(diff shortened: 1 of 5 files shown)


control/statesp.py
```diff
@@ -91,38 +91,21 @@ class StateSpace(LTI):
 
     """
 
-    def __init__(self, *args):
+    def __init__(self,A,B,C,D,dt=None):
```
moorepants: This change will break previous code where users made use of the key word arg, e.g. `StateSpace(1, 2, 3, 4, dt=0.1)` will fail with a TypeError.

BMaxV: I don't really know what you mean, you can still pass arguments as *args after the change, like this:

`args=(A,B,C,D)`
`StateSpace(*args)`
`args=(A,B,C,D,dt)`
`StateSpace(*args)`

The only thing the change will break is that you can no longer pass a StateSpace instance to `StateSpace.__init__` to create a new one, you'll have to be explicit. I think the work required to unpack the values is a reasonable tradeoff to the simplicity gained.

moorepants:
```
In [6]: def f(*args):
   ...:     return args
   ...: 

In [7]: f(1, 2, 3, 4, dt=0.5)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-7-ced6aff2bf09> in <module>()
----> 1 f(1, 2, 3, 4, dt=0.5)

TypeError: f() got an unexpected keyword argument 'dt'
```

BMaxV: Your code is an example of the old behavior, that I removed.

This is the new behavior:

```
>>> def f(a,b,c,d,dt=None):
...     return a,b,c,d,dt
...     
>>> f(1,2,3,4,dt=None)
(1, 2, 3, 4, None)
```

moorepants: You are right. Dyslexia I suppose :)

control/statesp.py
```diff
@@ -91,38 +91,21 @@ class StateSpace(LTI):
...
-        if len(args) == 4:
-            # The user provided A, B, C, and D matrices.
-            (A, B, C, D) = args
-            dt = None;
-        elif len(args) == 5:
-            # Discrete time system
-            (A, B, C, D, dt) = args
-        elif len(args) == 1:
-            # Use the copy constructor.
-            if not isinstance(args[0], StateSpace):
```
moorepants: I suggest using `if isinstance(A, StateSpace)` to preserve this behavior for backwards compatibility.

BMaxV: The point of the changes in the pull request is to remove the ambiguity of being able to define statespaces in two ways:

```
S1=StateSpace(A,B,C,D)
S2=StateSpace(oldstatespace)
```

Yes that will break code for people who use the second method. 

That probably should have been in the initial comment.

If this is breaking things too much for now, maybe it would be an option to create a "next major release" branch to include this change. That might make the transition obvious and it would make it easier to use the old version in the future.

If you don't want the module to be backwards incompatible at all, reject the pull request or tell me and I'll delete it.

Also your suggested change wouldn't work because the ```StateSpace.__init__ ```
```
__init__(A,B,C,D,dt=None):
```
would expect ```B,C,D``` and throw an error if you only gave it a StateSpace instance as first argument.

The only way to preserve the functionality of being able to define a StateSpace with both, a previous instance or a list of variables, is to keep the current code.

murrayrm:
The intent of allowing `StateSpace(sys)` was as a copy constructor, as described in the function documentation.  This change would remove that functionality (also present in several other classes).

BMaxV:
I can add a `.copy()` function if that would help? If you want I can change the other classes too, to reflect this change.

murrayrm:
It seems to me that we are fixing a non-problem.  I suggest we drop this pull request and leave the current functionality.

BMaxV:
You don't see a problem with `help(control.statesp.StateSpace) `showing the user (self,*args) as valid arguments for `__init__` and the docs saying the the way to create a statespace is to call `matlab.ss(*args)` and not to use the StateSpace contructor?

moorepants:
I am not a core dev with python-control but use it often. My rule of thumb is to not break backwards compatibility unless absolutely necessary. The amount of code you may break and potential amount of frustrated users is almost always not worth the change.

How about improving the documentation and examples of use for the class instead of changing the functionality? Then it is is a win win for everyone.

murrayrm:
Agree with @moorepants that a good approach is to update the documentation.  I'll try to take a cut at that in the coming days and will close out the PR if issue is fully addressed.

murrayrm:
I have created a new PR (#171) that fixes the documentation for the class constructors.

Issue #159, opened 2017-10-22
Updates for conda_build >= 3.0


kangwonlee:
Travis-CI build fails with an error as follows:

`ValueError: support for __conda_version__ has been removed as of Conda-build 3.0.Try Jinja templates instead: http://conda.pydata.org/docs/building/meta-yaml.html#templating-with-jinja`

Proposed [link is here](http://conda.pydata.org/docs/building/meta-yaml.html#templating-with-jinja).

This is from the following line:
https://github.com/python-control/python-control/blob/05d6bc5257d74a05ee2615d18ab7f5f58a2ec372/.travis.yml#L43

kangwonlee:
IMHO nowadays conda version is > 4.0. Why does this happen?

kangwonlee:
There were [prior warnings](https://travis-ci.org/python-control/python-control/jobs/291176646#L758) : 

`Deprecation warning: support for __conda_version__ will be removed in Conda build 3.0.Try Jinja templates instead: http://conda.pydata.org/docs/building/meta-yaml.html#templating-with-jinja`

`Setting version from __conda_version__.txt: 0.7.0`

`Deprecation warning: support for __conda_buildnum__ will be removed in Conda build 3.0.Try Jinja templates instead: http://conda.pydata.org/docs/building/meta-yaml.html#templating-with-jinja`

kangwonlee:
Seems related to [this](https://github.com/conda/conda-build/issues/1023)

kangwonlee:
Others had similar [experience](https://github.com/rdkit/conda-rdkit/issues/53)s, too, I guess.

kangwonlee:
In some other case, some [modification](https://github.com/conda/conda-build/issues/2371) seems expected.

kangwonlee:
Are the __conda_version__.txt and __conda_buildnum__.txt files from the channel? Anything that can be done to the source repository?

anton-malakhov:
It's a pity that support `__conda_version__.txt` is removed. parsing setup.py is not an option for non-python packages

murrayrm:
This has been fixed in PR #167.

Issue #164, opened 2017-11-15
Compatibility issue with SciPy 1.0


jcwinkler:
The following snippet

```python
import control as ctrl
G = ctrl.tf(1, [1, 1])
t, y = ctrl.step_response(G)
 ```

executed under Python 3.5.4 with **SciPy 1.0**, NumPy 1.13.3 and python-control 0.7 and **SlyCot NOT installed** fails to run.

After investigation it turns out that the problem arrises in the file `statesp.py`, function `_convertToStateSpace` line 641: python-control tries to solve the problem using SlyCot. Since SlyCot is not found, it falls back to the `ImportError` handling in which the SciPy's `lti` is called:
```python
            lti_sys = lti(squeeze(sys.num), squeeze(sys.den))
            return StateSpace(lti_sys.A, lti_sys.B, lti_sys.C, lti_sys.D,
                              sys.dt)
```
In SciyPy versions PRIOR to 1.0 the lti_sys object had the attributes A, B, C, D. However, in SciPy 1.0 these attributes have been removed. Hence, the code fails to execute.

Possible workaround would be (runs using SciPy 0.19 and 1.0):
```python
            from scipy.signal import tf2ss
            A, B, C, D = tf2ss(squeeze(sys.num), squeeze(sys.den))
            return StateSpace(A, B, C, D, sys.dt)
```

What do you think? Since many Windows user do not have Slycot available by default it would be nice to get this fixed.

Jan

roryyorke:
This Scipy change is documented [in the scipy 1.0 release notes](https://docs.scipy.org/doc/scipy/reference/release.1.0.0.html#backwards-incompatible-changes) (look for "scipy.signal.lti" in the linked section).

As I recall, we've been getting deprecation warnings about this for a while, so I suppose it's hardly a surprise.

It looks like 05d6bc5 (HEAD as of this writing) still has this bug.

I can't easily test the fix: I can't get conda to find Scipy 1.0.

murrayrm:
This fix has been implemented in PR #170.  Changes were required in both `statesp.py` and `xferfcn.py`.  In addition, there were some changes in the way that scalar types showed up that required some additional changes.  See PR #170 comments for details.

PR #167, opened 2017-12-23
Update conda-recipe for use with conda-build >= 3.0

murrayrm:
This PR updates the conda recipe so that it works correctly with `conda-build` 3.0.  The basic change is to extract the version information directly from git, using jinja commands.

This PR should run correctly using Travis CI, fixing issue #159.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -1,7 +1,12 @@
 package:
   name: control
+  version: {{ GIT_DESCRIBE_TAG }}
+
+source:
+  git_url: ../
 
 build:
+  number: {{ GIT_DESCRIBE_NUMBER }}
   script:
     - cd $RECIPE_DIR/..
     - $PYTHON make_version.py
--- a/make_version.py
+++ b/make_version.py
@@ -1,3 +1,24 @@
+# make_version.py - generate version information
+#
+# Author: Clancy Rowley
+# Date: 2 Apr 2015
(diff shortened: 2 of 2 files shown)


murrayrm:
Integration checks look OK.  Most are failing because of issue #164 (scipy 1.0 compatibility), but otherwise running properly.  There is also an unknown issue with python3.3 that seems to be related to an error while installing package `defaults::openssl-1.0.1k-1`.

murrayrm:
Followup comment: need to rethink this PR a bit.  By removing make_version.py, I messed up the automatic versioning that @cwrowley put into place in PR #52.  Probably still need to run `make_version.py` in the conda recipe, but not rely on it for the version information in conda (since `__conda_version__` is not supported).

murrayrm:
I've updated this PR to leave `make_version.py` in place, but still use `jinja` for generating the version information in `meta.yaml`.  This will allow conda-build 3.0 to work correctly.

Travis CI checks fail due to SciPy c1.0 incompatibility.  This is expected and solved by PR #170.

roryyorke:
FWIW, Python 3.3 is end-of-life (see https://devguide.python.org/#status-of-python-branches ).  Are you aware of any python-control users requiring it?  If not, we might as well drop it.

murrayrm:
Travis CI testing for Python 3.3 (and 3.4) has been removed in PR #169.  I don't see any need to support Python 3.3 going forward => haven't bother to debug the issue that was described above.

Issue #168, opened 2017-12-23
Slycot installation failing in Travis CI


murrayrm:
In the latest builds on `python-control`, the `slycot` installation is failing and so unit tests are not being run across the `slycot` version of the library.  It appears that what is happening is that the conda builds for `slycot` are out of date and so don't install properly under the Travis CI build script.

murrayrm:
Error message from PR #169 when executing `python -c "import slycot"`:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/slycot/__init__.py", line 16, in <module>
    from .analysis import ab01nd,ab05md,ab05nd,ab07nd,ab08nd, \
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/slycot/analysis.py", line 21, in <module>
    from . import _wrapper
ImportError: libgfortran.so.3: cannot open shared object file: No such file or directory
```

PR #169, opened 2017-12-23
Travis CI update for new/old scipy + improved slycot

murrayrm:
This PR addresses two related issues:

- The current version of python-control is incompatible with `scipy-1.0.0` and so all checks are failing.  In this build script we do explicit checks against `scipy-0.19.0` and `scipy-1.0.0`. This is related to issue #164.

- The way that tests were being done against slycot, they were actually not checking to make sure slycot was installed correctly => they 'succeeded' if slycot failed to install.  The new version of the build script separates out the slycot versus non-slycot builds so that this is more obvious.  Issue #168 describes the issue.

In addition to these changes, I updated the python version numbers to check against to 2.7, 3.5, and 3.6 (versions 3.3 and 3.4 seem old enough that we can stop checking against them).

If this PR works correctly, it should have a revised set of Travis CI checks that succeed on the 0.19.1 version of scipy with slicot not installed, but fail on all other cases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,19 @@ cache:
 
 python:
   - "2.7"
-  - "3.3"
-  - "3.4"
   - "3.5"
+  - "3.6"
+
+# Test against multiple version of SciPy, with and without slycot
+#
+# Because there were significant changes in SciPy between v0 and v1, we 
+# test against both of these using the Travis CI environment capability
+#
+# We also want to test with and without slycot
+env:
+  - SCIPY=0.19.1 SLYCOT=
+  - SCIPY=0.19.1 SLYCOT=slycot
+  - SCIPY=1.0.0 SLYCOT=
 
 # install required system libraries
 before_install:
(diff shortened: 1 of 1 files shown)


murrayrm:
Check completed as described above.

murrayrm:
Update PR to include new rules for building slycot from source, which allows proper checking against the slycot version of the library.  A couple of notes on the implementation:

- Need to be careful about how to install `gfortran` and `lapack`.  Install both via `apt-get` in the root environment (not the test-environment) to make sure that compilers use the right paths (found this out the hard way, but trying almost every other combination of things).

- Python3 implementations need to have the `LIBRARY_PATH` environment variable set up properly, otherwise slycot will fail to link properly.  For this, you need to link to the version in the local environment, so that you get the right version of python.

While I was at it, I made a few other changes:

- Reordered the build cases to start with the most up-to-date case (`python-3.6`, `scipy-1.0`, `slycot`) gets tested version.

- Added more comments throughout to explain what is going on.

- Don't bother to install `conda-build` since it is not needed any more.

Some of the Travis builds will still fail for this version since the corrections for `scipy-1.0` from PR #164 are not yet included.

PR #170, opened 2017-12-23
Fixes for SciPy 1.0 compatibility

murrayrm:
This PR fixes the SciPy 1.0 compatibility issues raised in issue #164.  There are two main changes:

- Updated `statesp.py` and `xferfcn.py` to use the new forms of the `scipy.signal` library, for which the `lti` class no longer has accessors for `num`, `den`, etc.  The fix is to directly call the `ss2tf` and `tf2ss` functions.

- Also had to add a check for scalars that are of a `numpy` data type, specifically for `int64`.  Prior to Python 3, checking against `int` worked correctly, but this is no longer the case.  As a fix, I added checks throughout the library against `numpy.number`, which captures all scalar data types in `numpy`.

I also checked (manually) to insure that this version works against `scipy-0.19.0`, which should mean it is OK for backward compatibility (this can be checked with PR #169 is merged).

### DIFF
--- a/control/bdalg.py
+++ b/control/bdalg.py
@@ -54,6 +54,7 @@ $Id$
 """
 
 import scipy as sp
+import numpy as np
 from . import xferfcn as tf
 from . import statesp as ss
 from . import frdata as frd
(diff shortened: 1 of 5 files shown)


control/bdalg.py
```diff
@@ -221,18 +222,18 @@ def feedback(sys1, sys2=1, sign=-1):
     """
 
     # Check for correct input types.
-    if not isinstance(sys1, (int, float, complex, tf.TransferFunction,
-                             ss.StateSpace, frd.FRD)):
+    if not isinstance(sys1, (int, float, complex, np.number,
```
roryyorke: I'm struggling to find docs on numpy.number; it has a __doc__ string, but no entry, e.g., here: https://docs.scipy.org/doc/numpy-1.13.0/genindex.html .  This use seems reasonable, just not sure how standard it is.

murrayrm: Numpy scalar type hierarchy is here: https://docs.scipy.org/doc/numpy-1.13.0/reference/arrays.scalars.html

PR #171, opened 2017-12-29
Update docstrings for LTI class constructors

murrayrm:
This PR addresses the documentation issues raised in PR #163 related to how the `StateSpace`, `TransferFunction`, and `FRD` constructors are documented.  In this PR, the first line of the docstring is modified for the each of the above classes, as well as their `init` methods and the corresponding `ss`, `tf`, and `frd` functions to have the most common call signature.

This PR leaves in place the copy constructor functionality and only modifies the documentation.

While I was at it, I also fixed some indentation problems and other issues that were causing warnings in sphinx.

### DIFF
--- a/control/ctrlutil.py
+++ b/control/ctrlutil.py
@@ -43,17 +43,12 @@
 # Packages that we need access to
 from . import lti
 import numpy as np
-from numpy import pi
-
-# Hack for sphinx.ext.autodoc: if numpy is a mock import, then numpy.pi 
-# will be assigned to _Mock() and this generates a type error
-if not isinstance(pi, float):
-    pi = 3.14
+import math
 
 __all__ = ['unwrap', 'issys', 'db2mag', 'mag2db']
 
 # Utility function to unwrap an angle measurement
-def unwrap(angle, period=2*pi):
+def unwrap(angle, period=2*math.pi):
     """Unwrap a phase angle to give a continuous curve
 
     Parameters
--- a/control/ctrlutil.py
+++ b/control/ctrlutil.py
@@ -45,6 +45,11 @@ from . import lti
 import numpy as np
 from numpy import pi
 
+# Hack for sphinx.ext.autodoc: if numpy is a mock import, then numpy.pi 
+# will be assigned to _Mock() and this generates a type error
+if not isinstance(pi, float):
+    pi = 3.14
+
 __all__ = ['unwrap', 'issys', 'db2mag', 'mag2db']
 
 # Utility function to unwrap an angle measurement
--- a/control/margins.py
+++ b/control/margins.py
@@ -50,11 +50,12 @@ Date: 14 July 2011
 $Id$
 """
 
+import math
 import numpy as np
+import scipy as sp
 from . import xferfcn
 from .lti import issiso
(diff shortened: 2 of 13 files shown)


murrayrm:
Added more information to main documentation on class constructors + updated docstrings for other functions with variable arguments.  The justification for including them in this PR is that they are consistent with the initial changes and the comments in PR #163.

doc/conventions.rst
```diff
@@ -9,6 +9,97 @@ Library conventions
...
+function.  A full list of functions can be found in :ref:`function-ref`.
+
+Transfer functions
+------------------
+The :class:`TransferFunction` class is used to represent input/output
+transfer functions
+
+.. math::
+
+  G(s) = \frac{\text{num}(s)}{\text{den(s)}}
```
roryyorke: in first case (s) is outside `\text{}`, in second case inside it.

doc/conventions.rst
```diff
@@ -9,6 +9,97 @@ Library conventions
 The python-control library uses a set of standard conventions for the way
 that different types of standard information used by the library.
 
+LTI system representation
+=========================
+
+Linear time invariant (LTI) systems are represented in python-control in
+state space, transfer function, or frequency response data (FRD) form.  Most
+functions in the toolbox will operate on any of these data types and
+functions for convering between between compatible types is provided.
```
roryyorke: converting

control/margins.py
```diff
@@ -338,7 +336,9 @@ def phase_crossover_frequencies(sys):
 
 
 def margin(*args):
-    """Calculate gain and phase margins and associated crossover frequencies
+    """margin(sys)
```
roryyorke: Just below (line 345) the name "sysdata" is used; perhaps that should be used here?

control/ctrlutil.py
```diff
@@ -45,6 +45,11 @@
 import numpy as np
 from numpy import pi
 
+# Hack for sphinx.ext.autodoc: if numpy is a mock import, then numpy.pi 
+# will be assigned to _Mock() and this generates a type error
+if not isinstance(pi, float):
+    pi = 3.14
```
roryyorke: how about "from math import pi" here?  Or does that result in a mocked instance of some sort too?

doc/conventions.rst
```diff
@@ -9,6 +9,97 @@ Library conventions
...
+The :class:`TransferFunction` class is used to represent input/output
+transfer functions
+
+.. math::
+
+  G(s) = \frac{\text{num}(s)}{\text{den(s)}}
+       = \frac{a_0 s^n + a_1 s^{n-1} + \cdots + a_n}
+              {b_0 s^m + b_1 s^{m-1} + \cdots + b_m},
+
+where n is generally greater than m (for a proper transfer function).
```
roryyorke: maybe greater than or equal to?

control/ctrlutil.py
murrayrm: Much better fix since math is included in the base python distribution.

murrayrm:
Good feedback @roryyorke.  Small updates pushed to address these.

Issue #116, opened 2016-10-09
Use SLICOT's TB05AD in Statespace.freqresp


roryyorke:
StateSpace.freqresp current converts to a transfer-function matrix (TFM) to evaluate the frequency response (see [1]) .  

Polynomials can give numerical problems (see below), so this method should probably use `[2]` (unfortunately not currently exposed in Slycot).  This technique uses an orthogonal similarity transform to make the system quicker to evaluate.

One possible problem is that if one evaluates a decoupled (diagonal TFM) system in which one system has poles on the imaginary axis, I imagine the whole frequency response will be infinite.  I'm not sure if this is really important.

[1] https://github.com/python-control/python-control/blob/cdd3e73838cea0b26a7155d0b2084aa698e5eefd/control/statesp.py#L398
[2] http://slicot.org/objects/software/shared/doc/TB05AD.html

A script to demonstrate the problem:

``` python
import numpy as np
import control

wn = 1;
zeta = 0.01;
g = control.tf(wn**2, [1,2*zeta*wn,wn**2])

def trivial_evalfr(gss,w):
    I = np.eye(*gss.A.shape)
    return np.dot(gss.C,np.linalg.solve(1j*w*I-gss.A,gss.B)) + gss.D

def trivpow(g,n):
    from functools import reduce
    from operator import mul
    assert g.inputs == g.outputs
    I = np.eye(g.inputs)
    return reduce(mul,[g for i in range(n)])

gss = control.ss(g)

print()
print('{:3s} {:>9s} {:>9s} {:>9s} {:>9s}'.format('exp','tfv/ref','tfv/ref','ssv/ref','ssv/ref'))
print('{:3s} {:>9s} {:>9s} {:>9s} {:>9s}'.format('','1','dB','1','dB'))

exps = range(1,21)

gtfs = [trivpow(g,i)
        for i in exps]
gsss = [trivpow(gss,i)
        for i in exps]

for exp,gtfi,gssi in zip(exps,gtfs,gsss):
    ref = g.evalfr(1)**exp
    tfv = gtfi.evalfr(1)
    ssv = trivial_evalfr(gssi,1)

    tferr = np.abs(tfv/ref)[0,0]
    sserr = np.abs(ssv/ref)[0,0]

    db = lambda x: 20*np.log10(x)
    print('{:3d} {:9.2e} {:9.2e} {:9.2e} {:9.2e}'.format(exp,tferr,db(tferr),sserr,db(sserr)))
```

result is

```
[...]

PR #172, opened 2018-01-02
Use slycot's tb05ad for faster and more accurate frequency response e…

rabraker:
This pull request addresses issue [#116](https://github.com/python-control/python-control/issues/116).

It uses the (relatively) newly wrapped [TB05AD](https://github.com/python-control/Slycot/pull/10) function from Slycot, when Slycot is available. If slycot is not available, I instead use the state space built-in method `horner`, rather than converting to a transfer function for the evaluation. 

This passes the unit tests on my machine (python 2.7), but currently it looks like the travis build is [failing](https://travis-ci.org/python-control/python-control/builds/323626585) so I'm not sure what happens with other versions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ __conda_*.txt
 record.txt
 build.log
 *.egg-info/
+.eggs/
 .coverage
 doc/_build
 doc/generated
@@ -18,3 +19,7 @@ examples/.ipynb_checkpoints/
 .project
 Untitled*.ipynb
 *.idea/
+
+# Files created by or for emacs (RMM, 29 Dec 2017)
+*~
+TAGS
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,13 +9,31 @@ cache:
     - $HOME/.local
 
 python:
-  - "2.7"
-  - "3.3"
-  - "3.4"
+  - "3.6"
   - "3.5"
+  - "2.7"
+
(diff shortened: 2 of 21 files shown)


murrayrm:
@rabraker Can you rebase these changes against the current master.  That should fix up the problems in Travis CI and allow testing against python 3.5 and 3.6.

rabraker:
I'm evidently all thumbs with this rebasing business. Whatever I did makes all the checks pass, but somehow there are commits from others going back almost a year in this PR now. I don't know how to fix that.

So it seems easiest to me to delete this PR and I'll just make a new, clean one.

murrayrm:
Pretty sure there is a way to rewrite history that will make this a clean commit.  @slivingston is a master at git, perhaps he has some suggestions?  

One possibility is that you didn't grab the upstream changes from python-control:master and that is causing the large number of extraneous commits.  As I check, I did a manual merge with my copy of the master branch and the only change was to statesp.py.  I think if you do `git pull upstream master` on your branch, it will pull in all of the latest changes and leave you with a clean commit that you can then push back.

rabraker:
With my branch `freqresp_tb05ad` checked out, I tried what you suggest:

```
git pull upstream master
From https://github.com/python-control/python-control
 * branch            master     -> FETCH_HEAD
Already up-to-date.
```
ie, this had no effect. I'll just close this PR and submit another.

PR #173, opened 2018-01-02
Use slycot's tb05ad for faster and more accurate frequency response 

rabraker:
This pull request addresses issue #116.

It uses the (relatively) newly wrapped TB05AD function from Slycot, when Slycot is available. If slycot is not available, I instead use the state space built-in method horner, rather than converting to a transfer function for the evaluation.

### DIFF
--- a/control/statesp.py
+++ b/control/statesp.py
@@ -386,22 +386,98 @@ but B has %i row(s)\n(output(s))." % (self.inputs, other.outputs))
         return array(resp)
 
     # Method for generating the frequency response of the system
-    def freqresp(self, omega):
-        """Evaluate the system's transfer func. at a list of ang. frequencies.
+    def freqresp(self, omega_s):
+        """Evaluate the system's transfer func. at a list of freqs, omega.
 
         mag, phase, omega = self.freqresp(omega)
 
-        reports the value of the magnitude, phase, and angular frequency of the
-        system's transfer function matrix evaluated at s = i * omega, where
-        omega is a list of angular frequencies, and is a sorted version of the
-        input omega.
+        Reports the frequency response of the system,
+
+             G(j*omega) = mag*exp(j*phase)
+
+        for continuous time. For discrete time systems, the response is
+        evaluated around the unit circle such that
+
+             G(exp(j*omega*dt)) = mag*exp(j*phase).
+
+        Inputs:
+        ------
+           omega_s: A list of frequencies in radians/sec at which the system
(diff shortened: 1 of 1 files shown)


control/statesp.py
```diff
@@ -386,22 +386,98 @@ def horner(self, s):
         return array(resp)
 
     # Method for generating the frequency response of the system
-    def freqresp(self, omega):
-        """Evaluate the system's transfer func. at a list of ang. frequencies.
+    def freqresp(self, omega_s):
```
murrayrm: Is there a reason to use `omega_s` here instead of `omega`?  In all other docstrings for `freqresp` (eg, `xferfcn.py`, `lti.py`) we just use `omega`.  Also, shows up as `omega` below.  I would leave as `omega` for consistency.

control/statesp.py
```diff
@@ -386,22 +386,98 @@ def horner(self, s):
...
+             G(j*omega) = mag*exp(j*phase)
+
+        for continuous time. For discrete time systems, the response is
+        evaluated around the unit circle such that
+
+             G(exp(j*omega*dt)) = mag*exp(j*phase).
+
+        Inputs:
+        ------
+           omega_s: A list of frequencies in radians/sec at which the system
```
murrayrm: Change to `omega`?

control/statesp.py
```diff
@@ -386,22 +386,98 @@ def horner(self, s):
...
-        # n-dimensional linear system at each frequency
-        tf = _convertToTransferFunction(self)
-        return tf.freqresp(omega)
+
+        # In case omega is passed in as a list, rather than a proper array.
+        omega_s = np.asarray(omega_s)
+
+        numFreqs = len(omega_s)
+        Gfrf = np.empty((self.outputs, self.inputs, numFreqs),
+                        dtype=np.complex128)
```
murrayrm: Why `complex128`?  I don't have a specific objection, but seems a bit arbitrary.  OK to leave, unless there is some justification for doing something else (perhaps just `complex', which presumably defaults to machine precision?).

rabraker: I think I did this because the numpy [page about dtypes](https://docs.scipy.org/doc/numpy-1.10.1/user/basics.types.html) doesn't list `complex` as a type.

Looking further down the page though, they say that  `dtype=complex` will be interpreted as `dtype=np.complex_` which is shorthand for `np.complex128`. To me, using `np.complex128` is the most explicit as to what's happening.

Issue #174, opened 2018-01-02
lti.damp is implemented only for continuous time systems


rabraker:
The `damp` function in [lti.py](https://github.com/python-control/python-control/blob/33bebc190ecb3f63a7da5750710dd235bfd24cce/control/lti.py#L87) is only implemented for continuous time systems, which is not mentioned in the doc string and does not return an error for discrete systems. 

I will submit a PR to add support for discrete time systems.

PR #175, opened 2018-01-02
Add support for discrete systems to lti/damp

rabraker:
This PR fixes issue [#174](https://github.com/python-control/python-control/issues/174) by adding support for discrete time systems to `lti.damp`.

I have also augmented the unit test in `lti_test.test_temp` to check the discrete time case.


murrayrm:
Reviewed.  Looks good.

Issue #117, opened 2016-10-11
Implement a better algorithm for pole placement


vmatos:
Hello!
I've happened to compare pole placement with Matlab's, and found out that returned matrix doesn't make the poles fall in the chosen values.
I made an example script just like `python-control/examples/slicot-test.py` for matlab and python.

Matlab script:

```
A = [
[                0, 0, 0, 1],
[-42.7207306947135, 0, 0, 0],
[                0, 0, 0, 0],
[ 47.0334901743703, 0, 0, 0]];

C = [[ 1.,  0.,  0.,  0.],
       [ 0.,  1.,  0.,  0.],
       [ 0.,  0.,  1.,  0.],
       [ 0.,  0.,  0.,  1.]];

p = [-31., -21., -20., -30.];

L = place(A',C',p)
eigs(A' - C'*L)

% L= 31.0000  -42.7207         0   47.0335
%          0   21.0000         0         0
%          0         0   20.0000         0
%     1.0000         0         0   30.0000
% eigs = -20, -21, -30, -31
```

Python script:

```
import numpy as np              # Numerical library
from scipy import *             # Load the scipy functions
from control.matlab import *    # Load the controls systems library

A = matrix([
[                0, 0, 0, 1],
[-42.7207306947135, 0, 0, 0],
[                0, 0, 0, 0],
[ 47.0334901743703, 0, 0, 0]])

C = matrix([[ 1.,  0.,  0.,  0.],
       [ 0.,  1.,  0.,  0.],
       [ 0.,  0.,  1.,  0.],
       [ 0.,  0.,  0.,  1.]])

p = [-31., -21., -20., -30.]
L = place(A.T, C.T, p)
print "Pole place: L = ", L
print "Pole place: eigs = ", np.linalg.eig(A.T - C.T * L)[0]

#Pole place: L =  [[  0.73152698  -1.12180554  -0.           5.01688279]
# [ -1.12180554  22.72030247  -0.          -7.69345088]
# [ -0.          -0.          20.          -0.        ]
# [  5.01688279  -7.69345088  -0.          34.40626723]]
#Pole place: eigs =  [ -6.85809669 -30.         -21.         -20.        ]
# >>> control.__version__
# '0.7.0'
```

Am I using it wrong?

robertobucher:
The problemis related to the MIMO system: it doesn't have a unique 
solution! SW like Matlab or Scilab or Scicoslab use algorithms that are 
able to find well conditioned solutions, python not!

[...]

PR #176, opened 2018-01-03
Update place to use scipy.signal.place_poles

rabraker:
This PR addresses issue [#117](https://github.com/python-control/python-control/issues/117) by using the `place_poles` function in `scipy.signals`.

Aside from just being run in `matlab_test.py`, there wasn't, as far as I could tell,  a unit test for `place` previously, so I added one. 

The one limitation of the algorithm used in scipy.signals is that it can't place multiple poles at the same location for SISO systems, which made the "test" in `matlab_tests.place` fail, so I modified the requested pole location there. 

Also worth pointing that the previous implementation would fail for discrete time systems because the slycot function needed to be called with a different flag and the `alpha` parameter needed to be computed differently.

### DIFF
--- a/control/tests/matlab_test.py
+++ b/control/tests/matlab_test.py
@@ -400,9 +400,8 @@ class TestMatlab(unittest.TestCase):
         modred(self.siso_ss3, [1], 'matchdc')
         modred(self.siso_ss3, [1], 'truncate')
 
-    @unittest.skipIf(not slycot_check(), "slycot not installed")
     def testPlace(self):
-        place(self.siso_ss1.A, self.siso_ss1.B, [-2, -2])
+        place(self.siso_ss1.A, self.siso_ss1.B, [-2, -2.5])
 
     @unittest.skipIf(not slycot_check(), "slycot not installed")
     def testLQR(self):
--- a/control/tests/matlab_test.py
+++ b/control/tests/matlab_test.py
@@ -400,8 +400,16 @@ class TestMatlab(unittest.TestCase):
         modred(self.siso_ss3, [1], 'matchdc')
         modred(self.siso_ss3, [1], 'truncate')
 
+    @unittest.skipIf(not slycot_check(), "slycot not installed")
+    def testPlace_varga(self):
+        place_varga(self.siso_ss1.A, self.siso_ss1.B, [-2, -2])
+
     def testPlace(self):
         place(self.siso_ss1.A, self.siso_ss1.B, [-2, -2.5])
+
+    def testAcker(self):
+        acker(self.siso_ss1.A, self.siso_ss1.B, [-2, -2.5])
+
 
     @unittest.skipIf(not slycot_check(), "slycot not installed")
     def testLQR(self):
--- a/control/tests/statefbk_test.py
+++ b/control/tests/statefbk_test.py
@@ -180,6 +180,27 @@ class TestStatefbk(unittest.TestCase):
         np.testing.assert_raises(ControlDimension, place, A[1:, :], B, P)
         np.testing.assert_raises(ControlDimension, place, A, B[1:, :], P)
 
+        # Check that we get an error if we ask for too many poles in the same
+        # location. Here, rank(B) = 2, so lets place three at the same spot.
+        P_repeated = np.array([-0.5, -0.5, -0.5, -8.6659])
+        np.testing.assert_raises(ValueError, place, A, B, P_repeated)
+
+    def testPlace_varga(self):
+        A = np.array([[1., -2.], [3., -4.]])
+        B = np.array([[5.], [7.]])
+
+        P = np.array([-2., -2.])
+        K = place_varga(A, B, P)
+        P_placed = np.linalg.eigvals(A - B.dot(K))
+        # No guarantee of the ordering, so sort them
+        P.sort()
+        P_placed.sort()
+        np.testing.assert_array_almost_equal(P, P_placed)
+
+        # Test that the dimension checks work.
+        np.testing.assert_raises(ControlDimension, place, A[1:, :], B, P)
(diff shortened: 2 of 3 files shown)


murrayrm:
I like this change, but it brings up an interesting issue: what should we do when there are multiple algorithms available to solve a certain problem, each with different limitations?  The previous version of the place command use a method by Varga (IEEE TAC, 1981), implemented via `slycot`,  that did not have the limitation on repeated poles, but may give solutions that are less well-conditioned (as pointed out in issue #117).

As an alternative, what if we implement an `algorithm` keyword that allows different algorithms to be used.  This would give:

* `place(sys, poles, algorithm='scipy')`: proposed implementation in this PR
* `place(sys, poles, algorithm-`varga`: original solution

We could have `place()` be a wrapper that looks for the algorithm keyword and decides whether to call `place_scipy` or `place_varga`.

Would be interested in getting thoughts on this from anyone paying attention, especially @rabraker, @vmatos, and @robertobucher (who have been involved in the conversation and code up to now).

cwrowley:
I like the suggestion to add a keyword argument to specify the algorithm.  I also think it would be good to retain the repeated-pole test, somehow marking it as "expected to fail" if the YT algorithm is used. Also, note that `scipy.signal.place_poles` has a keyword argument `method`, which can take values `YT` or `KNV0`. So it might make sense to have at least the three options (`varga`, `YT`, `KNV0`).  Might be cleanest to handle the keyword argument `varga` ourselves, and just pass along the other keyword arguments to `scipy.signal.place_poles`, in case other algorithms are implemented in `scipy.signal` in the future.

rabraker:
I think trying to cram too much functionality into a single function makes the code more difficult to test, maintain, use and document. Part of the problem is that varga and YT require different parameters. For example, to place discrete poles using varga, you need to pass `D` into the slycot function.  This is not required with YT. Because we are passing in just matrices (A, B) and not the full LTI system, then this flag must necessarily be an argument in the wrapper function.  Now, you have a parameter that only gets used sometimes and this must be explained in the documentation.

We already have `acker` as its own function. If it were up to me, I would keep the convention that different algorithms stay in their own functions. It can be up to us to decide which is the "canonical" implementation and gets put into `place` proper. The rest can reside in `place_varga` or `place_yt` or similar and will be discoverable via 'see also:' in the doc string.  

My proposal would be the following:

- keep the current code for place, but change the function name to `place_varga()`. Also, expose the `alpha`  parameter and `C` or `D` flag to the user. There is a bug in how the `alpha` parameter is currently calculated that causes the placement to fail in some situations. This should probably be fixed in a separate pull request.  
- Keep the YT implementation in `place` proper. It is my understanding that this algorithm most closely resembles the matlab implementation and would be least likely to surprise users.

ilayn:
Note that, the arbitrary eigenstructure assignment problem has gained a lot of attention lately and people like R. Schmid et al. seem to have the general solution to pole placement. I have asked Amit Pandey from UCSD and he kindly shared his matlab code with me. I will try to propose a PR to SciPy whenever I can find time. This means, if accepted, there will be at least one additional SciPy method.

murrayrm:
Creating separate functions for now seems OK.

@rabraker Can you implement your proposal above by updating this PR to pull out the previous functionality as `place_varga()` and create an issue to capture the `alphaa` parameter bug?  

@ilayn If you (or someone) wants to convert the MATLAB code for the R. Schmid et al method into a separate `place_XYZ` method, we could include here while waiting for the official SciPy update.

rabraker:
will do

rabraker:
As requested, I have submitted the [bug report](https://github.com/python-control/python-control/issues/177) and have put the old `place` function into `place_varga`.

control/tests/matlab_test.py
```diff
@@ -400,9 +400,8 @@ def testModred(self):
         modred(self.siso_ss3, [1], 'matchdc')
         modred(self.siso_ss3, [1], 'truncate')
 
-    @unittest.skipIf(not slycot_check(), "slycot not installed")
     def testPlace(self):
-        place(self.siso_ss1.A, self.siso_ss1.B, [-2, -2])
```
murrayrm: I think we should leave this test in place, but calling `place_varga` (assuming `slycot` is installed).  Also, can we add a check for the right type of error if we call `place` with repeated poles?

control/tests/statefbk_test.py
```diff
@@ -157,6 +157,29 @@ def testAcker(self):
...
+                      [-0.5814, -4.290,   0,      0.6750],
+                      [1.067,   4.273,  -6.654,  5.893],
+                      [0.0480,  4.273,   1.343, -2.104]])
+
+        B = np.array([[0,      5.679],
+                      [1.136,  1.136],
+                      [0,      0,],
+                      [-3.146,  0]])
+        P = np.array([-0.5+1j, -0.5-1j, -5.0566, -8.6659])
+        K = place(A, B, P)
```
murrayrm: Can we also add some tests for `place_varga` and `acker` here, as a way to increase coverage?

control/tests/matlab_test.py
rabraker: I have made this change. I have also added two place-holder tests (or whatever they are called) into matlab_test.py for place and acker.

 I added an assert_raises(ValueError....) test for place() to statefbk_test.py.

control/tests/statefbk_test.py
rabraker: I think `acker` already has a test [(line 125)](https://github.com/python-control/python-control/blob/df4ee828ec443a816c52285bc8df2589706ee132/control/tests/statefbk_test.py#L125)...

I added a test for place_varga.

ilayn:
> If you (or someone) wants to convert the MATLAB code for the R. Schmid et al method into a separate place_XYZ method, we could include here while waiting for the official SciPy update.

I will get back to this asap and let you know.

Issue #183, opened 2018-01-06
Matplotlib deprecation warning


roryyorke:
see, e.g.,

https://travis-ci.org/python-control/python-control/jobs/325460249#L2815

The warning says:
```
test_siso (control.tests.freqresp_test.TestFreqresp) ... /home/travis/miniconda/envs/test-environment/lib/python3.6/site-packages/matplotlib/cbook/deprecation.py:106: MatplotlibDeprecationWarning: Adding an axes using the same arguments as a previous axes currently reuses the earlier instance.  In a future version, a new instance will always be created and returned.  Meanwhile, this warning can be suppressed, and the future behavior ensured, by passing a unique label to each axes instance.
```

murrayrm:
This has been fixed in PR #187.  The issue is a planned change in the behavior of the `matplotlib` `axes` command (which is called by `matplotlib.subplot` in `control.bode_plot` and `control.gangof4_plot`).  In the old behavior, if you called `axes` or `subplot` with the same argument, you got a pointer to the previous object.  This behavior is not desired and so it is being deprecated.  Details in [matplotlib PR #9037](https://github.com/matplotlib/matplotlib/pull/9037).

What generated this behavior in the test suit was that we called `bode` twice in a row in the unit test and this caused `matplotlib.pyplot.subfig` to get called with the same arguments.  The fix was to add a randomly generated internal "label" (not displayed anywhere), as described in the warning message.

Issue #182, opened 2018-01-06
Deprecation warning from importlib due to \dot, etc., in docstrings


roryyorke:
See, e.g., 

  https://travis-ci.org/python-control/python-control/jobs/325460249#L2717

This particular one appears to be due to `\dot` at https://github.com/python-control/python-control/blob/7b2defe28034058eecdbde9207dc5854ab3faabd/control/statesp.py#L1056 ; the `\d` is interpreted as an (invalid) escape sequence.

The fix is probably to use a raw string.

murrayrm:
This particular line (and similar ones near it) are using a Sphinx `:math:` command (on the previous line) so that they are typeset as mathematics using sphinx (and on readthedocs).  Whatever fix is used, we should try to preserve this level of processing, so that the online documentation looks nice.  Not sure how raw strings are interpreted by Sphinx.

roryyorke:
A raw string is used in one place, namely the [long, long string at the end of `control/matlab/__init__.py`](https://github.com/python-control/python-control/blob/1801c9a4850bb82b4a3d350570440f9c4893c691/control/matlab/__init__.py#L89) --- however, I don't think that string is actually used (can't find that table on the readthedocs site).

Assuming Sphinx uses the Python interpreter to load modules, which seems likely (why write a Python parser if you don't have to?), Sphinx shouldn't care about the differences between raw and ordinary strings.

Separately: I still can't get Sphinx to work.  I can hack around the error I'm getting with this patch to numpydocs.py
```
--- numpydoc.py.orig	2017-06-20 11:30:40.000000000 +0200
+++ numpydoc.py	2018-01-08 21:34:58.110632976 +0200
@@ -115,7 +115,8 @@
         return
     doc = SphinxDocString(pydoc.getdoc(obj))
     sig = doc['Signature'] or getattr(obj, '__text_signature__', None)
-    if sig:
+    from mock import Mock
+    if sig is not None and not isinstance(sig,Mock):
         sig = re.sub(sixu("^[^(]*"), sixu(""), sig)
         return sig, sixu('')
```
so it's something to do with a Mock object.  I don't know the mock library, and couldn't figure out what is being mocked.

Even after this workaround, I don't get typeset equations (e.g., for `\dot x = A x + B u` I see that text, not the equation). I do have LaTeX installed, but perhaps there's some additional Sphinx plugin required.

murrayrm:
Fixed in PR #187 by escaping backlashes.

Created a separate issue #188.

PR #187, opened 2018-01-15
Fix deprecation and formatting warnings

murrayrm:
This PR fixes the warnings identified in issues #182 and #183:

* Changed `\dot` and `\cdot` to `\\dot` and `\\cdot` to avoid warnings in docstrings that contain mathematics (the `sphinx` documentation identifies this as the preferred method for including mathemathics).

* Added a randomly generated label to `subplot` calls to avoid a warning message in matplotlib 2.1.


control/freqplot.py
```diff
@@ -171,15 +171,23 @@ def bode_plot(syslist, omega=None, dB=None, Hz=None, deg=None,
             #! TODO: Not current implemented; just use subplot for now
 
             if (Plot):
+                # Create a unique label to fix bug in matplotlib<=2.1
+                # See https://github.com/matplotlib/matplotlib/issues/9024
+                import random
+                figlabel = str(random.randint(1, 1e6))
```
roryyorke: why not plt.clf() instead?

also, this is in a loop over systems (starting line 139) -- as I understand, this creates a new axis every time (unclear to me if the old one remains in existence).  Why I try this:
```
>>> g = control.tf(1,[1,1])
>>> h = control.tf(1,[1,2])
>>> _ = control.bode_plot([g, h])
>>> plt.show()
```

I see only the frequency response of h.

murrayrm: plt.clf(): the default behavior for `matplotlib` is that if you call plot twice, you get the new data on top of the previous data.  `bode` is supposed to emulate that => this PR implements that behavior.

For the example with a list of TFs: this is the incorrect behavior, but need to look into it.  Could be that we need to pass a tuple?  (Can't check right now.)

roryyorke: If I run this:
```python
#!/usr/bin/env python3

import control

g = control.tf(1,[1,1])
h = 1 - g

control.bode_plot(g)
control.bode_plot(h)

plt.show()
```
on your branch, I see only h.

My understanding from [the docs](https://matplotlib.org/api/_as_gen/matplotlib.figure.Figure.html#matplotlib.figure.Figure.add_axes) is that different labels will result in different axes being created; from that link:
```
The axes label attribute has been exposed for this purpose: if you want two axes that are otherwise identical to be added to the figure, make sure you give them unique labels
```

I didn't realize you were aiming to emulate Matplotlib's behaviour; this is reasonable, but seems like it will require some way to check if there are existing Bode plot axes.  If Matplotlib has the equivalent of Matlab's "findobj", it should be doable: search the current figure for axes with labels "pycontrol-bode-mag", and "[...]-phase", and re-use them if they exist; if they don't, create them.

roryyorke: this seems to work:
```diff
diff --git a/control/freqplot.py b/control/freqplot.py
index c53e83f..368bdfc 100644
--- a/control/freqplot.py
+++ b/control/freqplot.py
@@ -135,6 +135,20 @@ def bode_plot(syslist, omega=None, dB=None, Hz=None, deg=None,
             else:
                 omega = sp.logspace(np.log10(omega_limits[0]), np.log10(omega_limits[1]), endpoint=True)
 
+    if Plot:
+        fig = plt.gcf()
+        ax_mag = None
+        ax_phase = None
+        for ax in fig.axes:
+            if ax.get_label() == 'pycontrol-bode-mag':
+                ax_mag = ax
+            elif ax.get_label() == 'pycontrol-bode-phs':
+                ax_phase = ax
+        if ax_mag is None or ax_phase is None:
+            plt.clf()
+            ax_mag = plt.subplot(211, label = 'pycontrol-bode-mag')
+            ax_phase = plt.subplot(212, label = 'pycontrol-bode-phs', sharex=ax_mag)
+
     mags, phases, omegas, nyquistfrqs = [], [], [], []
     for sys in syslist:
         if (sys.inputs > 1 or sys.outputs > 1):
@@ -172,7 +186,6 @@ def bode_plot(syslist, omega=None, dB=None, Hz=None, deg=None,
 
             if (Plot):
                 # Magnitude plot
-                ax_mag = plt.subplot(211);
                 if dB:
                     pltline = ax_mag.semilogx(omega_plot, 20 * np.log10(mag), *args, **kwargs)
                 else:
@@ -186,7 +199,6 @@ def bode_plot(syslist, omega=None, dB=None, Hz=None, deg=None,
                 ax_mag.set_ylabel("Magnitude (dB)" if dB else "Magnitude")
 
                 # Phase plot
-                ax_phase = plt.subplot(212, sharex=ax_mag);
                 if deg:
                     phase_plot = phase * 180. / math.pi
                 else:
```

roryyorke: I've extended the above to the gang-of-four plots too: https://github.com/roryyorke/python-control/tree/rory/mplib-ax-warnings-2 ; this covers all of the subplot() calls in control/.  Successive calls to `bode_plot` and `gangof4_plot` will add to the existing axes, as desired.

However, this breaks examples/pvtol-nested.py, specifically, figure 6, which uses `subplot(211)` and `subplot(212)` to add annotations to the gain and phase plots.  This example will eventually break when Matplotlib changes subplot() as warned in the deprecation we're looking at.  (When running that example, Matplotlib warns that matplotlib.pyplot.hold() is going to disappear at some point too.)

I think to keep python-control working with the new Matplotlib way, we'll have to overhaul the plotting a bit; we should allow the user to get hold of the line objects (for creating legends), and to the axes (for adding lines, points, text, ...).  All of this is quite a bit of work, so perhaps we should leave it to after 0.8.0.

Separately, it might be good to add run the examples as part of Travis CI; and perhaps we should turn deprecation warnings into errors in all Travis CI runs.

murrayrm:
I've grabbed the changes above.  I think I am going to change the labels a bit and will also add a fix in `pvtol-nested.py`.

For the labels, I'm thinking that the format should be `control-plottype-subplotname`, so that we have `control-bode-magnitude` and `control-gangof4-t`, for example.

murrayrm:
I've put in a unit test to make sure the right number of lines are showing up on plots (crude, but would have caught the error that @roryyorke noted).  I also updated the axes label names, put in some documentation (code comments + user manual), and fixed the `pvtol-nested.py` example.

This should now be good to go, but would like someone else to have a look.

roryyorke:
Looks good to me.  I asked on Matplotlib-users about how we're using the labels [1], and the response was it's fine.  However, the recommendation was to not rely on the pyplot state machine (gcf, subplot, etc.).  I'll give this some thought, and open a new issue to discuss it.

[1] https://mail.python.org/pipermail/matplotlib-users/2018-January/001255.html

PR #193, opened 2018-02-19
Discrete omega-damping plot and tweaks

Sup3rGeo:
Other tweaks:
- Fixed zero markers (were invisible)
- Markers for pole and zero are black
- Width of y and x axis

### DIFF
--- a/control/pzmap.py
+++ b/control/pzmap.py
@@ -112,14 +112,14 @@ def pzmap(sys, Plot=True, title='Pole Zero Map'):
 
         # Plot the locations of the poles and zeros
         if len(poles) > 0:
-            plt.scatter(real(poles), imag(poles), s=50, marker='x')
+            plt.scatter(real(poles), imag(poles), s=50, marker='x', facecolors='k')
         if len(zeros) > 0:
             plt.scatter(real(zeros), imag(zeros), s=50, marker='o',
-                        facecolors='none')
+                        facecolors='none', edgecolors='k')
         # Add axes
         #Somewhat silly workaround
-        plt.axhline(y=0, color='black')
-        plt.axvline(x=0, color='black')
+        plt.axhline(y=0, color='black', lw=1)
+        plt.axvline(x=0, color='black', lw=1)
         plt.xlabel('Re')
         plt.ylabel('Im')
 
--- a/control/pzmap.py
+++ b/control/pzmap.py
@@ -40,10 +40,41 @@
 #
 # $Id:pzmap.py 819 2009-05-29 21:28:07Z murray $
 
-from numpy import real, imag
-from .lti import LTI
+from numpy import real, imag, linspace, pi, exp, cos, sin, sqrt
+from .lti import LTI, isdtime, isctime
 
 __all__ = ['pzmap']
+
+def zgrid(plt):
+    for zeta in linspace(0, 0.9,10):
+        factor = zeta/sqrt(1-zeta**2)
+        x = linspace(0, sqrt(1-zeta**2),200)
+        ang = pi*x
+        mag = exp(-pi*factor*x)
+        xret = mag*cos(ang)
+        yret = mag*sin(ang)
+        plt.plot(xret,yret, 'k:', lw=1)
+        xret = mag*cos(-ang)
+        yret = mag*sin(-ang)
+        plt.plot(xret,yret,'k:', lw=1)
+        an_i = int(len(xret)/2.5)
+        an_x = xret[an_i]
+        an_y = yret[an_i]
+        plt.annotate(str(zeta), xy=(an_x, an_y), xytext=(an_x, an_y), size=7)
+
+    for a in linspace(0,1,10):
+        x = linspace(-pi/2,pi/2,200)
+        ang = pi*a*sin(x)
+        mag = exp(-pi*a*cos(x))
+        xret = mag*cos(ang)
+        yret = mag*sin(ang)
+        plt.plot(xret,yret,'k:', lw=1)
+        an_i = -1 #int(len(xret)/2.5)
+        an_x = xret[an_i]
+        an_y = yret[an_i]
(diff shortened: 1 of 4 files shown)


control/pzmap.py
```diff
@@ -40,11 +40,42 @@
...
 # $Id:pzmap.py 819 2009-05-29 21:28:07Z murray $
 
-from numpy import real, imag
-from .lti import LTI
+from numpy import real, imag, linspace, pi, exp, cos, sin, sqrt
+from .lti import LTI, isdtime, isctime
 
 __all__ = ['pzmap']
 
+def zgrid(plt):
```
murrayrm: The zgrid function needs a docstring + some comments in the code to describe what is going on.

control/pzmap.py
```diff
@@ -40,11 +40,42 @@
...
+    for zeta in linspace(0, 0.9,10):
```
murrayrm: The range for zeta should probably be something with a default set of parameters that can be reset by the user.

control/pzmap.py
```diff
@@ -40,11 +40,42 @@
 #
 # $Id:pzmap.py 819 2009-05-29 21:28:07Z murray $
 
-from numpy import real, imag
-from .lti import LTI
+from numpy import real, imag, linspace, pi, exp, cos, sin, sqrt
```
murrayrm: Rather than using numpy.pi, please use math.pi to avoid problems with sphinx.  See commit aa6e0dfc05fde5e70d48f1add3f89e01adb1e8d5

control/pzmap.py
```diff
@@ -40,11 +40,42 @@
...
+        xret = mag*cos(ang)
+        yret = mag*sin(ang)
+        plt.plot(xret,yret, 'k:', lw=1)
+        xret = mag*cos(-ang)
+        yret = mag*sin(-ang)
+        plt.plot(xret,yret,'k:', lw=1)
+        an_i = int(len(xret)/2.5)
+        an_x = xret[an_i]
+        an_y = yret[an_i]
+        plt.annotate(str(zeta), xy=(an_x, an_y), xytext=(an_x, an_y), size=7)
```
murrayrm: Can/should we use parameters for the size?  7 seems like a magic number.

control/pzmap.py
```diff
@@ -75,16 +106,20 @@ def pzmap(sys, Plot=True, title='Pole Zero Map'):
 
     if (Plot):
         import matplotlib.pyplot as plt
+
+        if isdtime(sys):
```
murrayrm: This needs to be `isdtime(sys, strict=True)`, otherwise it will be true for systems with unspecified timebases (which could be continuous or discrete).

control/pzmap.py
```diff
@@ -75,16 +106,20 @@ def pzmap(sys, Plot=True, title='Pole Zero Map'):
...
+            zgrid(plt)
```
murrayrm: I think we should make plotting the zgrid an option, rather than the default.  This will give consistency with past behavior.

control/pzmap.py
Sup3rGeo: Ok! Do you have any example I could be based on?

murrayrm: Take a look at [nichols.py](https://github.com/python-control/python-control/blob/master/control/nichols.py) and, in particular, the `nichols_grid()` function.

control/pzmap.py
Sup3rGeo: I am actually having it similar to what we have on rlocus function, with an argument grid=False (even though IMHO plot is way better with the omega-damping grid)

I also realized we should also be using this on rlocus function, so I am making zgrid available for rlocus and making _sgrid_func() from rlocus available for pzmap.

control/pzmap.py
Sup3rGeo: making `zetas` and `wns` arguments so user can override default behaviour.

control/pzmap.py
Sup3rGeo: ok

control/pzmap.py
Sup3rGeo: So we are treating systems with unspecified timebases as continuous (showing s-plane omega-damping - see next comment)?

control/pzmap.py
murrayrm: Yes, this sounds like a good approach.  And if we can reuse an existing function, that is even better (although I note that `_sgrid_func()` is not all that well commented -:).

control/pzmap.py
murrayrm: The idea is that if the timebase is unspecified, it can represent either a continuous-time or discrete-time system (since for many operations it doesn't matter).  Because of this, if you call `isdtime` on a system for a system in which `timebase == None`, `isdtime` will return true.  If you want something to only happen when a system is *really* a discrete time system, use `isdtime(sys, strict=True)`.

Sup3rGeo:
Done some major refactoring on sgrid function based on matplotlib projections. It is not perfect though, as there is not much documentation available on using it, but it works well when moving the plot around and zooming. Please take a look at it in pzmap and rlocus plots and let me know what you think.

Currently the zgrid does not use projections so there is no update on zooming.

jwdinius:
I am new to this repo, so forgive me this question:  Why does this PR have so many commits?  As this PR seems to address plotting enhancements, wouldn't it be cleaner (and easier to track) to do a rebase and squash all commits into a single "plotting enhancements" commit?

Sup3rGeo:
@jwdinius just because this is my first contribution to an open-source project and no one suggested it so far :) Should I do this? I am actually waiting for a word from python-control maintainers

murrayrm:
Either way is OK.  5 commits is not _that_ many.

@Sup3rGeo: go ahead and squash if you want to give it a shot. 

@Abubakr95: I'm still waiting to find some time to review the changes and merge.

Sup3rGeo:
gentle ping

PR #192, opened 2018-02-19
Fix name errors in statesp module

Sup3rGeo:

PR #196, opened 2018-02-24
Fix namespace and logic errrors in statesp.freqresp + unit tests

murrayrm:
As pointed out in PR #192, there were some namespace errors in the statesp
module related to checking whether the frequency range as valid for discrete
time systems.  I added a unit test that covered this code (and exposed the
error) and fixed up the namespace problems (as in PR #192).  There was also
a logic error in the way that frequencies were checked and the warning message
referred to the wrong function.

### DIFF
--- a/control/statesp.py
+++ b/control/statesp.py
@@ -460,10 +460,8 @@ but B has %i row(s)\n(output(s))." % (self.inputs, other.outputs))
         if isdtime(self, strict=True):
             dt = timebase(self)
             cmplx_freqs = exp(1.j * omega * dt)
-            if ((omega * dt).any() > pi):
-                warn_message = ("evalfr: frequency evaluation"
-                                " above Nyquist frequency")
-                warnings.warn(warn_message)
+            if (max(omega) * dt > math.pi):
+                warn("freqresp: frequency evaluation above Nyquist frequency")
         else:
             cmplx_freqs = omega * 1.j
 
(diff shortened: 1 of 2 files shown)


control/tests/freqresp_test.py
```diff
@@ -107,6 +108,67 @@ def test_mimo(self):
...
+
+            # Convert to transfer function and test bode again
+            systf = tf(sys);
+            ret_tf = bode(systf)
+
+            # Make sure we can pass a frequency range
+            bode(sys, omega_ok)
+
+         else:
+            # Calling bode should generate a not implemented error
```
roryyorke: how about ```self.assertRaises(NotImplementedError, bode, (sys,))``` ?

control/statesp.py
```diff
@@ -460,10 +460,8 @@ def freqresp(self, omega):
         if isdtime(self, strict=True):
             dt = timebase(self)
             cmplx_freqs = exp(1.j * omega * dt)
-            if ((omega * dt).any() > pi):
-                warn_message = ("evalfr: frequency evaluation"
-                                " above Nyquist frequency")
-                warnings.warn(warn_message)
+            if (max(omega) * dt > math.pi):
```
roryyorke: perhaps ```max(abs(omega))```, to allow for negative omega?

nitpick: outermost parentheses are unnecessary.

Issue #177, opened 2018-01-03
place function fails for simple case


rabraker:
### The pole placement function `place` fails in some simple situations. 
This bug is related to the *current* implementation of `place`, which is a wrapper for slicot sb01bd. If PR [176](https://github.com/python-control/python-control/pull/176) is accepted, this bug report will refer to `place_varga` instead.

Example of failure:
```
import numpy as np
import control as pc
A = np.array([[0, 1],[100, 0]])
B = np.array([[0],[1]])
Pdes = np.array([-20 + 10*1j, -20 - 10*1j])
K = pc.place(A, B, Pdes)
```
This raises the error (with python 2.7)
"ValueError: an attempt was made to place a complex conjugate pair on the location of a real eigenvalue"

The problem is related to how the `alpha` parameter is computed when calling `sb01bd`, which is calculated as 
```
system_eigs = np.linalg.eig(A_mat)[0]
alpha = min(system_eigs.real);
```
Clearly, A has eigenvalues [-10, +10]. But np.lingalg.eigvals(A) reports 
eigvals = [10.000000000000002, -9.9999999999999982], and so we have set
```
alpha = -9.9999999999999982
```

`sb01bd` will only move eigenvalues that are greater than `alpha`. It seems that `sb01bd` must calculate the eigenvalues slightly differently than scipy, so it thinks it has only one eigenvalue to move (namely, the one at +10). This evidently corresponds to [error 4](http://slicot.org/objects/software/shared/doc/SB01BD.html).

If we fudge the `alpha` parameter, and call the sb01bd wrapper directly, this works as expected
```
import numpy as np
from slycot import sb01bd
A = np.array([[0, 1],[100, 0]])
B = np.array([[0],[1]])
Pdes = np.array([-20 + 10*1j, -20 - 10*1j])

alpha = min(np.linalg.eigvals(A).real)*2;
A_z,w,nfp,nap,nup,K,Z = sb01bd(B.shape[0], B.shape[1],
                               len(Pdes), alpha, A, B, Pdes, 'C')
print 'Placed Poles: %s'%np.linalg.eigvals(A + B.dot(K))
# yields on my system:
# Placed Poles: [-20.+10.j -20.-10.j]
````

### place() is not implemented for discrete time
[...]

PR #197, opened 2018-02-24
Fix place varga

rabraker:
This pull request address issue [#177 ](https://github.com/python-control/python-control/issues/177).

Summary of Changes:
1. Calculate the alpha parameter such it is always twice as small as the smallest (real part) open-loop eigenvalue. 

2. Add support for discrete time systems by exposing the DICO parameter

3. Expose the alpha parameter to the user, since it seems like interesting functionality to be able to place only the slowest eigenvalues.

4. Add tests to exercise these changes, including a test against the example in issue #177 .

### DIFF
--- a/control/statefbk.py
+++ b/control/statefbk.py
@@ -113,11 +113,11 @@ def place(A, B, p):
     return K
 
 
-def place_varga(A, B, p):
+def place_varga(A, B, p, DICO='C', alpha=None):
     """Place closed loop eigenvalues
-    K = place_varga(A, B, p)
+    K = place_varga(A, B, p, DICO='C', alpha=None)
 
-    Parameters
+    Required Parameters
     ----------
     A : 2-d array
         Dynamics matrix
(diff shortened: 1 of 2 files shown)


control/statefbk.py
```diff
@@ -159,24 +173,39 @@ def place_varga(A, B, p):
...
+    placed_eigs = np.array(p)
 
-    # SB01BD sets eigenvalues with real part less than alpha
-    # We want to place all poles of the system => set alpha to minimum
-    alpha = min(system_eigs.real);
+    if alpha is None:
+        # SB01BD ignores eigenvalues with real part less than alpha
+        # (if DICO='C') or with modulus less than alpha
+        # (if DICO = 'D').
+        if DICO == 'C':
```
murrayrm: What happens if the users passes something other than 'C' or 'D' for DICO?  There should probably be a final `else` that generates an error message.  Also, should we allow 'c' and 'd' (lower case) as well?

Alternatively, use `isdtime` as the option and then there are only two possibilities (True or not True).

control/statefbk.py
```diff
@@ -112,18 +112,32 @@ def place(A, B, p):
...
     A : 2-d array
         Dynamics matrix
     B : 2-d array
         Input matrix
     p : 1-d list
         Desired eigenvalue locations
+
+    Optional Parameters
+    ---------------
+    DICO : 'C' for continuous time pole placement or 'D' for discrete time.
```
murrayrm: The use of `DICO` as a parameter seems odd since it is not used in other functions.  A couple of possibilities:

* `_rss_generate` (internal function) uses `type`
* `gram` uses `dico` (lower case) as an internal variable
* We use use `isdtime` and `isctime` as functions to check for discrete and continuous time.  We could use `isdtime=False` as the option.

repagh: I am also in favour of isdtime, or maybe simply "dtime". The C / D character is a really Fortran-like approach.

murrayrm: Agree that `dtime=False` would be a better choice for the option (and default value).

murrayrm:
Thanks for the updates.  I think this looks good to go.  I will do some local testing and then merge in the coming days.

murrayrm:
Tested locally and looks fine.  Merging.

PR #200, opened 2018-04-18
Fix canonical form 'observable'

ipa-lth:
Minor fix to use 'solve' instead of 'inv'.
Asserts with prio 'inv' from (numpy.linalg.inv was also not imported, yet not needed anymore)

### DIFF
--- a/control/canonical.py
+++ b/control/canonical.py
@@ -133,7 +133,10 @@ def observable_form(xsys):
     Wrz = obsv(zsys.A, zsys.C)
 
     # Transformation from one form to another
-    Tzx = solve(Wrz, Wrx)  # matrix right division, Tzx = inv(Wrz) * Wrx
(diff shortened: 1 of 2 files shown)


control/canonical.py
```diff
@@ -133,7 +133,7 @@ def observable_form(xsys):
     Wrz = obsv(zsys.A, zsys.C)
 
     # Transformation from one form to another
-    Tzx = inv(Wrz) * Wrx
+    Tzx = solve(Wrz, Wrx)  # matrix right division, Tzx = inv(Wrz) * Wrx
```
roryyorke: Thanks, this looks right.  I see there are no tests for observable_form - would you mind adding one?

This would be left division in [Matlab's terminology](https://www.mathworks.com/help/matlab/ref/mldivide.html)

ipa-lth: you are right. See my update

PR #201, opened 2018-04-27
add canonical form 'modal'

ipa-lth:
The respective unittest is weak, yet I did not envisioned somthing better one


ipa-lth:
Hi,
I found something fishy in the transformation, but I cannot confirm if it is right or wrong yet!
After the transformation the poles of the system are identical, but the zeros seem to have doubled...

We should rule this out before merging.

Thx

ipa-lth:
I triple-checked and cannot find any error whatsoever. I will use it in my PhD for now. Additional test are welcome.

PR #204, opened 2018-05-17
Rlocus visual display of clicked point

icam0:
Added a visual emphasis on the clicked point in the Root locus plot while displaying the real and imagionary coordinates of the clicked point and the gain and the damping in the plot title.

root_locus


murrayrm:
Note: this PR has been included in PR #209.  If that PR is merged, can delete this one.

Issue #111, opened 2016-09-08
Too few poles for MIMO tf


roryyorke:
2x2 MIMO system I/s (diagonal, integrators on diagonal) is reported as having 1 pole.  Converting to state-space gives the correct number of poles.

tested with slycot https://github.com/python-control/Slycot/commit/a1f43abe0efadd2f03115ea0c990544627205230 and python-control cdd3e73838cea0b26a7155d0b2084aa698e5eefd

``` python
import numpy as np
import control

num = [ [ [1], [0] ],
        [ [0], [1] ] ]

den = [ [ [1,0], [1] ],
        [ [1],   [1,0] ] ]

g = control.tf(num,den)
print('g.pole()',g.pole())
print('control.ss(g).pole()',control.ss(g).pole())
```

output:

```
g.pole() [ 0.]
control.ss(g).pole() [-0. -0.]
```

murrayrm:
Took a quick look at this.  The problems seems to be in `control.TransferFunction._common_den`.  The method used to put together the common denominator for a MIMO transfer function is to create a list of all _unique_ poles that are found.  This obviously won't work in the case of repeated poles.  Definite bug.

repagh:
Does this still happen with my new implementation of _common_den ?

roryyorke:
I'll try to take a look this weekend. I see you have put the brakes on this
over at #194.

I see Matlab describe their MIMO tf pole algorithm here:
https://www.mathworks.com/help/control/ref/pole.html  (see "algorithms"
near the end)

I can't remember if that's what is intended here.

roryyorke:
Here's a moderately tricky case from Maciejowski's Multivariable Control, Example 2.2:
```python
num = [ [ [1],       [-1]      ],
        [ [1,1,-4],  [2,-1,-8] ],
        [ [1,-2],    [2,-4] ] ]

den = [ [ [1,3,2],   [1,3,2] ],
        [ [1,3,2],   [1,3,2] ],
        [ [1,1],     [1,1]   ]  ]

g = control.tf(num, den)
```

This has three poles: two at -1, and one at -2.  It also has a transmission zero at +2, which might be of interest as a test case for #205.

Matlab's pole(tf(num,den)) gives poles [-2,-2,-1,-1,-1] for this system (but pole(minreal(ss(tf(num,den)))) gives the right answer).

[...]

PR #206, opened 2018-05-29
TransferFunction _common_den rewrite - issue #194

repagh:
On my machines I got a consistent issue #194 error. (Mac OSX and Linux). 

After playing around with the seed in convert_test.py, I got slightly different results for the tests. I inspected _common_den (xferfcn.py) with the debugger, and found that in some cases a lot of cancelling poles would be added; the sorting for pole comparison did not consistently work. 

I then re-thought and re-worked the _common_den function, instead of sorting, cycling through all known poles, comparing to poles in the current denominator, marking which ones are found -- to prevent accepting double poles, and remembering which ones are not found.

The whole thing is a lot shorter, and I believe also a lot more robust. I hope this eliminates the elusive #194

### DIFF
--- a/control/statesp.py
+++ b/control/statesp.py
@@ -776,15 +776,14 @@ cannot take keywords.")
 
             # Change the numerator and denominator arrays so that the transfer
             # function matrix has a common denominator.
-            num, den = sys._common_den()
-            # Make a list of the orders of the denominator polynomials.
-            index = [len(den) - 1 for i in range(sys.outputs)]
-            # Repeat the common denominator along the rows.
-            den = array([den for i in range(sys.outputs)])
+            num, den, denorder = sys._common_den2()
+
             #! TODO: transfer function to state space conversion is still buggy!
-            #print num
-            #print shape(num)
-            ssout = td04ad('R',sys.inputs, sys.outputs, index, den, num,tol=0.0)
+            print("num", num.shape, "=", num)
+            print("den",den.shape,"=",den)
+            print("denorder", denorder)
+            ssout = td04ad('C', sys.inputs, sys.outputs, 
+                           denorder, den, num, tol=0.0)
 
             states = ssout[0]
             return StateSpace(ssout[1][:states, :states],
--- a/control/statesp.py
+++ b/control/statesp.py
@@ -775,11 +775,11 @@ cannot take keywords.")
 
             # Change the numerator and denominator arrays so that the transfer
             # function matrix has a common denominator.
-            num, den, denorder = sys._common_den2()
-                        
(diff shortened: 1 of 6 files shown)


repagh:
I average coverage actually decreased, simply because I made xferfcn.py 58 lines shorter and that file had a coverage wel above average.

repagh:
Please do not merge yet. I now realise that only poles common to an *input* can be removed from the tf's, (thanks to issue #111). I will try to resolve that over the next week or so.

murrayrm:
@repagh It would be great to add a unit test that catches issue #111 while you are at it.

repagh:
I think I got it this time. 
- The convert_test.py smoothly runs
- test case for #111 also works
- I modified/corrected a test case in xferfcn_test.py, same issue, deleted poles

Note that this relies on my fixes to Slycot; these fix a number of edge cases

murrayrm:
I fixed up the Travis CI issues in PR #210 and tested this code (locally) to make sure that it passes unit tests.  On MacOS 10.13 I get the following error:
```
ERROR: testModred (control.tests.matlab_test.TestMatlab)
----------------------------------------------------------------------
  File "/Users/murray/Dropbox/macosx/src/python-control/murrayrm/control/tests/matlab_test.py", line 400, in testModred
    modred(self.siso_ss3, [1], 'matchdc')
  File "/Users/murray/Dropbox/macosx/src/python-control/murrayrm/control/modelsimp.py", line 181, in modred
    raise ValueError("Matrix A22 is singular to working precision.")
ValueError: Matrix A22 is singular to working precision.
```
I haven't figured out what specifically in this PR is causing the problem, but need to chase this down before we can merge.

@repagh Can you run this against the latest version of `slycot` and see if you get the same error?

repagh:
I get the same error. 

However, I do not think this should be an issue. The modred is performed on a state-space system created with tf2ss, through slycot. One of the states is removed in modred, and since the tf2ss has been changed, a different set (of equally valid) states has come out of that step. Without control over what  states are chosen for a state-space system, one should not try to eliminate a specific single state. 

Using siso_ss1 instead of self.siso_ss3 for the test works.

repagh:
I also see a bunch of other tests failing (e.g. test_robust). I don't get that same behaviour locally.

murrayrm:
The other tests are failing due to changes that are addressed in PR #210.  I'll test locally and if everything is working then I'll merge PR #210 and then this PR, which should get us back to a working master branch.

murrayrm:
Confirmed that latest change (to `modred` unit tests) fixes the issue on my local machine.  PR #210 is now merged, so this PR is ready to go.

PR #209, opened 2018-06-03
Sisotool and dynamic root locus zoom

icam0:
This pull request introduces two new features to the python-control package. A matlab inspired sisotool and dynamic recomputing of the root locus plot when zooming.

**Sisotool**
In the pictures below a comparison can be seen between the matlab sisotool, the sisotool for matlplotlib 1.x.x and matlplotlib 2.x.x respectively. 

matlab
sisotool_matlab
matplotlib 1.x.x
sisotool_mpl1
matplotlib 2.x.x
sisotool_mpl2

The function is essentially a wrapper around the rootlocus, bode and step response functions. When a click is made on the rootlocus plot, the purple point(s) are moved and the other plots are updated as well. Note that a decision was made to keep most of the rootlocus plot update logic inside the rootlocus file and the logic behind updating the other 2 plots in _sisotoolupdate in sisotool.py. Also, use was made of **kwargs whenever neccessary in order to not confuse the end user with parameters in functions without documenting them. Also, this was tested on python 2.7, 3.3 and 3.6 and matplotlib 1.5.3 and 2.2.2 on mac os x. It would be greatly appreciated if more testers can try this on their machine.

**Dynamic zooming of the root locus plot **
Firstly, I want to point out an error in existing code in the following code snippet from _default_gains:
```
    while (indexes_too_far[0].size > 0) and (kvect.size < 5000):
        for index in indexes_too_far[0]:
            new_gains = np.linspace(kvect[index], kvect[index+1], 5)
            new_points = _RLFindRoots(num, den, new_gains[1:4])
            kvect = np.insert(kvect, index+1, new_gains[1:4])
            mymat = np.insert(mymat, index+1, new_points, axis=0)

```
Every iteration, 3 points are inserted at the index where the distance between 2 points is too large. What goes wrong however, is that the indexes are not updated. So when for example `indexes_too_far = [0,1,2]` and in the first iteration 3 points in between 0 and 1 are inserted. The next iteration the points will be inserted in between 1 and 2. However, this point has shifted its index with 3. This is fixed by the following:

```
    while (len(indexes_too_far) > 0) and (kvect.size < 5000):
        for counter,index in enumerate(indexes_too_far):
            index = index + counter*3
            new_gains = np.linspace(kvect[index], kvect[index + 1], 5)
            new_points = _RLFindRoots(num, den, new_gains[1:4])
            kvect = np.insert(kvect, index + 1, new_gains[1:4])
            mymat = np.insert(mymat, index + 1, new_points, axis=0)
```

Now a new function `_indexes_filt`  is added which also takes into account if the user is zoomed in on a part of the plot and adds more points to `indexes_too_far`. If no initial points are within the zoom view, it detects at which point one of the points crosses an xlim or ylim of the zoombox and then more points are added until the zoom tolerance requirements are met.

This pull request also includes the #199 and #204 pull requests, because they are an integral part of the sisotool as well. Please note that the Travis CI build sometimes fails due to issue #194 . I was able to get the build passing (https://travis-ci.org/icam0/python-control/builds/387446241) by restarting 2 subbuilds once.


repagh:
I tested locally, and this works fine!

control/timeresp.py
```diff
@@ -339,7 +339,7 @@ def forced_response(sys, T=None, U=0., X0=0., transpose=False):
 def _get_ss_simo(sys, input=None, output=None):
     """Return a SISO or SIMO state-space version of sys
 
-    If input is not specified, select first input and issue warning
+    If input is not sfpecified, select first input and issue warning
```
murrayrm: Please remove this change, which inserts a misspelling.

control/sisotool.py
```diff
@@ -0,0 +1,144 @@
...
+from .freqplot import bode_plot
+from .timeresp import step_response
+from .lti import issiso
+import matplotlib
+import matplotlib.pyplot as plt
+import warnings
+
+def sisotool(sys, kvect = None, xlim_rlocus = None, ylim_rlocus = None, plotstr_rlocus = 'b' if int(matplotlib.__version__[0]) == 1 else 'C0',rlocus_grid = False, omega = None, dB = None, Hz = None, deg = None, omega_limits = None, omega_num = None,margins_bode = True, tvect=None):
+
+    """Sisotool
```
murrayrm: Documentation style is a bit nonstandard.  I think the Sisotool on its own line can be removed.  This will look odd on `readthedocs` since the function name will already be listed.

control/sisotool.py
```diff
@@ -0,0 +1,144 @@
...
+
+    Sisotool style collection of plots inspired by the matlab sisotool.
```
murrayrm: MATLAB should be in all caps.

murrayrm:
Very nice tool.  I tried it out on Python 3.5 on MacOS 10.13 (High Sierra) and it worked well.

If someone can test this on Linux (and Windows?) and confirm that it works, we can incorporate into the next release.

murrayrm:
I tried out  the tool and it works most of the time.  But there can be problems if changing the gain if the scale of the plot is too large.  For example, try the following example:
```
import control
sys = control.tf([10,100], [1,5,9,5])
control.sisotool(sys)
```
This will generate  a  `matplotlib` window and you can click on the root locus branches on the right hand side to change the gain.  But if you try to click on the left branch (on the negative real axis), it doesn't get recognized.

I think the problem is that the scaling parameters that are used to determine if you are close enough to the axis don't work for the scaling on this plot.

@icam0 If there is no easy fix, we can merge this change as is and update later.  I'll wait for a few days to get your input.

icam0:
I suspect the gain tolerance to be too small for some reason (lines 455 and 457 of the rlocus.py file). I will get back to you during the weekend when I take a closer look at this.

icam0:
I have looked into it, what happens is that because this plot generates a very skewed aspect ratio due to the automatic matplotlib scaling of the plot the left side of the plot 'does not seem to work'. This is due to the fact that I made a decision to vary the gain tolerance according to the zoom and xlimit and ylimit of the plot. It selects the smallest of these 2 and bases the gain tolerance on the smallest of these 2 margins. When having a skewed aspect ratio it makes it hard for the user to click because the margin becomes very small. In the original implementation, there was a 'fixed' 0.04 factor in there which meant that if you would zoom in a lot, you could click very far away from the plot resulting in floating dots far away from the plot line. Using a scaled gain factor fixed this problem.

If you were to just increase the gain tolerance the behaviour becomes more unreliable and the point sometimes snaps to the incorrect line from then a user would intend. Therefore, for now, I'd like to propose to scale axes of the root locus plot to the same length of the smallest axis. Which makes sure the initial plot is always clickable. Something like this would then be added to the end of the rlocus.py plot (an edge case for handling when the user specifies the xlim or ylim would still need to be added):

```
x0, x1 = ax.get_xlim()
y0, y1 = ax.get_ylim()
if abs(x0-x1) >= abs(y0-y1):
    ax.set_xlim(0. - 0.5*abs(y0-y1) ,0. +0.5*abs(y0-y1))
else:
    ax.set_ylim(0. - 0.5 * abs(x0 - x1), 0. + 0.5 * abs(x0 - x1))
```

However, when the user zooms in with a very skewed aspect ratio clicking the line would still be very difficult which is why I think in the future a more sophisticated algorithm for click detection would be a worthwhile enhancement. I'd like to get your thoughts on this @murrayrm

murrayrm:
@icam0 Can you rebase this against the current master and resolve the conflicts.  I think this is ready to go into release 0.8.2 as soon as that is done.

murrayrm:
Did an inline correction of the conflicts.  Will merge if tests run correctly.

murrayrm:
Fixed up remaining issues.  Ready to merge.

PR #230, opened 2018-07-27
Fix pzmap not showing zeroes

geez0x1:
`pzmap` does not show the transfer function's zeroes, as the internal plotting argument has `facecolors='none'`:
```
   if len(zeros) > 0:
        plt.scatter(real(zeros), imag(zeros), s=50, marker='o',
                    facecolors='none')
```
Removing this argument solves the issue.


geez0x1:
screenshot from 2018-07-27 17-12-02

murrayrm:
Fixed in PR #193.

Issue #240, opened 2018-10-25
Erroneous time-domain sim (because of ss2tf) 


hungpham2511:
Hi, it seems python-control has a weird bug in the pipeline for converting from MIMO transfer function to state-space form. Consider the below gist

```python
import control as co
import numpy as np
from numpy import array

num = [[array([0.]), array([1.0])],
       [array([1.0]), array([0.0])]]


den = [[array([1.]), array([1, 1])],
       [array([  1.        , 4.0]), array([1.0])]]
sys = co.tf(num, den)
t, y = co.step_response(sys, input=0)
print("step response (output 1 from input 0) (old): {:}".format(y[1, -1]))

# slightly change the denominator
den = [[array([1.]), array([1e-10, 1, 1])],
       [array([  1.        , 4.0]), array([1.0])]]
sys = co.tf(num, den)
t, y = co.step_response(sys, input=0)
print("step response (output 1 from input 0) (changed): {:}".format(y[1, -1]))
```
The output on my computer is
```
step response (output 1 from input 0) (first): 0.24999999999982692
step response (output 1 from input 0) (second): 0.0
```

Clearly in the later case, there is a bug in the time-domain simulation function. The DC gain from input 1 to output 1 is 0.25, so the first simulation works alright. But in the second simulation, I obtain zero output. A preliminary investigation suggests that the problem is in the ss2tf procedure, particularly, in the slicot function.

Any suggestions would be great. Current I have to resort to matlab to do the transformation from transfer function to state-space in my application.

murrayrm:
Confirmed that this is a bug and that the problem is in the conversion to state space form.  The following code illustrates the error:
```
import control
import numpy as np
        
num =  [ [[0], [1]],           [[1],   [0]] ]
den1 = [ [[1], [1,1]],         [[1,4], [1]] ]
sys1tf = control.tf(num, den1)
sys1ss = control.tf2ss(sys1tf)

# slight perturbation
den2 = [ [[1], [1e-10, 1, 1]], [[1,4], [1]] ]
sys2tf = control.tf(num, den2)
sys2ss = control.tf2ss(sys2tf)

# Make sure that the poles match for StateSpace and TransferFunction
np.testing.assert_array_almost_equal(sys1tf.pole(), sys1ss.pole())
np.testing.assert_array_almost_equal(sys2tf.pole(), sys2ss.pole())
```
[...]

PR #242, opened 2018-11-04
fix issue #240: _common_den was not padding properly

murrayrm:
This PR fixes a problem where `xferfcn._common_den` was not properly padding the numerator coefficients and this was causing an apparent bug in `tf2ss` for MIMO systems (see issue #240).  The issue occurred in the following code in `xferfcn.py`:
```
m = npmax - len(numpoly)
if m < 0:
    num[i,j,::-1] = numpoly
else:
    num[i,j,:m:-1] = numpoly
```
where `npmax` is the maximum number of poles across all inputs.

What this code is doing is putting the coefficients for `numpoly` at the right end of the array `num[i,j]`, which has length `npmax+1`.  However, what `td04ad` looks at is only the portion of the numerator array that is the order of the denominator array for that *particular* input (not the max across all inputs).  So this code was shifting things too far to the right in the case where you had an input that had less than the maximum number of poles (discovered in issue #240, as part of a simulation bug).

The new code is
```
num[i, j, np+1-len(numpoly):np+1] = numpoly[::-1]
```
where `np` is the number of poles for this input (and hence `np+1` is the portion of the input array that `td04ad` will use.

In addition to fixing this bug, two unit tests were added:
* `control/tests/convert_test.py` has a test to make sure that `_common_den` does the right thing.
* `control/tests/timeresp_test.py` has a test that corresponds to the original bug identified in issue #240.

@repagh should have a look at this since it is in a section of code that he wrote.

### DIFF
--- a/control/tests/convert_test.py
+++ b/control/tests/convert_test.py
@@ -245,6 +245,29 @@ class TestConvert(unittest.TestCase):
         except ImportError:
             print("Slycot not present, skipping")
 
+    @unittest.skipIf(not slycot_check(), "slycot not installed")
+    def test_tf2ss_robustness(self):
+        """Unit test to make sure that tf2ss is working correctly.
+         Source: https://github.com/python-control/python-control/issues/240
+        """
+        import control
+        
+        num =  [ [[0], [1]],           [[1],   [0]] ]
+        den1 = [ [[1], [1,1]],         [[1,4], [1]] ]
+        sys1tf = control.tf(num, den1)
+        sys1ss = control.tf2ss(sys1tf)
+
+        # slight perturbation
+        den2 = [ [[1], [1e-10, 1, 1]], [[1,4], [1]] ]
+        sys2tf = control.tf(num, den2)
+        sys2ss = control.tf2ss(sys2tf)
+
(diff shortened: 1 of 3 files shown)


murrayrm:
Not able to replicate the unit test failure under MacOS.  Tried python3.5 and python3.7, but all unit tests pass.

Will have to start up a linux instance and see if I can generate the error.

murrayrm:
Tried recreating the error on Ubuntu linux (trusty), but all unit tests seemed to work.  ??

Also, `murrayrm/python-control:master`, which should just be a copy of `python-control/python-control:master` fails on unit tests for python 2.7 and 3.5, but is OK for 3.6 ([here](https://travis-ci.org/murrayrm/python-control/builds/450713817)).  Something funny going on.

repagh:
I checked out the PR and tested locally; looks good and correct results (Fedora 27). I also can't figure out why the tests fail on travis.

murrayrm:
Found the problem with Travis CI failures.  See issue #252 and PR #253.

Since @repagh OK'd the fix and we now know what was causing the Travis CI issue, plan to merge in the coming few days.

PR #247, opened 2018-11-26
added MIMO for TransferFunction.zero()

DDTantor:
Expanded TransferFunction.zero() to work on MIMO systems.

### DIFF
--- a/control/xferfcn.py
+++ b/control/xferfcn.py
@@ -584,12 +584,14 @@ has %i row(s)\n(output(s))." % (other.inputs, self.outputs))
 
     def zero(self):
         """Compute the zeros of a transfer function."""
-        if self.inputs > 1 or self.outputs > 1:
-            raise NotImplementedError("TransferFunction.zero is currently \
-only implemented for SISO systems.")
-        else:
-            #for now, just give zeros of a SISO tf
-            return roots(self.num[0][0])
+        sp = np.shape(self.num);
+        intersection = lambda a, b : [x for x in a if x in b];
+        zeros = roots(self.num[0][0]);
+        for i in range(sp[0]):
+            for j in range(sp[1]):
+                zeros = intersection(zeros, self.num[i][j]);
+
+        return zeros;
 
     def feedback(self, other=1, sign=-1):
         """Feedback interconnection between two LTI objects."""


control/xferfcn.py
```diff
@@ -584,12 +584,14 @@ def pole(self):
 
     def zero(self):
         """Compute the zeros of a transfer function."""
-        if self.inputs > 1 or self.outputs > 1:
-            raise NotImplementedError("TransferFunction.zero is currently \
-only implemented for SISO systems.")
-        else:
-            #for now, just give zeros of a SISO tf
-            return roots(self.num[0][0])
+        sp = np.shape(self.num);
```
murrayrm: Could remove semicolon in this line and the others (not needed in Python)

control/xferfcn.py
```diff
@@ -584,12 +584,14 @@ def pole(self):
...
+        intersection = lambda a, b : [x for x in a if x in b];
+        zeros = roots(self.num[0][0]);
+        for i in range(sp[0]):
+            for j in range(sp[1]):
+                zeros = intersection(zeros, self.num[i][j]);
```
murrayrm: Shouldn't this be the *union* (not intersection) and shouldn't the second argument be `roots(self.num[i][j])`?  

A simple test seems to fail for this code:
```
num = [[[1., 2.], [3., 4.]], [[5., 6.], [7., 8.]]]
den = [[[9., 8., 7.], [6., 5., 4.]], [[3., 2., 1.], [-1., -2., -3.]]]
sys = tf(num, den)
zero(sys)
```
returns
```
[]
```
instead of -2, -4/3, -6/5, -8/7.

murrayrm: The normal definition of zeros for a MIMO transfer function is that the matrix transfer function drops rank at a zero, not that the entire matrix evaluates to zero.  So you don't want "all coefficients of the matrix to be zero", I think.

See https://ocw.mit.edu/courses/electrical-engineering-and-computer-science/6-241j-dynamic-systems-and-control-spring-2011/readings/MIT6_241JS11_chap27.pdf for example.

murrayrm:
@DDTantor This PR still has problems and is now the master branch.  Can you please fix the problems identified above, and then rebase this against the current master.  If you are no longer actively developing this feature, let me know and we can delete this PR.

DDTantor:
> @DDTantor This PR still has problems and is now the master branch. Can you please fix the problems identified above, and then rebase this against the current master. If you are no longer actively developing this feature, let me know and we can delete this PR.

PR #248, opened 2018-12-12
Performance improvements to _remove_useless_states routine in statesp.py

adm78:
- avoid looping through states 
- significant performance improvement in the creation of large state space models


adm78:
Please let me know if you require any modifications to this PR. It would be great to get it merged into the master branch soon.

repagh:
Is np.concatenate correct? You could end up with states that are removed for both reasons. 
Maybe use np.union1d?
It seems to work now, but that relies on np.delete graciously accepting the indexes specified twice.

adm78:
I'm not sure it's possible but I've replaced the np.concatenate with np.union1d just to be safe. 

Looks like some of the tests are failing due to slycot import errors...

murrayrm:
Travis CI tests are fixed in PR #280.  Will merge that into master and then ask pending PRs to merge the change so that we can be sure things are working.

PR #259, opened 2018-12-30
Plot heat maps of state spaces

don4get:
Hello!

Thank you for maintaining this repo.

Here is a little contribution to plot heat maps of state spaces or matrices.
Example is available in appropriate folder.
While I got used to your coding convention, I also slightly cleaned `nichols.py`. If you prefer me to send two separate PR, just tell me :)


murrayrm:
It would be useful if you could split this into two PRs, as you suggest.  I have some questions about the use of heat maps; will post those once these have been split.

don4get:
Hello, I splitted the PR in two, feel free to ask questions about the feature presented here.
Happy coding!

murrayrm:
I'm not sure I understand the reasoning for having this function in the control systems toolbox.  Is there a reason that looking at the heat map for state space matrices is useful?  Since the matrices are not uniquely defined, seems like the heat map is kind of arbitrary.

murrayrm:
My inclination at this point is to drop this PR.  It seems to add functionality that is not specific to control systems (could be used for any matrix).  Marking as `wontfix` but leaving in place for a while in case others want to chime in.

PR #263, opened 2019-01-02
Make all examples Python 3 compatible

roryyorke:
This is built on #262.

It includes a Bash script to load all example scripts, and run them where possible (all scripts can be executed except tfvis.py).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,6 +75,7 @@ install:
 script:
   - 'if [ $SLYCOT != "" ]; then python -c "import slycot"; fi'
   - coverage run setup.py test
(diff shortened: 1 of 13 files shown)


murrayrm:
@roryyorke I get a failure on Travis CI on this PR.  When I pulled to my local account and then pushed to my personal repository, I got the error "Could not parse .travis.yml".  I think this is why there is no Travis CI info above.

roryyorke:
Where do you see the error?

I'm getting no Travis builds at all, but can't see why.  I thought maybe it was the webhook problem, though I see that only kicks in from February.

PR #266, opened 2019-01-04
Fix python2 print in examples

don4get:
This one is to answer to issue https://github.com/python-control/python-control/issues/208, concerning the `python2 print style`.

I realized the example testing the import of slycot was not working and I fixed it.

PR #267, opened 2019-01-05
Clean examples

don4get:
Since previous PR https://github.com/python-control/python-control/pull/266 came a little bit too late, this one only aims at cleaning 2 examples.


murrayrm:
I don't see what this PR is trying to accomplish.  Seems to just be renaming some files??

don4get:
Indeed, I agree that's not a critical MR.

Still, it does a little bit more than just renaming examples which original was confusing (test_feedback.py does not test feeback, but just controllability and observability).

In the example of pole placement, the original file full of commented code, without any explanation for the user nor any tips on how to use the library. Moreover, the commented code seems to imply that `ab01md` is the slycot routine for pole placement, which is not:
http://slicot.org/objects/software/shared/doc/AB01MD.html

So I cleaned commented code, I highlighted for new users of the library how they can try to use slycot with the useful `slycot_check`, and I compared scipy pole placement with slycot one.

To me, the example is a lot more clearer this way, don't you?

```# Eigenvalue placement
#from slycot import sb01bd
K = place(A, B, [-3, -2, -1])
print("Pole place: K = ", K)
print("Pole place: eigs = ", np.linalg.eig(A - B * K)[0])

# from slycot import ab01md
# Ac, Bc, ncont, info = ab01md('I', A, B)```


```# Python control may be used without slycot, for example for a pole placement.
# Eigenvalue placement
w = [-3, -2, -1]
K = place(A, B, w)
print("[python-control (from scipy)] K = ", K)
print("[python-control (from scipy)] eigs = ", np.linalg.eig(A - B * K)[0])

# Before using one of its routine, check that slycot is installed.
w = np.array([-3, -2, -1])
if slycot_check():
    # Import routine sb01bd used for pole placement.
    from slycot import sb01bd

    n = 3  # Number of states
    m = 1  # Number of inputs
    npp = 3 # Number of placed eigen values
    alpha = 1 # Maximum threshold for eigen values
    dico = 'D'  # Discrete system
    _, _, _, _, _, K, _ = sb01bd(n, m, npp, alpha, A, B, w, dico, tol=0.0, ldwork=None)
    print("[slycot] K = ", K)
    print("[slycot] eigs = ", np.linalg.eig(A + np.dot(B, K))[0])
else:
    print("Slycot is not installed.")```

Issue #255, opened 2018-12-23
Figure out a way to specific which algorithm to run when multiple choices are available


murrayrm:
See [discussion in PR #176](https://github.com/python-control/python-control/pull/176#issuecomment-355056802) about using the `algorithm` keyword to allow the user to chose the algorithm.

murrayrm:
PR #268 used `method' instead of `algorithm` (same as `scipy.signal`).  Might want to standardize on that.

PR #268, opened 2019-01-05
Refactor place method

don4get:
This PR aims to clean the API concerning pole placement, and add fallback strategy in case scipy or slycot are not installed.

There is also a little bit of style refactor and unit test of `test_statefbk.py` do not require np.matrix anymore, as it will soon be deprecated.

This PR is linked to this issue https://github.com/python-control/python-control/issues/255

### DIFF
--- a/control/exception.py
+++ b/control/exception.py
@@ -39,32 +39,44 @@
 #
 # $Id$
 
+
 class ControlSlycot(Exception):
     """Exception for Slycot import.  Used when we can't import a function
     from the slycot package"""
-    pass
+    def __init__(self, error):
+        print("slycot import failed")
+        print(error)
+
 
 class ControlDimension(Exception):
     """Raised when dimensions of system objects are not correct"""
     pass
 
+
 class ControlArgument(Exception):
     """Raised when arguments to a function are not correct"""
     pass
+
 
 class ControlMIMONotImplemented(Exception):
     """Function is not currently implemented for MIMO systems"""
     pass
 
+
 class ControlNotImplemented(Exception):
     """Functionality is not yet implemented"""
     pass
+
 
 # Utility function to see if slycot is installed
 def slycot_check():
     try:
         import slycot
-    except:
+    except ImportError as error:
+        print(error)
+        return False
+    except Exception as exception:
+        print(exception, False)
         return False
     else:
         return True
--- a/control/statefbk.py
+++ b/control/statefbk.py
@@ -1,10 +1,11 @@
-# statefbk.py - tools for state feedback control
-#
-# Author: Richard M. Murray, Roberto Bucher
-# Date: 31 May 2010
-#
-# This file contains routines for designing state space controllers
-#
+""" statefbk.py
+
+This file contains routines for designing state space controllers through state feedback control.
+"""
+
+__author__ = "Richard M. Murray"
+__contributor__ = "Roberto Bucher"
+
 # Copyright (c) 2010 by California Institute of Technology
 # All rights reserved.
 #
@@ -98,12 +99,15 @@ def place(A, B, p):
     # Convert the system inputs to NumPy arrays
     A_mat = np.array(A)
     B_mat = np.array(B)
-    if (A_mat.shape[0] != A_mat.shape[1]):
-        raise ControlDimension("A must be a square matrix")
+    if A_mat.shape[0] != A_mat.shape[1]:
+        raise ControlDimension("A must be a square matrix. "
+                               "rows: {} != columns: {}".format(A_mat.shape[0],
+                                                                A_mat.shape[1]))
 
-    if (A_mat.shape[0] != B_mat.shape[0]):
-        err_str = "The number of rows of A must equal the number of rows in B"
-        raise ControlDimension(err_str)
+    if A_mat.shape[0] != B_mat.shape[0]:
+        raise ControlDimension("The number of rows of A ({}) must equal "
+                               "the number of rows in B ({})". format(A_mat.shape[0],
+                                                                      B_mat.shape[0]))
 
     # Convert desired poles to numpy array
     placed_eigs = np.array(p)
@@ -171,14 +175,13 @@ def place_varga(A, B, p, dtime=False, alpha=None):
     try:
         from slycot import sb01bd
     except ImportError:
-        raise ControlSlycot("can't find slycot module 'sb01bd'")
+        raise ControlSlycot("Can't find slycot module 'sb01bd'.")
 
     # Convert the system inputs to NumPy arrays
     A_mat = np.array(A)
     B_mat = np.array(B)
-    if (A_mat.shape[0] != A_mat.shape[1] or
-        A_mat.shape[0] != B_mat.shape[0]):
-        raise ControlDimension("matrix dimensions are incorrect")
+    if A_mat.shape[0] != A_mat.shape[1] or A_mat.shape[0] != B_mat.shape[0]:
+        raise ControlDimension("Matrix dimensions are incorrect.")
 
     # Compute the system eigenvalues and convert poles to numpy array
     system_eigs = np.linalg.eig(A_mat)[0]
@@ -186,13 +189,13 @@ def place_varga(A, B, p, dtime=False, alpha=None):
 
     # Need a character parameter for SB01BD
     if dtime:
-        DICO = 'D'
+        dico = 'D'
     else:
-        DICO = 'C'
+        dico = 'C'
 
     if alpha is None:
         # SB01BD ignores eigenvalues with real part less than alpha
-        # (if DICO='C') or with modulus less than alpha
+        # (if DICO = 'C') or with modulus less than alpha
         # (if DICO = 'D').
         if dtime:
             # For discrete time, slycot only cares about modulus, so just make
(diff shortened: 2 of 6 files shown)


control/statefbk.py
```diff
@@ -422,22 +474,24 @@ def obsv(A, C):
...
+    O = cmat
     for i in range(1, n):
-        obsv = np.vstack((obsv, cmat*amat**i))
-    return obsv
+        O = np.vstack((O, cmat * amat ** i))
+    return O
+
 
-def gram(sys,type):
+def gram(sys, desired_computation_str):
```
murrayrm: This seems very cumbersome as can named parameter.  Was there a reason to switch from `type`?  If so, can we find something short (ideally one word)?

control/tests/test_statefbk.py
```diff
@@ -1,15 +1,16 @@
 #!/usr/bin/env python
 #
-# statefbk_test.py - test state feedback functions
+# test_statefbk.py - test state feedback functions
```
murrayrm: Not sure why we are changing the name of this file.  All other files in this directory have `_test` as a suffix on the file name.  Suggest we leave unchanged.

doc/control.rst
```diff
@@ -108,7 +108,6 @@ Control system synthesis
 .. autosummary::
    :toctree: generated/
 
-    acker
```
murrayrm: As mentioned above, suggest we leave `acker` as a publicly visible function.

murrayrm:
This PR has several changes, some of which seem fine but other that I have questions about.  See review comments.

control/statefbk.py
```diff
@@ -45,11 +46,11 @@
 from . import statesp
 from .exception import ControlSlycot, ControlArgument, ControlDimension
 
-__all__ = ['ctrb', 'obsv', 'gram', 'place', 'place_varga', 'lqr', 'acker']
+__all__ = ['ctrb', 'obsv', 'gram', 'place', 'lqr']
```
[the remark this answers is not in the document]
don4get: Sorry, I understand we should not change the API for compatibility reasons.

Still I find it a bit confusing to have multiple access points for pole placement.

I fixed it.

control/statefbk.py
```diff
@@ -61,6 +62,19 @@ def place(A, B, p):
         Input matrix
     p : 1-d list
         Desired eigenvalue locations
+    method : string - optional
+        Method used for pole placement: {"YT", "KNV0", "varga", "acker"}
+    dtime: optional (useful only if method=="varga")
+        False for continuous time pole placement or True for discrete time.
+        The default is dtime=False.
+    alpha: double scalar (useful only if method=="varga")
+       If DICO='C', then _place_varga will leave the eigenvalues with real
```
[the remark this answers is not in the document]
don4get: I fixed it in the sphinx comment description 👍

control/statefbk.py
```diff
@@ -91,29 +105,60 @@ def place(A, B, p):
...
-    K = result.gain_matrix
-    return K
-
+    if method == "acker":
+        return _acker(A, B, p)
+    if method == "varga":
+        try:
+            return _place_varga(A, B, p, dtime=dtime, alpha=alpha)
+        except ControlSlycot as error:
+            print("[Pole placement] Fallback strategy: using Tits-Yang method from scipy.")
```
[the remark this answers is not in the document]
don4get: Fixed 👍

control/statefbk.py
```diff
@@ -91,29 +105,60 @@ def place(A, B, p):
...
+                                                                         B_mat.shape[0]))
+
+        # Convert desired poles to numpy array
+        placed_eigs = np.array(p)
 
-def place_varga(A, B, p, dtime=False, alpha=None):
+        try:
+            result = place_poles(A_mat, B_mat, placed_eigs, method=method)
+        except ValueError as error:
+            print("[Pole placement] Redundant pole location. "
```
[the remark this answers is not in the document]
don4get: Done.

control/statefbk.py
```diff
@@ -160,39 +205,38 @@ def place_varga(A, B, p, dtime=False, alpha=None):
...
-        DICO = 'D'
+        dico = 'D'
     else:
-        DICO = 'C'
+        dico = 'C'
 
     if alpha is None:
         # SB01BD ignores eigenvalues with real part less than alpha
-        # (if DICO='C') or with modulus less than alpha
+        # (if DICO = 'C') or with modulus less than alpha
```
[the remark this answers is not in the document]
don4get: Done 💯

control/statefbk.py
```diff
@@ -205,23 +249,24 @@ def place_varga(A, B, p, dtime=False, alpha=None):
...
     # Return the gain matrix, with MATLAB gain convention
     return -F
 
-# Contributed by Roberto Bucher <roberto.bucher@supsi.ch>
-def acker(A, B, poles):
+
+def _acker(A, B, poles):
     """Pole placement using Ackermann method
 
+    Contributed by Roberto Bucher <roberto.bucher@supsi.ch>
```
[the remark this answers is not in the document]
don4get: Done. 👍

control/statefbk.py
don4get: I don't like the idea of a variable named as a build-in function: 
https://docs.python.org/3/library/functions.html#type

I suggest `gramian_type`, what do you think about it?

control/tests/test_statefbk.py
don4get: I saw in issue https://github.com/python-control/python-control/issues/95 that you plan to organise the unit tests as in numpy. In fact, in this library as well as in many others, tests are always prefixed with `test_`:
image

in this PR, as I refactored code related to state feedback, I also started to rename unit tests properly. Do you want me to make this change in a dedicated PR, for all other tests?

doc/control.rst
don4get: I reverted this change :) 

As I said above, I think it is a little bit misleading to a several entry points for a single functionality. It reduces the redundancy.

control/exception.py
```diff
@@ -39,32 +39,44 @@
 #
 # $Id$
 
+
 class ControlSlycot(Exception):
     """Exception for Slycot import.  Used when we can't import a function
     from the slycot package"""
-    pass
+    def __init__(self, error):
```
repagh: I don't think it is the right approach to immediately print this error reason. This creates an exception with a by-product.

This also does not call Exception.__init__, (which would normally store the error string) so if caught and printed later, the original exception reason has disappeared.

rabraker:
This may just be personal preference, but I don't particularly like the idea of placing the poles with a method other than what the user specified (i.e., the fallback strategies). In my view, there are several drawbacks to this:

1. The code to achieve this is rather complex. This makes it more difficult to reason about what the function will *actually do* and also makes testing all the possible code paths harder. For instance, the path `"varga" -> "YT" -> "acker"` is, as far as I can tell, not  exercised in the updated test suite.

2. The second problem is that the methods are not really equivalent. For example, if the user specified "varga", and that fails for some reason, now we call `place` with "YT". But unless `alpha is None`, the result will not be what the user wanted. Similarly for devolving to `acker` when "YT" fails, because "YT" is numerically stable but `acker` is not. This can matter a great deal for some systems. 

3. It is possible that this PR will break users code. It is not too far fetched to imagine a scenario (e.g., a parameter sweep) that relies on `place` producing an error when two desired pole locations become too close. This PR would break (or at least change the behavior of) such a script.

In short, if the only way to get the poles placed is by calling `place` with "acker" or whatever, I don't see the downside to just letting the user decide that.

One final concern is that, as it stands, the docstring is not really accurate any more, because it describes (in the "algorithms" and "limitations" sections) the YT algorithm, not Ackermann's method or Varga's method or KNV0.

control/statefbk.py
```diff
@@ -61,6 +63,18 @@ def place(A, B, p):
         Input matrix
     p : 1-d list
         Desired eigenvalue locations
+    method : string - optional
```
murrayrm: [Numpydoc](https://numpydoc.readthedocs.io) suggests the format "string, optional" to indicate optional arguments and also suggests that the default value be specified in the description of the parameter.

control/statefbk.py
```diff
@@ -91,26 +105,57 @@ def place(A, B, p):
...
-    # Convert desired poles to numpy array
-    placed_eigs = np.array(p)
+    if method == "acker":
+        return acker(A, B, p)
+    if method == "varga":
+        try:
+            return place_varga(A, B, p, dtime=dtime, alpha=alpha)
+        except ControlSlycot as error:
+            warnings.warn("[Pole placement] Fallback strategy: using Tits-Yang method from scipy.")
+            return place(A, B, p)
```
murrayrm: If this is kept, I suggest putting the method explicitly in the call to `place`.  Otherwise if the default is changed, the warning will not be accurate.

control/statefbk.py
```diff
@@ -91,26 +105,57 @@ def place(A, B, p):
...
+        try:
+            from scipy.signal import place_poles
+        except ImportError as error:
+            warnings.warn(error)
+            warnings.warn("[Pole placement] Fallback strategy: using Varga method from slycot.")
+            try:
+                K = place_varga(A, B, p, dtime=dtime, alpha=alpha)
+                return K
+            except ControlSlycot as error:
+                warnings.warn("[Pole placement] Fallback strategy: using Ackermanm method.")
```
murrayrm: Ackermann misspelled.

control/statefbk.py
```diff
@@ -205,19 +248,19 @@ def place_varga(A, B, p, dtime=False, alpha=None):
             # the same as what slicot thinks are the eigs. So we need some
             # numerical breathing room. The following is pretty heuristic,
             # but does the trick
-            alpha = -2*abs(min(system_eigs.real))
+            alpha = -2 * abs(min(system_eigs.real))
     elif dtime and alpha < 0.0:
-        raise ValueError("Need alpha > 0 when DICO='D'")
-
+        raise ValueError("Need alpha > 0 when dico='D'.")
```
murrayrm: This error message is confusing since `dico` is not the name of the parameter that is passed to the function (it is `dtime`).

control/tests/test_statefbk.py
murrayrm: Current practice in python-control is to (mainly) use `_test` at the end of the filename, so prefer to leave this as is (no need to change).

murrayrm:
@don4get There are some good changes in this PR, but needs to address the comment in this this thread.  Specifically:

* We need to rethink the "automatic" use of a backup strategy, as pointed out by @rabraker.  One possibility would be to allow `method` to be a list, where you try the next method in the list if the first fails.  Another would be to add a keyword `fallback` that would default to `False` (to preserve existing behavior) but that could be used to enable the use of fallback strategies.

* @repagh's comments on the the `ControlSlycot` exception should be addressed.  Note also that PR #305 makes the `ControlSlycot` exception derive from `ImportError` instead of `Exception`.

* Please revert `test_statefbk.py` back tot `statefbk_test.py` to be consistent with the predominant naming convention for test files in `tests/` (I understand this is different than `numpy`),

* Other small comments raised in the reviews should be resolved.

* Need to resolve conflicts against the current master branch.

murrayrm:
PR has been idle for > 6  months => stale.  @don4get I will close this PR in a few days unless you will have time to work on it soon.  No worries if not; just re-open when you have responded to the comments above.

PR #269, opened 2019-01-05
Feature/check dimension function

don4get:
Add a function to check dimensions between state space matrices.

It will allow to make this kind of check uniform within the library, in a future PR.

The function requires numpy arrays as inputs but I am not sure we want to force it. Can you give me insights?

I also added some unit tests, feel free to point out some cases I might have missed.

### DIFF
--- a/control/exception.py
+++ b/control/exception.py
@@ -46,7 +46,9 @@ class ControlSlycot(Exception):
 
 class ControlDimension(Exception):
     """Raised when dimensions of system objects are not correct"""
-    pass
+    def __init__(self, error):
+        print("[Control] Dimension error.")
+        print(error)
 
 class ControlArgument(Exception):
     """Raised when arguments to a function are not correct"""
--- a/control/statesp.py
+++ b/control/statesp.py
@@ -10,7 +10,7 @@ python-control library.
 
 # Python 3 compatibility (needs to go here)
 from __future__ import print_function
-from __future__ import division # for _convertToStateSpace
+from __future__ import division  # for _convertToStateSpace
 
 """Copyright (c) 2010 by California Institute of Technology
 All rights reserved.
@@ -166,9 +166,9 @@ class StateSpace(LTI):
         if 0 == self.states:
             # static gain
             # matrix's default "empty" shape is 1x0
-            A.shape = (0,0)
-            B.shape = (0,self.inputs)
-            C.shape = (self.outputs,0)
+            A.shape = (0, 0)
+            B.shape = (0, self.inputs)
+            C.shape = (self.outputs, 0)
 
         # Check that the matrix sizes are consistent.
         if self.states != A.shape[0]:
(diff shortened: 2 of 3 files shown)


murrayrm:
I don't understand why this functionality is useful?  It seems like it is just adding a function that is not likely to be used by users?

don4get:
Hello! Thank you for your review ;) 

As I said in the description of the PR, the point here is to create an internal function to make dimension check uniform in the library.

Currently, the check is partial and varies from a control function to another:
in statefbk.py -> place:
```
if A_mat.shape[0] != A_mat.shape[1]:
            raise ControlDimension("A must be a square matrix. "
                                   "rows: {} != columns: {}".format(A_mat.shape[0],
                                                                    A_mat.shape[1]))

        if A_mat.shape[0] != B_mat.shape[0]:
            raise ControlDimension("The number of rows of A ({}) must equal "
                                   "the number of rows in B ({})".format(A_mat.shape[0],
                                                                         B_mat.shape[0]))
```

in place_varga:
```
if A_mat.shape[0] != A_mat.shape[1] or A_mat.shape[0] != B_mat.shape[0]:
        raise ControlDimension("Matrix dimensions are incorrect.")
```

in lqr:

```
# Check dimensions for consistency
    nstates = B.shape[0]
    ninputs = B.shape[1]
    # TODO: Create / Figure out a generic function to check for inconsistent system
    if A.shape[0] != nstates or A.shape[1] != nstates:
        raise ControlDimension("Inconsistent system dimensions.")
```

And it continues like this in the entire library. 

With this new internal function, the check will be uniform, factorized and unit tested :)

control/exception.py
```diff
@@ -46,7 +46,9 @@ class ControlSlycot(Exception):
 
 class ControlDimension(Exception):
     """Raised when dimensions of system objects are not correct"""
-    pass
```
murrayrm: As discussed by @repagh in PR #268, this change has side effects.  Suggest leaving this alone since the default behavior should be enough.

control/statesp.py
```diff
@@ -1331,3 +1331,105 @@ def ssdata(sys):
     """
     ss = _convertToStateSpace(sys)
     return ss.A, ss.B, ss.C, ss.D
+
+
+def consistent_dimensions(A=None, B=None, C=None, D=None):
```
murrayrm: Should this be an internal function (`_consistent_dimensions`)?  Also, a more description function name might be useful.  Perhaps `_check_statespace_dimensions`?

control/statesp.py
```diff
@@ -1331,3 +1331,105 @@ def ssdata(sys):
...
+    Return a boolean stating if dimensions of the state-space matrices are consistent.
+
+    A state space representation should verify:
+    - A square matrix [n_states x n_states]
+    - B [n_states x n_inputs]
+    - C [n_outputs x n_states]
+    - D [n_outputs x n_inputs]
+    Parameters
+    ----------
+    A: ndarray - optional
```
murrayrm: Per [numpydocs](http://numpydoc.readthedocs.io), this should be "A : ndarray, optional".  Same below.

control/tests/statesp_test.py
```diff
@@ -443,11 +442,39 @@ def empty(shape):
...
             m.shape = shape
             return m
+
         np.testing.assert_array_equal(empty((0, 0)), g.A)
         np.testing.assert_array_equal(empty((0, D.shape[1])), g.B)
         np.testing.assert_array_equal(empty((D.shape[0], 0)), g.C)
         np.testing.assert_array_equal(D, g.D)
 
+    def test_consistent_dimensions_322(self):
+        true = consistent_dimensions(self.A322, self.B322, self.C322, self.D322)
```
murrayrm: A tighter format might be `self.assertTrue(_check_statespace_dimensions(self.A322, self.B322, self.C322, self.D322))`

murrayrm:
If we are going to include this PR, a couple of requests:

* Please see review comments and address as needed.

* I think we should also find the spots in the current code that would use this functionality and add it to those portions of the code.  That way we know the new functionality actually does something useful (and works in the places where we need it).

* Need to resolve conflicts with current master.

murrayrm:
PR has been idle for > 6  months => stale.  @don4get I will close this PR in a few days unless you will have time to work on it soon.  No worries if not; just re-open when you have responded to the comments above.

PR #270, opened 2019-01-06
Pep8 clean robust control

don4get:
**Obsolete description, new PR for this: https://github.com/python-control/python-control/pull/289**

I wanted to add this functionality to the library:
https://ch.mathworks.com/help/robust/ref/makeweight.html

Above the Matlab behavior, I added the possibility to compute higher orders weight functions.

The discrete behavior is not implemented yet as I don't know how to deal with it. 
I tried the naive approach, creating the laplace variable as follows: `s = tf([1, 0], [1], Ts)`. 
However, the DC gain of the resulted weight function is not correct, as one could have expected. 
If someone knows how to solve this problem, I would be glad to help him.

**New description:**
Pep8 cleaning of robust control, result of a split of the initial PR as asked by @murrayrm

### DIFF
--- a/control/robust.py
+++ b/control/robust.py
@@ -45,7 +45,8 @@ from .exception import *
 from .statesp import StateSpace
 from .statefbk import *
 
-def h2syn(P,nmeas,ncon):
+
+def h2syn(P, nmeas, ncon):
     """H_2 control synthesis for plant P.
 
     Parameters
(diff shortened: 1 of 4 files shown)


murrayrm:
This PR has a mixture of PEP8 fixes and a new function. Can we  refactor so that we just include the new functionality in this PR and leave the PEP8 fixes for later (see also discussion in closed PR #283).

don4get:
Thank you for the review, the split is done :)

PR #276, opened 2019-02-24
Add _repr_latex_ for Jupyter notebooks

alchemyst:
This allows better representation inside the browser:

image

PR #280, opened 2019-03-19
install scikit-build for testing with slycot

repagh:
We need scikit-build when building and testing with slycot, should solve the issue with PR #276


repagh:
Why doesn't this build on Travis? More patience needed?

murrayrm:
I tried this out on my version of the repository and as soon as I merged it into a branch that used to be calling Travis CI, it stopped running Travis CI tests.  

It looks like the problem is that there is a parsing error.  If I look at the request history on my local branch, I see the following error:
```
Could not parse murrayrm/python-control/.travis.yml@b35def48d67bbe27
```

repagh:
@murrayrm : thanks for the hint!

PR #284, opened 2019-04-01
Added an initial version for step_info

joaoantoniocardoso:
Hello,

failing on trying to follow some exercises of my control classes (that uses Matlab) I implemented this initial `step_info` version (and a Matlab compatible `stepinfo`) but because I don't have a licensed Matlab version I can't assert if everything is working nice.

In testing, I just use an example from Matlab help docs for `stepinfo` and added an arbiratry tolerance of 2%, is it acceptable?

Thanks!

### DIFF
--- a/control/matlab/timeresp.py
+++ b/control/matlab/timeresp.py
@@ -4,7 +4,7 @@ Time response routines in the Matlab compatibility package
 Note that the return arguments are different than in the standard control package.
 """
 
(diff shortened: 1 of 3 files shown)


murrayrm:
Looks good, including unit test.  Will merge with 0.8.2 after checking to make sure there are no conflicts with other changes that need to be resolved.

Issue #239, opened 2018-10-25
non-intuitive Time-Domain Simulation functions behavior


hungpham2511:
Hi, I notice that the simulation functions: `step_response`, `impulse_response`, for discrete time systems is not very intuitive, and is also inconsistent with matlab's behavior (I sort of have the habit of checking matlab documentation for a python-control function if there are suprises). 

Consider this snippet
```python
import control as co
import numpy as np
A1 = co.tf([1], [1, 6, 5])
A1 = co.c2d(A1, 0.5)
T, y = co.step_response(A1, np.arange(0, 5, 0.2))
```

we would expect `T` to be the time instances that correspond to values in `y` (matlab behavior), but it turns out not the case. I got `T.shape = (25,)` while `y.shape=(1, 10)`.  The same behavior can be observed for other time-domain functions for discrete-time systems (continuous-time systems are alright).

ilayn:
It is just giving you back the time array but since the sampling time of the system doesn't match the time increment it is resampling the bounds.

hungpham2511:
That's right. 

However, this behavior is not intuitive to users. I would prefer it to be the actual time-array that is used to compute the impulse response. This information can be used, say, for plotting. Also, with the current behavior, the returned time-array T does not convey any information.

murrayrm:
I agree that this behavior is non-intuitive and probably incorrect.  Marking as a bug to be fixed in the next release.

murrayrm:
I did a bit of looking into this.  Some observations and issues:

* The 'bug' in this code is something that is apparently inherited from the `scipy.signal.dlsim()`, which is what `step_response()` eventually calls (through a call to `forced_response()`).  In particular, it appears that if you pass `dlsim()` a time vector that doesn't match the sampling time, it doesn't generate an error and, furthermore, it returns whatever you passed it (same behavior as what was noticed here).

* There is a further inconsistency in the way that `T=None` is handled in the `control` simulation functions for discrete time systems: the time vector that is generated will have `dt=1` independent of what the actually sampling time for the system is.

* There may also be some issues with the way that the default response time is determined when you pass `T=None` for discrete time systems.  For `step_response()` and `initial_response()` this is supposed to be based on the dynamics matrix `A`.  This uses the `_default_response_times()` function from `scipy.signal.ltisys`, but I believe that function assumes continuous time (need to check).

I think that the correct behavior for discrete time systems is the following (which would match MATLAB):

[...]

PR #286, opened 2019-04-10
fix discrete-time simulation time output

murrayrm:
This PR fixes the non-intuitive time-domain simulation behavior for discrete time systems identified in issue #239.  In this version:

* The values of the output argument `T` always matches the times at which the corresponding outputs `Y` are obtained.  This is consistent with the idea that you plot the outputs of a simulation function by using commands of the form
```
t, y = step_response(sys)
plot(t, y)
```

* If you pass an explicit set of sampling times for a discrete time system, the time increment must be an integer multiple of the system sampling time (with `dt=True` allowing any time increment).  If not, a ValueError exception is raised.

* If you omit the sampling time, the value of `T` will be determined based on either the input signal (for `forced_response()` or the time constants coming from the system dynamics.  In either case, the difference between time points will be equal to the sampling time of the system (or 1 if `dt=True`).  (Note: net yet clear whether the time constant calculation is correct for discrete time systems; will create a new issue to track this.)

Unit tests are included to make sure all of the statements above are checked.

### DIFF
--- a/control/tests/timeresp_test.py
+++ b/control/tests/timeresp_test.py
@@ -45,6 +45,14 @@ class TestTimeresp(unittest.TestCase):
                       "0. 9. ")
         self.mimo_ss1 = StateSpace(A, B, C, D)
 
+        # Create discrete time systems
+        self.siso_dtf1 = TransferFunction([1], [1, 1, 0.25], True)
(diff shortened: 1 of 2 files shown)


repagh:
Nice work!

PR #293, opened 2019-04-17
Rework at step_info

joaoantoniocardoso:
Hi,

I did some rework trying to fix some issues that it found using `control.step_info(),` specially when I tried to compare with MATLAB for some system with inverted output like `sys = control.TransferFunction([0.0002, -0.0002], [1, 0.01, 0.0002])`.

Another thread that I found was that the output resolution for `control.step_response()` for discrete systems, so I increased it using the idea of MATLAB, that produces the output with *all* samples for a particular discrete system from 0 to steady state. Also, I noticed that MATLAB thinks differently about how to define the size of the arrays for step... but i didn't mess with it this time, it is working fine.

Besides that, I have no clue about what undershoot should be like.


joaoantoniocardoso:
Oops, I did break the tests for step, really missed somewhere. I'll look it soon and pull again.

PR #295, opened 2019-04-21
fix time series inconsistency between continuous and discrete time simulations

murrayrm:
As pointed out in issue #239, there is an inconsistency between the way that continuous time simulations and discrete time simulations return their outputs for single output systems.  In continuous time, the code follows the documentation and the following works:
```
t, y = step_response(sys)
plot(t, y)
```
However, in discrete time simulations the returned value for `y` had shape `(1, len(t))`, which meant that you had to do the following for discrete time systems:
```
t, y = step_response(sysd)
plot(t, y[0])
```

The problem was that for continuous time simulations `forced_response()` (the ultimate function that gets called) had an `np.squeeze()` at the end but this wasn't present for discrete time simulations. This PR fixes that problem so that the behavior is consistent, resolving the bug in issue #239.

Other small changes:

* Added unit tests that catch the original bug

* Added a `squeeze` keyword that controls whether this behavior is implemented (default = `True`).

* Did some PEP8 cleanup on `timeresponse.py`

Note that this may make some existing code break since discrete time simulations can now return an output that is a 1D array instead of 2D.  The fix is either to remove the output index (`y[0]` becomes `y`) or set the `squeeze` keyword to `False`).

### DIFF
--- a/control/timeresp.py
+++ b/control/timeresp.py
@@ -336,7 +336,7 @@ def forced_response(sys, T=None, U=0., X0=0., transpose=False,
 
     else:
         # Discrete type system => use SciPy signal processing toolbox
-        if (sys.dt is not True):
+        if sys.dt is not True:
             # Make sure that the time increment is a multiple of sampling time
 
             # First make sure that time increment is bigger than sampling time
--- a/control/timeresp.py
+++ b/control/timeresp.py
@@ -61,6 +61,7 @@ from .lti import isdtime, isctime
 __all__ = ['forced_response', 'step_response', 'step_info', 'initial_response',
            'impulse_response']
 
+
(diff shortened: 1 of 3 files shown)


control/timeresp.py
```diff
@@ -323,14 +332,11 @@ def forced_response(sys, T=None, U=0., X0=0., transpose=False,
...
             yout = dot(C, xout) + dot(D, U)
-
         tout = T
-        yout = squeeze(yout)
-        xout = squeeze(xout)
 
     else:
         # Discrete type system => use SciPy signal processing toolbox
-        if (sys.dt != True):
+        if (sys.dt is not True):
```
repagh: Any reason why there are brackets around the test?

PR #297, opened 2019-05-12
Fix sphinx warnings and use readthedocs theme for local doc generation

murrayrm:
This PR fixes up some warnings that were being generated by `sphinx` and changes the default theme used for generated the documentation on a local machine to match the ReadTheDocs theme that is used online (handy so that you know what things will look like when rendered on ReadTheDocs).

In fixing up one of the warning messages in `statesp.py`, I also took the opportunity to update other docstrings to be numpydoc compatible.

### DIFF
--- a/control/lti.py
+++ b/control/lti.py
@@ -54,8 +54,9 @@ class LTI:
 
         Parameters
         ----------
-        strict: bool (default = False)
-            If strict is True, make sure that timebase is not None
+        strict: bool, optional
+            If strict is True, make sure that timebase is not None.  Default 
+            is False. 
         """
 
         # If no timebase is given, answer depends on strict flag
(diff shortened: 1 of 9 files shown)


roryyorke:
Looks fine to me.

It's probably not important, but FYI: with Python 3.7 and Sphinx 2.0.1, the only python-sphinx variant I tried, I get this warning when building docs:
```
/home/rory/src/python-control/doc/generated/control.frd.rst: WARNING: document isn't included in any toctree
```

murrayrm:
Thought I found the problem (a stray "frd" in `matlab.rst` instead of "FRD"), but turns out the problem was more subtle:

On MacOS (where I usually work), the file system is "case preserving" but not "case sensitive".  This is a problem when you have two functions with the same name but different case.  In the `control` package, there is a *class* `FRD` and a *function* `frd`.  These conflict with each other in the generated/ subdirectory.

There was an error that was getting masked by this effect: in the documentation, we need to reference both `FRD` (when referring to the class) and `frd` (when referring to the function).  The reference to the function in the "system creation" portion of the manual was `FRD` when it should have been `frd`.  I've fixed that here, but it does cause a problem compiling the documentation on MacOS.

@roryyorke Can you check to see if this fix works for you (on windows or linux, I assume).

roryyorke:
No more warning due to FRD/frd, and I can see both control.frd and control.FRD in the generated docs.

PR #298, opened 2019-05-12
Replace py3.5 with py3.7 in Travis CI + update to xenial distribution

murrayrm:
This PR updates Travis CI to use python 2.7, 3.6 and 3.7 instead of 2.7, 3.5 and 3.6.  The main motivation is to start testing against python 3.7 while at the same time keeping the number of checks to be unchanged (9 variants total).

This also updates the linux distribution from `trusty` to `xenial` (required for Python 3.7).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 sudo: false
 language: python
+dist: xenial
+
+services:
+  - xvfb
 
 cache:
   apt: true
@@ -9,8 +13,8 @@ cache:
     - $HOME/.local
 
 python:
+  - "3.7"
   - "3.6"
-  - "3.5"
   - "2.7"
 
 # Test against multiple version of SciPy, with and without slycot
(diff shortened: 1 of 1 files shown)


roryyorke:
Looks good to me.

FWIW in the Slycot Travis setup we don't use the Travis-provided Python at all; we just install the desired version from Conda, so we didn't need to specify Xenial to get Python 3.7.

jgspiro:
> Looks good to me.
> 
> FWIW in the Slycot Travis setup we don't use the Travis-provided Python at all; we just install the desired version from Conda, so we didn't need to specify Xenial to get Python 3.7.

Hey, I found an error in the travis configuration for Slycot, where Python 3.5 and 3.6 are not tested at the moment.  I created a pull request to fix this, and an additional error in `setup.py` that prevents Slycot to be installed from source in a python 2.7 environment.  
Pull request : https://github.com/python-control/Slycot/pull/62

PR #301, opened 2019-05-19
initial release of I/O systems module

murrayrm:
This PR introduces a new module for representing (nonlinear) input/output systems and support for simulation, linearization, and composition.  The following features are supported:

* Input/output systems can be created using the `LinearIOSystem` and `NonlinearIOSystem` classes, and I/O systems can be simulated in continuous and discrete time using the `input_output_response()` function, which is equivalent to the `forced_response()` function for LTI system.

* Input/output systems can be composed using the existing `series`, `parallel`, and `feedback` functions, as well as using the `*` and `+` operators.  In addition, the `InterconnectedSystem` class can be used to interconnect a collection of I/O subsystems, with functionality similar to the `connect` function in MATLAB.

* Input/output systems and input/output signals can be named and Interconnections between I/O systems can be specified using system and signal names.

* Input/output systems can be linearized about an equilibrium point using the `linearize()` function, which returns a `LinearIOSystem` object.  The `LinearIOSystem` is derived from the `StateSpace` class, allowing all operations on linear systems to be applied to the linearized system.  The `find_eqpt` function can be used to find an equilibrium point for an input/output system.

* Documentation and examples are included to illustrate the use of the `iosys` module. 

For the most part, the changes in this PR are restricted to the new `iosys.py` module and the corresponding test and documentation files.  Some small modifications were required in other modules to support the features of this new module.

This PR is an initial release that implements fairly minimal functionality.  A list of additional features that should be added are included at the top of the `iosys.py` file.


murrayrm:
Since this PR is orthogonal to existing code (since it adds a new module), I am going to go ahead and merge this so that people can play around with it a bit.  (Also, I need it for the python code that goes along with the second edition of FBS...).

PR #307, opened 2019-06-08
Updated examples to be PEP compliant

billtubbs:
Made various edits to all the example scripts including:

- Replaced Python 2.7 `print` statements with `print(...)`
- Removed MATLAB-like formatting such as semi-colons
- Changed indenting and syntax to be PEP compliant (using PyCharm)
- Changed imports from `from ___ import *` to conventions such as `import matplotlib.pyplot as plt` and `import numpy as np`
- Removed tabs
- Added the following line to scripts that produce plot output:

```
if 'PYCONTROL_TEST_EXAMPLES' not in os.environ:
    plt.show()
```

Outstanding issues:
- Some examples still use out-dated `np.matrix`


examples/pvtol-lqr.py
```diff
@@ -21,35 +20,37 @@
...
+     [0, 0, 0, 0, 1, 0],
+     [0, 0, 0, 0, 0, 1],
+     [0, 0, (-ue[0] * np.sin(xe[2]) - ue[1] * np.cos(xe[2])) / m, -c / m, 0, 0],
+     [0, 0, (ue[0] * np.cos(xe[2]) - ue[1] * np.sin(xe[2])) / m, 0, -c / m, 0],
+     [0, 0, 0, 0, 0, 0]]
+)
 
 # Input matrix
-B = matrix(
+B = np.matrix(
```
murrayrm: Since the Numpy `matrix` class is being deprecated, it might make more sense to use `np.array` here, which should still work and won't require changes later.

Note, however, that if we make that change then we can no longer use `*` for matrix multiplication.  We have to use `dot` or `@` (python3 only).

examples/pvtol-nested-ss.py
```diff
@@ -36,102 +37,111 @@
...
+k, a, b = 200, 2, 50
+Ci = k*tf([1, a], [1, b])  # lead compensator
 
 # Convert to statespace
-Ci = tf2ss(Ci);
+Ci = tf2ss(Ci)
 
 # Compute the loop transfer function for the inner loop
-Li = Pi*Ci;
+Li = Pi*Ci
```
murrayrm: Add spaces around `*` for consistency with some of the formulas above?

examples/pvtol-nested-ss.py
```diff
@@ -36,102 +37,111 @@
...
 
 # Compute out the gain and phase margins
 #! Not implemented
 # (gm, pm, wcg, wcp) = margin(Li);
 
 # Compute the sensitivity and complementary sensitivity functions
-Si = feedback(1, Li);
-Ti = Li * Si;
+Si = feedback(1, Li)
+Ti = Li*Si
```
murrayrm: Is there a guideline for when to use spaces around `*` versus not?  I see a mixture in this file.

examples/pvtol-lqr.py
```diff
@@ -21,35 +20,37 @@
 #
 
 # System parameters
-m = 4;				# mass of aircraft
-J = 0.0475;			# inertia around pitch axis
-r = 0.25;			# distance to center of force
-g = 9.8;			# gravitational constant
-c = 0.05;	 		# damping factor (estimated)
+m = 4  # mass of aircraft
```
murrayrm: My personal preference is to have the comments lined up; makes them easier to read.

examples/pvtol-nested.py
```diff
@@ -37,110 +39,118 @@
...
+# In later versions of matplotlib the call to plt.subplot will clear the
 # axes and so we have to extract the axes that we want to use by hand.
 # In addition, hold() is deprecated so we no longer require it.
 #
-for ax in gcf().axes:
+for ax in plt.gcf().axes:
     if ax.get_label() == 'control-bode-magnitude':
         break
-ax.semilogx([1e-4, 1e3], 20 * np.log10([1, 1]), 'k-')
+    ax.semilogx([1e-4, 1e3], 20*np.log10([1, 1]), 'k-')
```
murrayrm: Should this be indented?  It means that it could get executed multiple times?

murrayrm:
@billtubbs Did a quick look through the changes.  Overall they look good!  Some small comments/questions for you to consider, but I think this is fine as is.  Will give you a chance to update in case any of the comments resonate, but then we can merge these in.

examples/pvtol-lqr.py
billtubbs: I can make that change now if you want.  I wasn't sure whether now is the right time.

examples/pvtol-nested-ss.py
billtubbs: PEP8 talks about using space around operators:

> operators with different priorities are used, consider adding whitespace around the operators with the lowest priority(ies). Use your own judgment; however, never use more than one space, and always have the same amount of whitespace on both sides of a binary operator.

Usually `*` and `/` have no space because they take higher priority over `+` and `-` but in a single statement I don't think this is relevant.  Personally I unless use spaces around `*` and `/` when they are between larger expressions (usually with parenthesis) so they don't appear joined together and to allow line-breaks.

I recommend removing space in this case (consistently throughout the file).

examples/pvtol-lqr.py
murrayrm: It's a good question about whether to do this now or later.  There is some discussion about how we are going proceed in issue #233 and PR #292.  Since this is existing (as opposed to new) code, it would be OK to leave.

On the other hand, it can be annoying to get warning message when you run an example (which is presumably what happens here, since Numpy is putting out DeprecationPending warnings whenever a new `matrix` object is created).

You are the author of the PR, so in the end you get to decide -:).

examples/pvtol-lqr.py
billtubbs: Agreed.

examples/pvtol-nested-ss.py
murrayrm: Sounds reasonable.  I see some inconsistencies, but it looks like they are in different files.  As long as the spacing in a given example file is consistent, seems OK.

billtubbs: I removed a bunch of space around `*` and `/` operators in various files.  Hopefully I caught them all.

examples/pvtol-nested.py
billtubbs: Oops!  I misunderstood that.  Will revert it back.

examples/pvtol-lqr.py
billtubbs: Turns out there were still tab characters all over the place, hence the misalignment.  Someone has their text editor set to allow tabs still maybe?

examples/pvtol-lqr.py
billtubbs: Turns out converting from `np.matrix` to `np.array` is not going to be that easy in all cases.  The problem with np.array is that when a result is one-dimensional (vector) it tends to reduce the result to shape (n, ) whereas matrices always remain 2-dimensional.

I made the conversion for two of the scripts but `pvtol-lqr.py` is going to take a bit more of a time investment so I propose leaving that for another day.

murrayrm: Agree, the updates are a bit of a pain (which is why we are delaying things in PR #292, in fact).  Seems fine to leave alone and we can handle them at a later date.

examples/pvtol-nested.py
murrayrm: Looks like this line is still indented and shouldn't be?

examples/pvtol-lqr.py
billtubbs: If this final push passes the tests then I think it's all ready to be merged.  I put some TODO comments in `pvtol-lqr.py` to help with the future conversion from array -> matrix.

Issue #310, opened 2019-06-09
rlocus generates a *lot* of data points


murrayrm:
While doing some development I noticed that running the root locus tests in the unit test suite takes a *long* time.  This occurs in `control/tests/matlab_test` and `control/tests/rlocus_tests.py`.  I did a bit of checking and it appears that the root locus command `rlocus` is generating 15K+ points in the root locus diagram, which can take a while...

Not sure when this crept in, but it would be nice for everyone (users, developers, Travis CI) to have the number of data points be a reasonable fraction of the number of pixels across a typical screen rather than a large multiple.

murrayrm:
More details:

* For the system
```
sys1 = TransferFunction([1, 2], [1, 2, 3])
rlist, klist = rlocus(sys1)
```
the gain calculation yields 183 points (reasonable).

* For the system
```
sys2 = StateSpace([[1., 4.], [3., 2.]], [[1.], [-4.]], [[1., 0.]], [[0.]])
rlist, klist = rlocus(sys2)
```
the gain calculation yields 12549 points (unreasonable).

murrayrm:
I chased down the problem to a calculation that is used to insert additional points in the root locus plots if the current set of points are too sparse.  There is a computation of the desired spacing on the x and y axes, and additional points are added if the initial set of points for the root locus are too far apart.  However, if the root locus plot is completely on the x axis (all roots are real), then the computation was returning a tolerance of zero and so *all* points on the root locus were being expanded by 5X => lots of new points.

I updated the calculation to look for the special case where the points are all in a line (vertical or horizontal) and then handled those as special cases.  This speeds up the root locus plot by quite a bit.

There is still some delay in setting up the zoom dispatcher.  This seems to be something associated with `matplotlib` and not our code.  I put in a TODO to look into it at some point.

PR coming soon (need to make sure the unit tests work on all platforms).

PR #313, opened 2019-06-09
Improve detection of when to add additional points to root_locus

murrayrm:
This PR provides improved checking of when to add points to a root locus plot and hence decreases the number of extraneous points that are generated.  See issue #310 for details.  The change is to detect when the root locus is on a line (typically the real axis) and to correct a calculation that was causing all points to be expanded (unnecessarily).

Also did some PEP8 and numpydoc cleanup while I was in the file.

### DIFF
--- a/control/rlocus.py
+++ b/control/rlocus.py
@@ -56,14 +56,14 @@ from .xferfcn import _convert_to_transfer_function
 from .exception import ControlMIMONotImplemented
 from .sisotool import _SisotoolUpdate
 from functools import partial
-from .lti import isdtime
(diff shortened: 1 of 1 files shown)


murrayrm:
@icam0, @gonmolina, @repagh Small fix to `root_locus`.  Since you have all worked on this file, would be great if you could take a quick look to make sure that what I am doing makes sense.

repagh:
There were many formatting changes, so I couldn't easily find the functional change. 
I threw the problems at it which were getting odd-angled root-loci with the old code (before this spring's fixes), and these were handled fine. 

Will add these as an example later.

PR #314, opened 2019-06-15
Allow np.array or np.matrix for state space matrices, operations

murrayrm:
This PR addresses issue #233 regarding (eventual) deprecation of the `numpy.matrix` class.  In this PR, state space objects are stored as either `numpy.matrix` (default) or `numpy.array` objects, with the selection determined based on the `use_numpy_array()` function.  

For backward compatibility, the `numpy.matrix` class is the current default, but all operations are compatible with using `numpy.array` as the default storage and output class and this will be the default when `numpy.matrix` is deprecated.

A set of unit tests is also included for testing basic operations when the `numpy.array` class is used as the state space storage and output class.  This PR does not convert all of the existing unit tests to use `numpy.array` instead of `numpy.matrix`, so there are still lots of warning messages.  At some point we will need to update all unit tests to work with `numpy.array` objects (as we as done in abandoned PR #292).

### DIFF
--- a/control/bdalg.py
+++ b/control/bdalg.py
@@ -326,7 +326,7 @@ def connect(sys, Q, inputv, outputv):
     >>> sys1 = ss([[1., -2], [3., -4]], [[5.], [7]], [[6, 8]], [[9.]])
     >>> sys2 = ss([[-1.]], [[1.]], [[1.]], [[0.]])
     >>> sys = append(sys1, sys2)
-    >>> Q = [[1, 2], [2, -1]])  # negative feedback interconnection
+    >>> Q = [[1, 2], [2, -1]]  # negative feedback interconnection
     >>> sysc = connect(sys, Q, [2], [1, 2])
 
     """
--- a/control/bdalg.py
+++ b/control/bdalg.py
@@ -53,7 +53,6 @@ $Id$
 
 """
 
-import scipy as sp
 import numpy as np
 from . import xferfcn as tf
 from . import statesp as ss
@@ -61,17 +60,18 @@ from . import frdata as frd
 
 __all__ = ['series', 'parallel', 'negate', 'feedback', 'append', 'connect']
 
+
 def series(sys1, *sysn):
     """Return the series connection (... \* sys3 \*) sys2 \* sys1
 
     Parameters
     ----------
-    sys1: scalar, StateSpace, TransferFunction, or FRD
-    sysn: other scalars, StateSpaces, TransferFunctions, or FRDs
+    sys1 : scalar, StateSpace, TransferFunction, or FRD
+    sysn : other scalars, StateSpaces, TransferFunctions, or FRDs
 
     Returns
     -------
-    out: scalar, StateSpace, or TransferFunction
+    out : scalar, StateSpace, or TransferFunction
 
     Raises
     ------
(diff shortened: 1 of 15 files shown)


control/bdalg.py
```diff
@@ -324,28 +323,30 @@ def connect(sys, Q, inputv, outputv):
 
     Examples
     --------
-    >>> sys1 = ss("1. -2; 3. -4", "5.; 7", "6, 8", "9.")
-    >>> sys2 = ss("-1.", "1.", "1.", "0.")
+    >>> sys1 = ss([[1., -2], [3., -4]], [[5.], [7]], [[6, 8]], [[9.]])
+    >>> sys2 = ss([[-1.]], [[1.]], [[1.]], [[0.]])
     >>> sys = append(sys1, sys2)
-    >>> Q = sp.mat([ [ 1, 2], [2, -1] ]) # basically feedback, output 2 in 1
+    >>> Q = [[1, 2], [2, -1]])  # negative feedback interconnection
```
roryyorke: trailing parenthesis here looks wrong.  Do we run these with doctest?

control/frdata.py
```diff
@@ -436,12 +440,13 @@ def feedback(self, other=1, sign=-1):
...
+            fresp[:, :, k] = np.dot(
+                self.fresp[:, :, k],
                 linalg.solve(
-                eye(self.inputs) +
-                other.fresp[:, :, k].view(type=matrix) *
-                self.fresp[:, :, k].view(type=matrix),
-                eye(self.inputs))
+                    eye(self.inputs)
+                    + np.dot(other.fresp[:, :, k], self.fresp[:, :, k]),
+                    eye(self.inputs))
```
roryyorke: I know this is not the point of this change, but is  `linalg(X, eye(shape))` any better than `np.linalg.inv`.  I'd have thought a "right divide" (`linalg.solve` with appropriate transposes) would have been a more standard approach.

control/mateqn.py
```diff
@@ -41,16 +41,17 @@
 Author: Bjorn Olofsson
 """
 
-from scipy import shape, size, asarray, asmatrix, copy, zeros, eye, dot
+from scipy import shape, size, array, asarray, copy, zeros, eye, dot
```
roryyorke: again not the point, but since you remove `import scipy as sp`, I think this should be `from numpy import ...`

control/mateqn.py
```diff
@@ -69,7 +70,9 @@ def lyap(A,Q,C=None,E=None):
         :math:`A X E^T + E X A^T + Q = 0`
 
     where Q is a symmetric matrix and A, Q and E are square matrices
-    of the same dimension. """
+    of the same dimension.
+
```
roryyorke: Not super-familiar with PEP8 and so on, but this extra blank line seems odd.

control/statefbk.py
```diff
@@ -252,13 +253,13 @@ def acker(A, B, poles):
 
     # Place the poles using Ackermann's method
     n = np.size(p)
-    pmat = p[n-1]*a**0
+    pmat = p[n-1] * np.linalg.matrix_power(a, 0)
```
roryyorke: `np.eye(a.shape)` ?  I suppose the `matrix_power` call sets the stage for the upcoming for-loop.

control/statefbk.py
```diff
@@ -252,13 +253,13 @@ def acker(A, B, poles):
...
     for i in np.arange(1,n):
-        pmat = pmat + p[n-i-1]*a**i
+        pmat = pmat + np.dot(p[n-i-1], np.linalg.matrix_power(a, i))
```
roryyorke: Again somewhat beside the point of this commit, but would it be possible to use Horner's method here?

control/bdalg.py
murrayrm: Fixed.

We are not currently using doctest, but I ran the revised version manually to make sure it was OK.

control/frdata.py
murrayrm: I agree that this could use `sp.linalg.solve` instead, but since I wasn't the original author I just did the conversion to remove the `np.matrix` dependence.  I've added a TODO for someone to fix this, so at least we don't forget about it (in principle).

control/mateqn.py
murrayrm: The extra blank line is a `numpydoc` standard: https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard

control/statefbk.py
murrayrm: Added a TODO for someone to do this in a separate PR.

murrayrm:
CI tests are failing due to issue #321.  Need to fix before merging.

Issue #299, opened 2019-05-13
Change FRD class to FrequencyResponseData?


murrayrm:
The `FRD` class name is non-pythonic (or at least non PEP8) and generates problems in compiling the documentation on MacOS (see PR #297).  If we change the name of the class to `FrequencyResponseData` this will be more consistent with the other classes (`StateSpace`, `TransferFunction`) and also remove this problem on MacOS.

Note that the `frd` function would still exist for creating frequency response data objects (like `ss` and `tf`).

murrayrm:
Addressed in PR #326

PR #326, opened 2019-07-11
update FRD class name to fix MacOS sphinx build problems

murrayrm:
As described in issue #299, when building the documentation on operating systems whose file systems are not case sensitive (eg, MacOS) the use of `FRD` and `frd` as different objects causes problems.  This PR changes the class for frequency response data (FRD) systems to be `FrequencyResponseData` (instead of `FRD`) while at the same time making `FRD` be an alias (copy) of the class => backwards compatible.  This fixing the documentation issue on MacOS and allows all previous code to run.

The only visible change is that the documentation now says `FrequencyResponseData` as the name of the class instead of `FRD` (which may be a feature).

### DIFF
--- a/control/frdata.py
+++ b/control/frdata.py
@@ -1,51 +1,47 @@
+# Copyright (c) 2010 by California Institute of Technology
+# Copyright (c) 2012 by Delft University of Technology
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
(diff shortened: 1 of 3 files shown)


murrayrm:
@repagh If you have a chance, it would be useful to get your input on this PR, which modifies the FRD class that you created quite a while back.

murrayrm:
Going ahead and merging this since it does not change the user interface.

PR #351, opened 2019-12-10
added preliminary version of lqe function, which calculates the gain …

sawyerbfuller:
…for a steady-state Kalman filter.

### DIFF
--- a/control/statefbk.py
+++ b/control/statefbk.py
@@ -43,10 +43,11 @@
 import numpy as np
 import scipy as sp
 from . import statesp
+from .mateqn import care
 from .statesp import _ssmatrix
 from .exception import ControlSlycot, ControlArgument, ControlDimension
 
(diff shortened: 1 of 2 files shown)


bnavigator:
Travis CI fails on some jobs for two different reasons:

- Python3.7 with slycot: missing symbol _dgebal. This has nothing to do with this PR and looks like a broken blas/lapack environment.

- Python 2.7: Invalid syntax using the `@` operator. Replace by `numpy.dot()` or wait until Python 2 support is dropped.

A unit test would not hurt.

sawyerbfuller:
Hi Ben, thanks for the suggestion. Added a unit test and python 2 compatibility

murrayrm:
Travis CI is failing because of `slycot` issue addressed in PR #353.  Checked via personal repository that all tests pass with that fix => ready to merge.

Issue #349, opened 2019-11-18
coverall reports coverage decrease despite tests being added


bnavigator:
My recent pull requests enhanced some unit tests, but coverall reports a decrease in coverage.

I believe this happens because it counts the results from the travis jobs without slycot. Most of the unittests are skipped without slycot.

bnavigator:
See the reported decrease for my last contribution to #345 (commit d26d6395288d7d22181cfcfcc73baced420f52c9):

[It is compared to the master branch](https://coveralls.io/builds/2560008) : "3875 of 4625 relevant lines covered (83.78%)"
[but now counts many more lines](https://coveralls.io/builds/27285682) : "26201 of 32361 relevant lines covered (80.96%)"

Something is broken with the coveralls setup.

murrayrm:
It looks like slycot may not be getting installed properly in the runs where it is supposed to be => code that relies on slycot is no longer being tested.  Looking to see what is wrong...

murrayrm:
Fixed in PR #353.

bnavigator:
Coverage still does not count the correct line numbers or compares to a wrong reference. See #366

bnavigator:
I don't think this is fixed by #353. With my recent PR #370, [coverall reports](https://coveralls.io/builds/28364668)

> 3952 of 4670 relevant lines covered (84.63%)

but on the [master build it is compared to](https://coveralls.io/builds/27948073)
> 34471 of 41958 relevant lines covered (82.16%)

PR #353, opened 2019-12-28
Fix slycot link error

murrayrm:
This PR address a problem in the `slycot` build process (documented [here](https://github.com/python-control/Slycot/pull/85)) in which building `slycot` from source is currently failing.  This is related to issue #349 where coveralls was reporting a decrease in coverage (because `slycot` was not present => some code was not being tested).

To fix this, I changed the `.travis.yml` file to use the `conda-forge` version of `slycot` by default.  This is probably the better thing to do in any case, since that  is how most users are likely to get `slycot`.  I also added two "optional" (allowed to fail) builds that create the `slycot` package from source code, as was originally done.  I put in one for python-2.7 (still working) and one for python-3.7 (failing).

I didn't squash the commits when  I created this PR, but suggest doing that when it is merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,13 +36,13 @@ jobs:
       dist: xenial
       services: xvfb
       python: "2.7"
-      env: SCIPY=scipy SLCOT=source
+      env: SCIPY=scipy SLYCOT=source
     - name: "linux, Python 3.7, slycot=source"
       os: linux
       dist: xenial
       services: xvfb
       python: "3.7"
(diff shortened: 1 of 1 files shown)


bnavigator:
The Python 3.7 ``slycot=source`` job fails because Travis mixes OpenBLAS with BLAS/LAPACK. See [my comment](https://github.com/python-control/Slycot/pull/85#issuecomment-569844824) in python-control/Slycot#85

Issue #377, opened 2020-03-04
arrow size in Nyquist plot


geekonloose:
I am experiencing weird issue. My code is as below. 
```
import control
import numpy as np
import matplotlib.pyplot as plt
from matplotlib.patches import *
num = 240
den = [1, 10, -11, -180]

# Creating a transfer function G = num/den
G = control.tf(num, den)
# w = np.logspace(-3, 3, 5000)
control.nyquist_plot(G, omega=None, Plot=True, labelFreq=0)
```
It creates wierd arrow size as shown in below figure. Is there any way to reduce the size of arrow?
![Figure_1](https://user-images.githubusercontent.com/30520900/75840675-c31a3e80-5df1-11ea-9750-b5edfed82d95.png)

murrayrm:
See discussion on issue #303

geekonloose:
Added in #379

PR #378, opened 2020-03-04
Nyquist_plot changed to accommodate arrow size

geekonloose:
Now functions is read like this: ```nyquist_plot(syslist, hl=0.1, hw=0.1, omega=None, Plot=True,   labelFreq=0, *args, **kwargs)```


geekonloose:
This is regarding issue #377

PR #379, opened 2020-03-04
Arrow head length and head width option is added in nyquist_plot function

geekonloose:


murrayrm:
It looks like the line style or something changed so that *every* line in the is marked as deleted then added.

Can you try to reset the line style back to the original so that we can see only those lines that changed?  Much easier to track updates that way.

geekonloose:
My editor was automatically formatting to pep8 format sorry for that. I have reformatted file as possible. Now it should read alright.

murrayrm:
Can you just go back to the original file and only edit the lines that actually need to be changed?  Just putting the lines changed in the PR comments won't let people see what actually changed if they do `git diff`.

murrayrm:
Looks like 8f8565b3b7d58544972ea2556fa6261ee1b31dc1 fixed the problem but then 1f4885d changed it back?

murrayrm:
Also: I suggest using keywords that are more descriptive than `hw` and `hl`.  How about `arrowhead_width` and `arrowhead_length`?

geekonloose:
ok I will do that.

PR #380, opened 2020-03-18
Switch to pytest and add optional Python 3.8 test

murrayrm:
This PR switches the unit testing to use `pytest` for Travis CI instead of the deprecated `setup.py test`.  It also adds an optional test for Python 3.8.


murrayrm:
For some reason the `allow_failures` tests are not being run.  This was working when they were originally generated (several months back).  Did something change in Travis CI?

Without the `allow_failures` functionality, the Python 3.8 test does not get run at all.

murrayrm:
Rebased changes on top of master to get some of the fixes from #366 the eased precision on some of the tests.

.travis.yml
```diff
@@ -118,7 +130,7 @@ install:
 # command to run tests
 script:
   - 'if [ $SLYCOT != "" ]; then python -c "import slycot"; fi'
-  - coverage run setup.py test
+  - coverage run -m pytest --disable-warnings control/tests
```
bnavigator: is `--disable-warnings` needed? I would rather have them printed and give every contributor the opportunity to fix problems early before e.g. `DeprecationWarnings` turn into errors after a Python or some library update.

murrayrm: There are 18,209 warning messages and unfortunately pytest prints error messages first => I found it hard to locate the errors.  Most of the warnings seem to be around the use of `numpy.matrix`, which I figure we will get rid of starting in v0.9.0.

control/tests/freqresp_test.py
```diff
@@ -236,7 +236,7 @@ def test_options(self):
 
 
 def suite():
-   return unittest.TestLoader().loadTestsFromTestCase(TestTimeresp)
+   return unittest.TestLoader().loadTestsFromTestCase(TestFreqresp)
 
 if __name__ == '__main__':
    unittest.main()
```
bnavigator: This is code from the unittest framework. Pytest does not use it. IMHO, let's remove `suite()` and the `__main__` sections from all test files.

murrayrm: I have removed the `suite()` functions, which were often not properly set up and were only used in `tests/run_all.py`, which I don't think we were using anyway.

I left in the `__main__` sections, so that you could (if you wanted) call a test script directly (rather than having to use `pytest`).

Issue #373, opened 2020-02-17
Bode plots use Hz instead of rad/sec in MATLAB compatibility mode


paulvicioso:
Hi everyone,

When I try the bode function from the control.matlab library to plot the Bode of a simple function (H(s)=1/(s+1)) I'm getting the wrong angular frequencies (shifted one decade). In other words, the break frequency is at 0.1 rad/s instead of 1.0 rad/s (see image). I'm not sure if this is an issue or I'm just doing something wrong. Could you please help me with this problem.

My system is a macOS Catalina 10.15.3 and I'm using Python 3.7.5.

```
import numpy as np
import control.matlab as ml
import matplotlib.pyplot as plt

num = np.array([1])
den = np.array([1, 1])

G = ml.tf(num, den)

print(G)

mag, phase, freq = ml.bode(G)
plt.show()
```

![Bode_plot](https://user-images.githubusercontent.com/8233543/74689704-8ffd6b80-51b2-11ea-9c22-4506daf9939a.png)

Thank you

murrayrm:
The units for frequency are in Hertz, not rad/sec.   1 rad/sec = 1/(2*pi) Hz = 0.16, which is where the breakpoint is (easiest to see as the point where the phase crosses 45 degrees).

It looks like this is a bug in the way that defaults are set up for the MATLAB compatibility module, which sets `bode.Hz = True`.  Can't remember if this has always been that way or whether the bug was introduced when the new `config` module was set up.

Quick fix: use `bode(..., Hz=False)`.

Proper fix: we should change the default settings to match MATLAB.

paulvicioso:
Thank you, that solved the problem.

I did not check that the frequencies were in Hz, I just made the assumption that everything was the same as in MATLAB.

Issue #291, opened 2019-04-16
sgrid() and zgrid() call matplotlib.pyplot.figure()


murrayrm:
While debugging some code I noticed that `examples/pvtol-nested.py` was producing a blank figure instead of a pole-zero map that was expected.  I chased this down tto the fact that the `sgrid()` function calls `matplotlib.pyplot.figure()`.  This is in contrast to all other plotting routines and the consequence is that you get a new set of axes.  This causes the following code (from `examples/pvtol-nested.py`) not to work as expected:
```
figure(10); clf();
(P, Z) = pzmap(T, Plot=True)
print("Closed loop poles and zeros: ", P, Z)

# Gang of Four
figure(11); clf();
gangof4(Hi*Po, Co);
```
Figure 10, which should have the output of `pzmap` will be blank since `pzmap` calls `figure()` (through `sgrid`) and creates a new set of axis (figure #11).  Figure #11 will then get erased (away goes the pole/zero map) and be replace by the Gang of 4.

billtubbs:
I'm wondering whether functions should be calling `plt.figure()` at all.  Shouldn't that be left to the highest-level function or even the user?  Likewise with `plt.show()`.  I think it’s better to leave the user with full control of where the plot is going and to allow additional formatting/additions.

How about we have an optional keyword argument, `ax` say, that is the matplotlib axis that the sub-function should add the plot to?  If `ax` is `None` then the function could create a new figure.

If there are other ways to do this please suggest them.  I can have a go at implementing this with `pzmap` and also fix `examples/pvtol-nested.py`.

billtubbs:
See `ax` argument in documentation below for example:

```
hist_series(self, by=None, ax=None, grid=True, xlabelsize=None, xrot=None, ylabelsize=None, yrot=None, figsize=None, bins=10, **kwds)
    Draw histogram of the input series using matplotlib.
    
    Parameters
    ----------
    by : object, optional
        If passed, then used to form histograms for separate groups
    ax : matplotlib axis object
        If not passed, uses gca()
    grid : boolean, default True
        Whether to show axis grid lines
    xlabelsize : int, default None
        If specified changes the x-axis label size
    xrot : float, default None
        rotation of x axis labels
    ...
```

Should we follow this approach?

murrayrm:
[...]

PR #382, opened 2020-03-21
Fix plot issues

murrayrm:
This PR fixes some problems with plotting functions  in `python-control`:

* Fixed sgrid, zgrid to use existing axes if they exist (fixes issue #291)
* Use Hz=False for MATLAB + docstring corrections (fixes issue #373)
* Fixed `gangof4` when `dB=True`, which was not using `semilogx` for plotting
* Changed `Plot` and `PrintGain` keywords to `plot` and `print_gain`, with deprecation warnings (better consistency with the rest of `python-control`).
* Removed conversion to state space in `examples/pvtol_nested.py` that was giving spurious zero at infinity
* PEP8 cleanup

### DIFF
--- a/control/freqplot.py
+++ b/control/freqplot.py
@@ -80,7 +80,7 @@ _bode_defaults = {
 
 
 def bode_plot(syslist, omega=None,
-              Plot=True, omega_limits=None, omega_num=None,
+              plot=True, omega_limits=None, omega_num=None,
               margins=None, *args, **kwargs):
     """Bode plot for a system
 
@@ -100,7 +100,7 @@ def bode_plot(syslist, omega=None,
     deg : bool
         If True, plot phase in degrees (else radians).  Default value (True)
         config.defaults['bode.deg']
-    Plot : bool
+    plot : bool
         If True (default), plot magnitude and phase
     omega_limits: tuple, list, ... of two values
         Limits of the to generate frequency vector.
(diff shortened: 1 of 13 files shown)


control/freqplot.py
```diff
@@ -110,11 +110,11 @@ def bode_plot(syslist, omega=None,
...
         If True, plot gain and phase margin.
-    *args
-        Additional arguments for :func:`matplotlib.plot` (color, linestyle, etc)
-    **kwargs:
+    *args : `matplotlib` plot positional properties, optional
+        Additional arguments for `matplotlib` plots (color, linestyle, etc)
+    **kwargs : `matplotlib` plot keyword properties, optional
         Additional keywords (passed to `matplotlib`)
-
+w
```
bnavigator: ?

control/freqplot.py
```diff
@@ -429,12 +437,13 @@ def gen_zero_centered_series(val_min, val_max, period):
...
         return mags, phases, omegas
 
+
 #
 # Nyquist plot
 #
 
-def nyquist_plot(syslist, omega=None, Plot=True,
-                 labelFreq=0, arrowhead_length=0.1, arrowhead_width=0.1, 
+def nyquist_plot(syslist, omega=None, plot=True, labelFreq=0,
```
bnavigator: change to `label_freq` for the same reason as the change to `plot` and `print_gain`?

Issue #374, opened 2020-02-20
Bug in step_response()


rlegnain:
Hi, 
I found a bug in step_response(). An error happens when use step_response() on G(s)=1/1. Here is the code:
```
    u = control.TransferFunction([1], [1])  # or control.tf([1], [1])
    t, y = control.step_response(u)   # error happen in this line. 
```

When I use scipy, it works perfectly

 The error message is:
 ```
 File "C:\Users\rlegnain\AppData\Roaming\Python\Python35\site-packages\control\timeresp.py", line 455, in step_response
    T = _default_response_times(sys.A, 1000)
  File "C:\Users\rlegnain\AppData\Roaming\Python\Python35\site-packages\scipy\signal\ltisys.py", line 2058, in _default_response_times
    vals = linalg.eigvals(A)
  File "C:\Users\rlegnain\AppData\Roaming\Python\Python35\site-packages\scipy\linalg\decomp.py", line 769, in eigvals
    homogeneous_eigvals=homogeneous_eigvals)
  File "C:\Users\rlegnain\AppData\Roaming\Python\Python35\site-packages\scipy\linalg\decomp.py", line 233, in eig
    compute_vr=compute_vr)
  File "C:\Users\rlegnain\AppData\Roaming\Python\Python35\site-packages\scipy\linalg\lapack.py", line 617, in _compute_lwork
    "%d" % (info,))
ValueError: Internal work array size computation failed: -5

```

**Comment by bnavigator** [comment]
Can reproduce (with a different Exception raised). Properly formatted (Use the "insert code" function of github when posting code:

```
In [1]: import control

In [2]: u = control.TransferFunction([1],[1])

In [3]: u
Out[3]: 

1
-
1


In [4]: t, y = control.step_response(u)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-4-7a899e9160fd> in <module>
----> 1 t, y = control.step_response(u)

/usr/lib/python3.7/site-packages/control/timeresp.py in step_response(sys, T, X0, input, output, transpose, return_x, squeeze)
    514     if T is None:
    515         if isctime(sys):
--> 516             T = _default_response_times(sys.A, 100)
    517         else:
    518             # For discrete time, use integers

/usr/lib64/python3.7/site-packages/scipy/signal/ltisys.py in _default_response_times(A, n)
   2056     # TODO: This could use some more work.
   2057     # For example, what is expected when the system is unstable?
[...]

PR #383, opened 2020-03-22
fix computation of default response time for constant systems

murrayrm:
Fixes issue #374 by added a check for constant systems (`sys.A.shape == (0, 0)`) and setting the default response time for those systems to either 1 second (continuous system) or 100 time steps (discrete system).

### DIFF
--- a/control/timeresp.py
+++ b/control/timeresp.py
@@ -512,20 +512,7 @@ def step_response(sys, T=None, X0=0., input=None, output=None,
     """
     sys = _get_ss_simo(sys, input, output)
     if T is None:
-        if isctime(sys):
-            if sys.A.shape == (0, 0):
-                # No dynamics; use the unit time interval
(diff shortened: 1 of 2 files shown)


control/timeresp.py
```diff
@@ -698,11 +715,19 @@ def initial_response(sys, T=None, X0=0., input=0, output=None,
...
+                T = _default_response_times(sys.A, 100)
         else:
             # For discrete time, use integers
-            tvec = _default_response_times(sys.A, 1000)
-            T = range(int(np.ceil(max(tvec))))
+            if sys.A.shape == (0, 0):
+                # No dynamics; use 100 time steps
+                T = range(100)
+            else:
+                tvec = _default_response_times(sys.A, 100)
```
bnavigator: length reduced from 1000 to 100

Issue #391, opened 2020-04-18
input_output_response for discrete time system confuses state with derivative


salomonsters:
In iosys, function input_output_response supports both continuous time and discrete time system. However, the discrete-time system simulation doesn't work as intended.

Compare the following blocks:
```
    def ivp_rhs(t, x): return sys._rhs(t, x, u(t))

    # Perform the simulation
    if isctime(sys):
        if not hasattr(sp.integrate, 'solve_ivp'):
            raise NameError("scipy.integrate.solve_ivp not found; "
                            "use SciPy 1.0 or greater")
        soln = sp.integrate.solve_ivp(ivp_rhs, (T0, Tf), X0, t_eval=T,
                                      method=method, vectorized=False)
```
with the code below it for discrete-time:
```
        for i in range(len(T)):
            # Store the current state and output
            soln.y.append(x)
            y.append(sys._out(T[i], x, u(T[i])))

            # Update the state for the next iteration
            x = sys._rhs(T[i], x, u(T[i]))
```
Both blocks use sys._rhs, however, in a different way. In the continous-time it is used as an argument of scipy's solve_ivp, where the first argument should be the _differential equation._ So far, so good.

However, for the discrete-time system, the vaule sys._rhs is directly assigned to the _state_ (x). I believe that state update should be:
`x = x + sys._rhs(T[i], x, u(T[i]))*dt`

murrayrm:
For discrete time I/O systems, the system dynamics are of the form
```
x[k+1] = f(x[k], u[k])
y[k] = h(x[k], u[k])
```
so the right-hand side is not the difference of the next state from current state, but the actual next state.  This is consistent with the way that discrete time state space are handled in `python-control`, which is inherited from `spicy.signal.dlsim`.  This error shows up in PR #392, where unit tests are failing.

PR #392, opened 2020-04-18
Fix issue #391

salomonsters:


murrayrm:
Good catch.  A couple of suggestions:

* Can you add a unit test that catches the previous error (eg, compare output from `iosys` step response for a linear system to what you would get from `step_response`) and demonstrates that the new one is correct.

* Will your code work if `dt=True` (rather than a specific sampling time)?  This should also be part of the unit test.

I can take care of these if you don't have time/interest, though it will take me a while to get to them.

murrayrm:
Also: unit tests are failing, so there is something wrong in the current code.

murrayrm:
Discrete time systems are modeled as
```
x[k+1] = f(x[k], u[k])
```

PR #397, opened 2020-05-08
Enter Function which can act similar to Matlab input function and can handle integer, float or arrays 

nabanita-sarkar:
## Enter Function
A simple function replicating the Matlab input function.
In Matlab the built-in input function handles integer, float or arrays and treat all of then as arrays.
But in python there is no such built-in function. In control system analysis this function can be helpful.
### Demo
     a = enter()
### Input format: 
- Single Integer or float
  Eg: 5
- Integers or floats seperated by comma
 Eg: 8,9  
 - List of integers or floats
  Eg: [6,7]

### DIFF
--- a/control/pzmap.py
+++ b/control/pzmap.py
@@ -81,7 +81,7 @@ def pzmap(sys, Plot=True, grid=False, title='Pole Zero Map'):
         The system's zeros.
     """
     # Get parameter values
-    Plot = config._get_param('rlocus', 'Plot', grid, True)
+    Plot = config._get_param('rlocus', 'Plot', Plot, True)
     grid = config._get_param('rlocus', 'grid', grid, False)
     
     if not isinstance(sys, LTI):


murrayrm:
If we are going to add a function like this, then it seems to me that:
* It should go in the `matlab` module, since it is trying to emulate a MATLAB capability
* It should match the name and syntax of the MATLAB function
```
Syntax
x = input(prompt)
str = input(prompt,'s')

Description
x = input(prompt) displays the text in prompt and waits for the user to input 
a value and press the Return key. The user can enter expressions, like pi/4 
or rand(3), and can use variables in the workspace.
If the user presses the Return key without entering anything, then input 
returns an empty matrix.
If the user enters an invalid expression at the prompt, then MATLAB displays 
the relevant error message, and then redisplays the prompt.

str = input(prompt,'s') returns the entered text, without evaluating the input 
as an expression.
```

In addition, the following items should be addressed:
* There needs to be a doctoring that describes the use of the function, using `numpydoc` standards
* There needs to be a unit test that confirms the correct functionality, including handling error conditions.

Given that `python` already has a perfectly good `input` input, I'm not sure there is enough value in this function to add it.  If we do add it, the code itself should be more pythonic (eg, a single line for the regular expression parsing, a list comprehension for the final construction of the matrix.

murrayrm:
No response on this for many months, so I'm going to close it out.

PR #400, opened 2020-05-14
Fixed InterconnectedSystems name bugs.

samlaf:
3 fixes to InterconnectedSystem's init:

1. It wasn't setting the name properly
2. It also wasn't setting the output and input names properly
3. I also made it so that instead of using generic state names, it uses subsys.statename when available

For example, if we Interconnected these two systems
```
System: predprey
Inputs (1): u, 
Outputs (2): H, L, 
States (2): H, L, 

System: control
Inputs (3): Ld, u1, u2, 
Outputs (1): y[0], 
States (0):
```

we would get
```
System: (None)               # point (1)
Inputs (0): 
Outputs (3):                 # point (2)
States (2): x[0], x[1],      # point (3)
```

Now instead, we get
```
System: io_closed
Inputs (0): 
Outputs (3): y[0], y[1], y[2], 
States (2): predpreyH, predpreyL, 
```


control/iosys.py
```diff
@@ -928,18 +930,21 @@ def __init__(self, syslist, connections=[], inplist=[], outlist=[],
             sysobj_list.append(sys)
             sysname_list.append(sys.name)
 
+        # This might cause a problem if there are duplicate names...
```
bnavigator, line 933: The block above makes sure there are no duplicate sysnames and you construct the statenames with a sysname prefix.

control/iosys.py
```diff
@@ -928,18 +930,21 @@ def __init__(self, syslist, connections=[], inplist=[], outlist=[],
...
+        states = states or nstatenames
```
bnavigator, line 934: What about if `states == 0` is an empty statenames list a valid parameter for the constructor below?

control/iosys.py
```diff
@@ -928,18 +930,21 @@ def __init__(self, syslist, connections=[], inplist=[], outlist=[],
...
+            states=states, params=params, dt=dt, name=name)
 
         # If input or output list was specified, update it
         nsignals, self.input_index = \
             self._process_signal_list(inputs, prefix='u')
         if nsignals is not None and len(inplist) != nsignals:
             raise ValueError("Wrong number/type of inputs given.")
         nsignals, self.output_index = \
-            self._process_signal_list(outputs, prefix='y')
+            self._process_signal_list(outputs or len(outlist), prefix='y')
```
bnavigator, line 947: What is the purpose of this change and why was it not applied to `inputs` directly above? The function returns `None, {}` for `_process_signal_list(None)`

control/iosys.py
```diff
@@ -918,7 +918,6 @@ def __init__(self, syslist, connections=[], inplist=[], outlist=[],
             # TODO: look for duplicated system names
             self.syslist_index[sys.name] = system_count
             system_count += 1
```
bnavigator, line 920: Sorry, my previous suggestion deleted the whole line, not just the whitespace. Please revert.

```suggestion
            system_count += 1

```

control/iosys.py
```diff
@@ -908,6 +909,7 @@ def __init__(self, syslist, connections=[], inplist=[], outlist=[],
             self.state_offset.append(nstates)
 
             # Keep track of the total number of states, inputs, outputs
+            nstatenames += [sys.name + statename for statename in list(sys.state_index)]
```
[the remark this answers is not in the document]
samlaf: I'm thinking about fixing the <None> name problem at a higher level in the class hierarchy.
I would add a default name "sys[n]" (where n would be an incrementing counter) for every unnamed InputOutputSystem. This would change certain default outputs however, such as line 169:
`str = "System: " + (self.name if self.name else "(None)") + "\n"`
What do you think? Can I go ahead with such changes?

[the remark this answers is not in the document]
murrayrm: Seems fine to assign default names if none is specified.   I suggest using 'x' for state, 'u' for input, 'y' for output, since this is the common terminology used in textbooks.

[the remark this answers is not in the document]
samlaf: This is already what is used. Default names are x[i], u[i], y[i]. So I'm thinking of using sys[i] for default sys name (instead of having None).
When we interconnect systems, I would call default state names sys[i].x[j]
This is because it always annoyed me in my code that when I cascaded a PID with a plant, the state names would revert back to the defaults x[0],x[1],x[2],etc. and I would lose track of which state belongs to which subsystem.
Is this ok?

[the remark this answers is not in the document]
murrayrm: Oops, sorry, I misunderstood.  Seems reasonable.

The system and signal names was done in a hurry.  It would be great to have a better implementation.

Don't forget to add unit tests (including situations that should fail and make sure that they do with the appropriate error message) and also update the documentation.

control/iosys.py
samlaf: I made a comment in the new version I just pushed.
But in short, the block above doesn't make sure there are no duplicates, it only warns the user when there are. Maybe we could extend the warning message to tell the user something like: "If you want unique names, use subsys.copy() before connecting".

control/iosys.py
samlaf: nstatenames was the wrong way to do things. I fixed this differently in the new code, so this is not an issue anymore.

control/iosys.py
samlaf: Also fixed this differently in new code. Thank you for pointing this out.

samlaf:
Here's a short code snippet to visualize the signal names for different Connections:
```
import control

nlios111 = control.NonlinearIOSystem(
    lambda t,x,u,params: x, inputs=2, outputs=2, states=2
)
nlios101 = control.NonlinearIOSystem(
    None, lambda t,x,u,params: u, inputs=2, outputs=2
)

series = nlios111 * nlios101
parallel = nlios111 + nlios101
neg = - nlios111
feedback = nlios101.feedback(nlios111)
dup = nlios111 * nlios111.copy() #if we remove the .copy() we get a warning and names are messed up
hierarchical = series*nlios111

print("1 input, 1 state, 1 output:\n", nlios111, "\n")
print("1 input, no state, 1 output:\n", nlios101, "\n")
print("Series connection: sys[0] * sys[1]\n", series, "\n")
print("Parallel connection: sys[0] + sys[1]\n", parallel, "\n")
print("Negative connection: -sys[0]\n", neg, "\n")
print("Feedback connection: sys[1].feedback(sys[0])\n", feedback, "\n")
print("Connection with duplicate systems: sys[0] * sys[0].copy()\n", dup, "\n")
print("Hierarchical connection: series * sys[0]\n", hierarchical)
```

And the output after my last push:

> 1 input, 1 state, 1 output:
>  System: sys[0]
> Inputs (2): u[0], u[1], 
> Outputs (2): y[0], y[1], 
> States (2): x[0], x[1],  
> 
> 1 input, no state, 1 output:
>  System: sys[1]
> Inputs (2): u[0], u[1], 
> Outputs (2): y[0], y[1], 
> States (0):  
> 
> Series connection: sys[0] * sys[1]
>  System: sys[2]
> Inputs (2): u[0], u[1], 
> Outputs (2): y[0], y[1], 
> States (2): sys[0].x[0], sys[0].x[1],  
> 
> Parallel connection: sys[0] + sys[1]
>  System: sys[3]
> Inputs (2): u[0], u[1], 
> Outputs (2): y[0], y[1], 
> States (2): sys[0].x[0], sys[0].x[1],  
> 
> Negative connection: -sys[0]
>  System: sys[4]
> Inputs (2): u[0], u[1], 
> Outputs (2): y[0], y[1], 
> States (2): sys[0].x[0], sys[0].x[1],  
> 
> Feedback connection: sys[1].feedback(sys[0])
>  System: sys[5]
> Inputs (2): u[0], u[1], 
> Outputs (2): y[0], y[1], 
> States (2): sys[0].x[0], sys[0].x[1],  
> 
> Connection with duplicate systems: sys[0] * sys[0].copy()
>  System: sys[7]
> Inputs (2): u[0], u[1], 
> Outputs (2): y[0], y[1], 
> States (4): sys[6].x[0], sys[6].x[1], sys[0].x[0], sys[0].x[1],  
> 
> Hierarchical connection: series * sys[0]
>  System: sys[8]
> Inputs (2): u[0], u[1], 
> Outputs (2): y[0], y[1], 
> States (4): sys[0].x[0], sys[0].x[1], sys[2].sys[0].x[0], sys[2].sys[0].x[1],

murrayrm:
@samlaf Can you add some unit tests to cover your new code and show that things are working?  The examples that you have in your comment from 20 May would probably be fine.  You might also try mixing named systems and named signals with unnamed systems/signals to make sure that nothing funny happens.  And, as a harder test, try creating a system/signal with a name that would conflict with the default name (eg, create a system whose name is `sys[1]` and combine it with an unnamed systems).

control/iosys.py
```diff
@@ -108,6 +108,14 @@ class for a set of subclasses that are used to implement specific
       The default is to return the entire system state.
 
     """
+
+    idCounter = 0
```
bnavigator, line 112: Make this an instance variable?

control/iosys.py
```diff
@@ -959,10 +956,18 @@ def __init__(self, syslist, connections=[], inplist=[], outlist=[],
...
-            self.output_map[index, ylist_index] = gain
+        for index, outspec in enumerate(outlist):
+            if isinstance(outspec, (int, str, tuple)): outspec = [outspec]
+            for spec in outspec:
+                ylist_index, gain = self._parse_output_spec(spec)
+                self.output_map[index, ylist_index] = gain
+        # self.output_map = np.zeros((self.noutputs, noutputs + ninputs))
+        # for index in range(len(outlist)):
+        #     ylist_index, gain = self._parse_output_spec(outlist[index])
+        #     self.output_map[index, ylist_index] = gain
```
bnavigator, line 970:
```suggestion
```
remove commented out old code

control/iosys.py
```diff
@@ -959,10 +956,18 @@ def __init__(self, syslist, connections=[], inplist=[], outlist=[],
                 self.input_map[self._parse_input_spec(spec), index] = 1
 
         # Convert the output list to a matrix: maps subsystems to system
+        # why can subsystem inputs map to outputs? shouldn't it be
+        # system inputs (hence self.ninputs instead of ninputs)?
         self.output_map = np.zeros((self.noutputs, noutputs + ninputs))
```
bnavigator, line 961:
```suggestion
        self.output_map = np.zeros((self.noutputs, noutputs + ninputs))
```
`(self.noutputs, noutputs + ninputs)` should be correct. See: `set_output_map()`:
https://github.com/python-control/python-control/blob/b68872927b79cf5c10dbf7ae27f93affbefb5ffe/control/iosys.py#L1291-L1315

This also made me realize, that  `inputs` and `outputs` are missing in the `Parameters` section of the docstring for `InterconnectedSystem`.

samlaf:
> @samlaf Can you add some unit tests to cover your new code and show that things are working? The examples that you have in your comment from 20 May would probably be fine. You might also try mixing named systems and named signals with unnamed systems/signals to make sure that nothing funny happens. And, as a harder test, try creating a system/signal with a name that would conflict with the default name (eg, create a system whose name is `sys[1]` and combine it with an unnamed systems).

Should I be enforcing sys/signals names here? Wouldn't this render the API a bit too rigid (in case people want to change the naming conventions later, etc.)
Otherwise, what kind of unit tests did you have in mind?

control/iosys.py
```diff
@@ -925,23 +911,34 @@ def __init__(self, syslist, connections=[], inplist=[], outlist=[],
...
             sysname_list.append(sys.name)
 
+        # This won't work when there are duplicate sysnames
+        # Should we throw an error rather than just warning above?
+        # Users can use sys.copy() to get a unique name for the dup system.
+        if states is None:
+            states = []
+            for sys in sysobj_list:
+                states += [sys.name + '.' + statename for statename in sys.state_index.keys()]
+
```
samlaf, line 925: @murrayrm This is the part that would break if you add a named signal with a generic 'sys[i]' that will cause a conflict (it's no different from having two systems with the same name). The only "bad" thing that happens is that the system will be missing the state names of the systems with conflicting names (only the first 'sys[i].states' will be present).
Right now we are only raising a warning. As I mention in the comment, users can use sys.copy() first to create a unique name. Is this ok?

control/iosys.py
samlaf: This is the class variable used to count how many object instances were created so far (new instances without names are called sys[idCounter]). Is there a naming convention I am not following (eg. should class variables be prefixed by an underscore (_idCounter))?

bnavigator: Ok, my bad. Sorry.

control/iosys.py
```diff
@@ -108,6 +108,14 @@ class for a set of subclasses that are used to implement specific
       The default is to return the entire system state.
 
     """
+
+    idCounter = 0
+    def name_or_default(self, name=None):
+        if name is None:
+            name = "sys[{}]".format(InputOutputSystem.idCounter)
+            InputOutputSystem.idCounter += 1
```
bnavigator, line 116: Re: conflicting names. Maybe register all system names module or class-wide and check here before assigning. If the name is already taken, just increment the counter once more.

samlaf: I tried implementing this with a __del__ method that would remove an object's name from the class-wide register after this object gets deleted or runs out of scope (to prevent annoying errors being raised when trying to create an object with the same name but in a different scope)

It seems that python's __del__ method isn't really a destructor, and all sorts of bugs happen because of python's reference-counting garbage collector. (https://eli.thegreenplace.net/2009/06/12/safely-using-destructors-in-python)

One potential solution they say is to have users create objects using (https://stackoverflow.com/questions/865115/how-do-i-correctly-clean-up-a-python-object)
`with InputOutputSystem() as sys:
   ...`
But this seems utterly annoying.

bnavigator: Or just let the user take care of it. If the user want's to create 100 systems with the same name, fine. Just either throw an error at the time of combining two systems with the same name, or, if you want to be more graceful, rename them into `samename_1` and  `samename_2`.

samlaf: Just made these changes. Also added unit tests that make sure the number of states/inputs/outputs is correct (which requires having renamed all duplicate system names).

control/tests/iosys_test.py
```diff
@@ -802,10 +802,64 @@ def test_named_signals(self):
         )
         ss_feedback = ct.feedback(self.mimo_linsys1, self.mimo_linsys2)
         lin_feedback = ct.linearize(ios_connect, 0, 0)
-        np.testing.assert_array_almost_equal(ss_feedback.A, lin_feedback.A)
-        np.testing.assert_array_almost_equal(ss_feedback.B, lin_feedback.B)
-        np.testing.assert_array_almost_equal(ss_feedback.C, lin_feedback.C)
-        np.testing.assert_array_almost_equal(ss_feedback.D, lin_feedback.D)
+        for M, N in ((ss_feedback.A, lin_feedback.A), (ss_feedback.B, lin_feedback.B),
+                     (ss_feedback.C, lin_feedback.C), (ss_feedback.D, lin_feedback.D)):
+            np.testing.assert_array_almost_equal(M, N)
```
bnavigator, line 807: Please avoid loops like this. It makes debugging failing tests in pytest more difficult, because the assertion error message now reads "M is not equal N" instead of directly showing which one failed.

control/tests/iosys_test.py
```diff
@@ -802,10 +802,64 @@ def test_named_signals(self):
...
+        
+        nlios222 = ct.NonlinearIOSystem(
+            lambda t,x,u,params: x, inputs=2, outputs=2, states=2
+        )
+        nlios202 = ct.NonlinearIOSystem(
+            None, lambda t,x,u,params: u, inputs=2, outputs=2
+        )
+        check_name_and_ios_counts(nlios202)
+
+        # Unnamed/unnamed connections
```
bnavigator, line 828: Could you please split up the checks into separate tests? They could use the same data by defining the system `nlios202 `and `nlios222` class wide (or through a pytest fixture).

control/tests/iosys_test.py
```diff
@@ -802,10 +802,64 @@ def test_named_signals(self):
...
+                     (ss_feedback.C, lin_feedback.C), (ss_feedback.D, lin_feedback.D)):
+            np.testing.assert_array_almost_equal(M, N)
+
+        # We also enforce generic names to be present when systems are created
+        # without names (sys, states, inputs and outputs)
+        def check_name_and_ios_counts(sys):
+            self.assertIsNotNone(sys.name)
+            self.assertEqual(len(sys.state_index), sys.nstates)
+            self.assertEqual(len(sys.input_index), sys.ninputs)
+            self.assertEqual(len(sys.output_index), sys.noutputs)
```
bnavigator, line 815: Use docstring for function.

This only checks for the correct count.

It would be worth to also check that desired names are really set or have the correct fallback name (if you choose to implement my suggestion in https://github.com/python-control/python-control/pull/400#discussion_r455335103)

control/iosys.py
```diff
@@ -540,10 +541,10 @@ def linearize(self, x0, u0, t=0, params={}, eps=1e-6):
         linsys = StateSpace(A, B, C, D, self.dt, remove_useless=False)
         return LinearIOSystem(linsys)
 
-    def copy(self):
+    def copy(self, newname=None):
         """Make a copy of an input/output system."""
         newsys = copy.copy(self)
-        newsys.name = self.name_or_default(None)
+        newsys.name = self.name_or_default(newname)
```
bnavigator, line 547: How about

```suggestion
        newsys.name = self.name_or_default("copy of " + self.name if not newname)
```

control/iosys.py
```diff
@@ -918,30 +925,46 @@ def __init__(self, syslist, connections=[], inplist=[], outlist=[],
...
+                sysname = sys.name + "_" + str(count)
+                sysobj_name_dct[sys] = sysname
+                warn("Duplicate name found in system list. Renamed to {}".format(sysname))
+            else:
+                sysname_count_dct[sys.name] = 1
+                sysobj_name_dct[sys] = sys.name
+                
+        # This won't work when there are duplicate sysnames
+        # Should we throw an error rather than just warning above?
+        # Users can use sys.copy() to get a unique name for the dup system.
```
bnavigator, line 946: Now obsolete, right? Maybe document the sys.copy part it in the function docstring?

murrayrm:
Unit tests are failing because
```
nlsys * nlsys
```
no longer works.  It seems like it should be possible to take two copies of a system and put them in series (which is what this does) without generating an error.

control/tests/iosys_test.py
samlaf: Yes, totally. I'll revert back in the next commit. The three previous checks in this test were using this "for M, N" structure, so I figured this one was an anomaly and changed it. I will remove all for loops then.

control/iosys.py
samlaf: Done. sys.copy() is no longer necessary, so no need to add to docstring.

control/tests/iosys_test.py
samlaf: Done for system names. Still need to do signal names.

samlaf:
Thank you for pointing this out @murrayrm, and thanks a lot to both of you two for your patience... newbie at work here.
I was concentrating on the signal_names test for too long and forget to rerun the other tests afterwards. Should be fixed now.

The system naming convention code should be fine now. Still need to write the signal name convention unit test though.

murrayrm:
Looks like git got confused by the new unit test.  I don't see any actual conflicts, but you may need to fiddle with things to allow the automatic merge check to work.

samlaf:
It seems like a conflict between my branch and master branch?

murrayrm:
That's what it's saying, but what's odd is that I don't see the conflict.  You added in a test and it shouldn't conflict with other tests.  If you agree, I can try to fix it via the web interface and we'll see if that gets things running again.

samlaf:
Sure, please do. I wouldn't know what to try anyways and would be scared to break other things.

samlaf:
I added the last unit tests. This is ready to be reviewed and hopefully should be close to done.

bnavigator:
The changes remove test coverage of the functions `set_input_map()` and `set_output_map()`
https://coveralls.io/builds/32204258/source?filename=control%2Fiosys.py#L1284

In fact, those functions are not used by the whole package anymore. But they are public API. So a unit test for them should be present. (I would personally prefer that they are created in the style of #438. Feel free to add them as a PR against the `bnavigator:array-matrix-tests` branch)

samlaf:
Sure I can do that.
Just to make sure I understand, I think these functions were used in the __add__, __mul__, etc. functions which interconnect systems. Before they would create a blank interconnected system and then use the setters like set_input_map to populate the object. I changed it so that the "setting" was done in the init code.
Anyways, my question is: this code being "covered" before means that the unit test was calling them? So even if there wasn't a designed unit test for these specific functions, at least if they had caused a bug that would have caused some unit test to fail, but now that's not the case anymore, so we should add an explicit unit test. Is this correct?

bnavigator:
Yes, that is correct. The unit tests should make sure that using those functions results in the documented behavior. Previously, this was more or less implicit by the tests when `__add__` , `__mul__`. `__neg__`  and `feedback` were called.

bnavigator:
Hi @samlaf, if you still intend to update your PR with the requested  unit tests, please do so based on the iosys_test.py in #438. We are discussing in #451 about the proper merge sequence, and #438 will likely be the first.

samlaf:
Hi @bnavigator, I was planning on doing these, but I got major neck pain recently, and am spending some time away from the keyboard. Are you waiting on me to merge the other PRs? I'm not sure when I'll have time to write these, but if it's not a hurry I will eventually get to them.

bnavigator:
No hurry, get well soon! As I wrote, I removed the dependency in my PR so your PR can be merged later. Whenever you can get to it. You will have to rebase.

bnavigator (approved):

PR #403, opened 2020-05-14
small corrections for continuous-to-discrete approximations

sawyerbfuller:
* combined dtime/sample_system and dtime/c2d into the same function because they do the same thing. old c2d also had an unnecessary test
* fixed TransferFunction.sample and StateSpace.sample to have updated docstring that adds recent methods added to scipy.signal.cont2discrete (except for 'matched', which is only available for TransferFunction and only for python-control)

### DIFF
--- a/control/dtime.py
+++ b/control/dtime.py
@@ -49,7 +49,7 @@ $Id: dtime.py 185 2012-08-30 05:44:32Z murrayrm $
 from .lti import isctime
 from .statesp import StateSpace, _convertToStateSpace
 
-__all__ = ['sample_system', 'c2d']
+__all__ = ['sample_system']
 
 # Sample a continuous time system
 def sample_system(sysc, Ts, method='zoh', alpha=None):
@@ -65,7 +65,7 @@ def sample_system(sysc, Ts, method='zoh', alpha=None):
     Ts : real
         Sampling period
     method : string
-        Method to use for conversion, e.g. 'matched', 'zoh' (default)
+        Method to use for conversion, e.g. 'zoh' (default)
 
     Returns
     -------
@@ -80,7 +80,7 @@ def sample_system(sysc, Ts, method='zoh', alpha=None):
     Examples
     --------
     >>> sysc = TransferFunction([1], [1, 2, 1])
-    >>> sysd = sample_system(sysc, 1, method='matched')
+    >>> sysd = sample_system(sysc, 1, method='zoh')
     """
 
     # Make sure we have a continuous time system
@@ -88,5 +88,3 @@ def sample_system(sysc, Ts, method='zoh', alpha=None):
         raise ValueError("First argument must be continuous time system")
 
     return sysc.sample(Ts, method, alpha)
-
-c2d = sample_system 
--- a/control/dtime.py
+++ b/control/dtime.py
@@ -52,7 +52,7 @@ from .statesp import StateSpace, _convertToStateSpace
 __all__ = ['sample_system', 'c2d']
 
 # Sample a continuous time system
-def sample_system(sysc, dt, method='zoh', alpha=None):
+def sample_system(sysc, Ts, method='zoh', alpha=None):
     """Convert a continuous time system to discrete time
 
     Creates a discrete time system from a continuous time system by
@@ -62,10 +62,10 @@ def sample_system(sysc, dt, method='zoh', alpha=None):
     ----------
     sysc : linsys
         Continuous time system to be converted
-    dt : real
+    Ts : real
         Sampling period
     method : string
-        Method to use for conversion, e.g. 'matched', 'tustin', 'zoh' (default)
+        Method to use for conversion, e.g. 'matched', 'zoh' (default)
 
     Returns
     -------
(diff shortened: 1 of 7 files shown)


control/dtime.py
```diff
@@ -89,32 +89,4 @@ def sample_system(sysc, Ts, method='zoh', alpha=None):
 
     return sysc.sample(Ts, method, alpha)
 
-
-def c2d(sysc, Ts, method='zoh'):
```
bnavigator, line 93: This removes the documentation for `c2d` doesn't it? Not what we want, probably. This is clearly written as a wrapper.

sawyerbfuller:
Hm, looks like my previous (and first) pull request adding dlqe is still in here. don't yet know how to take that out of this pull request, let me know if it's easier if I somehow "merge" branches first and then resubmit this PR

murrayrm:
@sawyerbfuller If you want to get a "clean" PR, I think you need to create a new branch on your account and rebase off of the current master.  But likely we will incorporate this PR after you other one, so should be fine to just build on this one.

bnavigator:
As there is no other open PR from your master branch, you can rebase your master onto the upstream master.

1. If you have not added the remote upstream yet:
`git remote add upstream https://github.com/python-control/python-control`
2. Fetch all
`git fetch --all`
3. Rebase your last 3 commits after e3125696a onto upstream/master
`git rebase --onto upstream/master e312569 master`


Next time first create a separate branch and submit the PR from that.

control/dtime.py
sawyerbfuller: Hi Ben, this PR makes it so both c2d and sample_system are identical (one is equal to the other) so that they have the same documentation. Which is a shortened version of eg. Transferfunction.sample, which is itself a modified version of scipy.signal.cont2discrete's docs. For both cases, do you think it would make sense to just refer the reader to  scipy.signal.cont2discrete, to avoid the need to continually update documentation?

bnavigator: I know, but are you sure that your change retains the documentation for c2d in https://python-control.readthedocs.io/en/0.8.3/generated/control.matlab.c2d.html? The call signature will change for sure, because `c2d()` did not have the `alpha` parameter before.

sawyerbfuller: Hi Ben, do you think it's ok if c2d gets an alpha parameter now (that matlab doesn't have?) 

Either way, I can update the code to duplicate its docstring so that the docstring example is for c2d rather than sample_system.

murrayrm:
Since this is really a "convenience" function for MATLAB users, it might be worth considering:

* Moving `c2d` into the `matlab` module.

* To avoid confusing, adding a `c2d` specific docstring that matches the typical usage and then pointing people to `sample_system` for other options.

The `c2d` function could just be:
```
def c2d(...)
   """Docstring...."""
   return sample_system(...)
```

sawyerbfuller:
OK moved c2d to matlab module. Now it's only available if you import control.matlab (unlike many other matlab-like commands like tf, bode, ... ). Not sure what the philosophy should be here, but let me knw if you think it should still be available from the control module.

murrayrm:
Perhaps we leave a `c2d` in the main `python-control` package (in addition to in the `matlab` module) but with a `DeprecationPending` warning?  There's an example of this in `FRD.evalfr`.  We just have to be careful to make sure we don't get a namespace conflict if someone does `from control import *` or something like that.

bnavigator:
> Since this is really a "convenience" function for MATLAB users, it might be worth considering:
> 
> * Moving `c2d` into the `matlab` module.

Moving the function without referencing it in the original module implies you have to change some tests too and it will probably break a lot of old code, as evident from the most recent CI failure:

```
______________________ ERROR collecting tests/lti_test.py ______________________
754ImportError while importing test module '/home/travis/build/python-control/python-control/control/tests/lti_test.py'.
755Hint: make sure your test modules/packages have valid Python names.
756Traceback:
757control/tests/lti_test.py:7: in <module>
758    from control import c2d
759E   ImportError: cannot import name 'c2d' from 'control' (/home/travis/build/python-control/python-control/control/__init__.py)
```


----------------------------------

> The `c2d` function could just be:
> 
> ```
> def c2d(...)
>    """Docstring...."""
>    return sample_system(...)
> ```

That's basically the current situation, bar some probably unnecessary output conversion.

----------------------------------

> @sawyerbfuller If you want to get a "clean" PR, I think you need to create a new branch on your account and rebase off of the current master. But likely we will incorporate this PR after you other one, so should be fine to just build on this one.

The LQE code has been merged in December 2019 already (72262c65749d1a15e58680f3ba6efc4842a0448f). The commit of the CI test merge 3ba8b7d11b8c80ca3f33bb263723d86223c5007c looks clean. I would rebase it as descibed anyway to have a nice diff visible here in the PR, which will also keep the history log look cleaner when finally merging into python-control/master

murrayrm:
I fixed things up and put them in PR #404.

sawyerbfuller:
Thanks! I turn into a pumpkin over the weekend with the kids at home!

PR #404, opened 2020-05-17
small corrections for continuous-to-discrete approximations

murrayrm:
This PR replaces #403.

From sawyerbfuller:

* combined dtime/sample_system and dtime/c2d into the same function because they do the same thing. old c2d also had an unnecessary test

* fixed TransferFunction.sample and StateSpace.sample to have updated docstring that adds recent methods added to scipy.signal.cont2discrete (except for 'matched', which is only available for TransferFunction and only for python-control)

Additional modifications (murrayrm):

* Added `c2d` back into `dtime.py` with a `PendingDeprecationWarning` to fix unit tests.

* Fixed up a few PEP8 problem along the way

### DIFF
--- a/control/dtime.py
+++ b/control/dtime.py
@@ -49,7 +49,7 @@ $Id: dtime.py 185 2012-08-30 05:44:32Z murrayrm $
 from .lti import isctime
 from .statesp import StateSpace, _convertToStateSpace
 
-__all__ = ['sample_system', 'c2d']
+__all__ = ['sample_system']
 
 # Sample a continuous time system
 def sample_system(sysc, Ts, method='zoh', alpha=None):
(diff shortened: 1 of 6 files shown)


control/dtime.py
```diff
@@ -88,3 +90,19 @@ def sample_system(sysc, Ts, method='zoh', alpha=None):
...
+def c2d(sysc, Ts, method='zoh'):
+    """(Deprecated) Return a discrete-time system; use sample_system instead"""
+    warn("The function `c2d` will be deprecated in a future release.  Use"
+         "`sample_system` (or `matlab.c2d`) instead.",
+         PendingDeprecationWarning)
+
+    #  Call the sample_system() function to do the work
+    sysd = sample_system(sysc, Ts, method)
+
+    # TODO: is this check needed?  If sysc is  StateSpace, sysd is too?
```
sawyerbfuller: I think this check isn't needed anymore. 

sample_system calls the system's 'sample' function, and StateSpace.sample always returns a stateSpace system.

PR #417, opened 2020-06-11
added ability to 'prewarp' the conversion of continuous to discrete-time systems

sawyerbfuller:
(in functions sample, sample_system, and c2d) , so that their gain and phase match at a specific desired frequency.

### DIFF
--- a/control/statesp.py
+++ b/control/statesp.py
@@ -841,7 +841,7 @@ but B has %i row(s)\n(output(s))." % (self.inputs, other.outputs))
         j = indices[1]
         return StateSpace(self.A, self.B[:, j], self.C[i, :], self.D[i, j], self.dt)
 
-    def sample(self, Ts, method='zoh', alpha=None):
+    def sample(self, Ts, method='zoh', alpha=None, prewarp_frequency=None):
         """Convert a continuous time system to discrete time
 
         Creates a discrete-time system from a continuous-time system by
@@ -866,6 +866,10 @@ but B has %i row(s)\n(output(s))." % (self.inputs, other.outputs))
             should only be specified with method="gbt", and is ignored
             otherwise
 
+        prewarp_frequency : float within [0, infinity)
+            The frequency [rad/s] at which to match with the input continuous-
+            time system's magnitude and phase
+
         Returns
         -------
         sysd : StateSpace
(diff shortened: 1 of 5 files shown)


control/statesp.py
```diff
@@ -885,8 +889,12 @@ def sample(self, Ts, method='zoh', alpha=None):
...
 
         sys = (self.A, self.B, self.C, self.D)
-        Ad, Bd, C, D, dt = cont2discrete(sys, Ts, method, alpha)
-        return StateSpace(Ad, Bd, C, D, dt)
+        if prewarp_frequency is not None:
+            Twarp = 2*np.tan(prewarp_frequency*Ts/2)/prewarp_frequency
+            Ad, Bd, C, D, _ = cont2discrete(sys, Twarp, method, alpha)
+        else:
+            Ad, Bd, C, D, _ = cont2discrete(sys, Ts, method, alpha)
+        return StateSpace(Ad, Bd, C, D, Ts)
```
murrayrm, line 897: This could be shortened to
```
Twarp = Ts if prewarp_frequency is None else 2*np.tan(prewarp_frequency*Ts/2)/prewarp_frequency
Ad, Bd, C, D, _ = cont2discrete(sys, Twarp, method, alpha)
```

control/xferfcn.py
```diff
@@ -1010,8 +1014,12 @@ def sample(self, Ts, method='zoh', alpha=None):
...
             return _c2d_matched(self, Ts)
         sys = (self.num[0][0], self.den[0][0])
-        numd, dend, dt = cont2discrete(sys, Ts, method, alpha)
-        return TransferFunction(numd[0, :], dend, dt)
+        if prewarp_frequency is not None:
+            Twarp = 2*np.tan(prewarp_frequency*Ts/2)/prewarp_frequency
+            numd, dend, _ = cont2discrete(sys, Twarp, method, alpha)
+        else:
+            numd, dend, _ = cont2discrete(sys, Ts, method, alpha)
+        return TransferFunction(numd[0, :], dend, Ts)
```
murrayrm, line 1022: This could be shortened to
```
Twarp = Ts if prewarp_frequency is None else 2*np.tan(prewarp_frequency*Ts/2)/prewarp_frequency
numd, dend, _ = cont2discrete(sys, Twarp, method, alpha)
```

control/statesp.py
sawyerbfuller: This might work except that we need to keep the original Ts because it gets used a few lines later to construct the discrete time system that the function returns.

sawyerbfuller: Never mind, I think you're right.

sawyerbfuller: fixed a failing test and realized I should add an additional condition that prewarping will only apply to bilinear transformation.

sawyerbfuller:
Remark: this is used in discrete control design when your controller is first designed in continuous time and then Discretized. discretization typically changes the mag and phase; ‘prewarping’ allows  you to keep the same mag and phase of your cont-time controller in discrete time at one specific frequency. Usually that is the gain=1 crossover frequency.

Issue #287, opened 2019-04-10
Default simulation times for discrete-time step, impulse and initial condition response is wrong


murrayrm:
The current method for computing the default simulation time for discrete-time simulations uses a call to `scipy.signal._default_response_times()`, but this uses an eigenvalue calculation that assumes continuous time.  For discrete time systems, need to figure out what the right default values are for the length of a simulation for impulse, step, and initial condition responses.  See also discussion in issue #239.

ilayn:
Not just discrete time systems but also to get meaningful arrays, continuous time systems also need to be checked a bit more thoroughly. Main issue is that by just getting the pole structure is not sufficient to get the acceptable dt and tfinal. My goto examples 

```python
import control as cnt
import matplotlib.pyplot as plt
plt.plot(*cnt.step_response(cnt.tf(1, [1, 1001, 1000])))
plt.plot(*cnt.step_response(cnt.tf([1, 2, 0], [1, 1001, 1000])))
```

These give 
![image](https://user-images.githubusercontent.com/1303842/55857388-480ec080-5b6d-11e9-9b75-4dd0e1e1d4b9.png)

and

![image](https://user-images.githubusercontent.com/1303842/55857392-4e9d3800-5b6d-11e9-9109-5785279d6a33.png)

Here the issue is that only the slowest stable root is being evaluated for the tfinal and the sampling rate. We can also make the same argument for oscillating modes that certain combinations would lead to ugly results for example `plt.plot(*cnt.step_response(cnt.tf([1, -200, -500], [1, 2, 1000])))` gives 

![image](https://user-images.githubusercontent.com/1303842/55857707-219d5500-5b6e-11e9-959a-363396c05d2e.png)

I've spent quite some time on this and definitely I'm quite not finished but here is the main argument in the docstring https://github.com/ilayn/harold/blob/master/harold/_time_domain.py#L280-L454

In summary, only the modes that contribute to the response should be taken into account. I still don't know how matlab does it but for example, I couldn't figure out how to check the repeated (Jordan blocks) contribute.

murrayrm:
Any estimate for when it might be done?  It would be great when you get things working if you could contribute it to the python-control library!

ilayn:
Well I am kind of stuck at this point. It works relatively well but there are a few corner cases that I don't know how to proceed. I am wondering whether pushing it to SciPy is a better solution since then python-control codebase doesn't need any change.

murrayrm:
[...]

PR #420, opened 2020-06-18
Improved default time vector and handling for time response functions step, impulse, and initial 

sawyerbfuller:
step, impulse, and initial, and step_info, as well as their matlab equivalents

* before, it always returned a time vector with unit time intervals, rather than sys.dt time intervals. 
* introduced a new utility function <s>`_get_default_tfinal(sys)`</s> `_get_ideal_tfinal_and_dt(sys)` that computes a time window equal to seven times the slowest time constant of sys, inspired by scipy.signal.ltisys._default_response_times(A, n), that can also accommodate discrete-time systems


murrayrm:
Looks like this broke some unit tests?

sawyerbfuller:
Ok take a look, fixed the tests. Also implemented a convenience that you can now specify T as a scalar in step, impulse, and initial to quickly change the length of the simulation. Easier than specifying the whole T vector every time (which you can still do). This is similar to how you can do it in bode, and how Matlab does it. I think these conveniences will not break compatibility with extant code.

sawyerbfuller:
I’m not sure how solidified things are considered to be in the package, so I’m making guesses, trying balance adding functionality I would like to have, and avoiding breaking backwards compatibility. Please let me know if I am going too far in one direction or the other!  

If these changes seem ok, I can add a couple more unit tests.

sawyerbfuller:
Looks like this PR addresses half of the issue of issue #287 , but I essentially just used the same coarse algorithm to estimate a reasonable time horizon that scipy uses. And expanded it to work with Discrete-time systems.

murrayrm:
This looks good to me.  I need to check it a bit more carefully, but nice to have a proper method for computing response times and some enhancements on how to specify the final time.

sawyerbfuller:
Great. following issue #287, I took a stab at automatically generating dt as well if not specified. And added unit tests.

sawyerbfuller:
I hope that for now this is maybe a “good enough” version of the more comprehensive solution that  ilayn was working on in #287 – that also adds discrete-time support.

control/tests/sisotool_test.py
```diff
@@ -27,7 +27,7 @@ def test_sisotool(self):
         assert_array_almost_equal(ax_rlocus.lines[2].get_data(),initial_point_2)
 
         # Check the step response before moving the point
-        step_response_original = np.array([ 0.,          0.02233651,  0.13118374,  0.33078542,  0.5907113,   0.87041549, 1.13038536,  1.33851053,  1.47374666,  1.52757114])
+        step_response_original = np.array([0.    , 0.0217, 0.1281, 0.3237, 0.5797, 0.8566, 1.116 , 1.3261, 1.4659, 1.526 ])
```
murrayrm, line 30: Why do these values need to be changed?  The system is the same and so shouldn't the step response be the same?

control/tests/sisotool_test.py
```diff
@@ -59,7 +59,7 @@ def test_sisotool(self):
         assert_array_almost_equal(ax_mag.lines[0].get_data()[1][10:20],bode_mag_moved)
 
         # Check if the step response has changed
-        step_response_moved = np.array([[ 0.,          0.02458187,  0.16529784 , 0.46602716 , 0.91012035 , 1.43364313, 1.93996334 , 2.3190105  , 2.47041552 , 2.32724853] ])
+        step_response_moved = np.array([0.    , 0.0239, 0.161 , 0.4547, 0.8903, 1.407 , 1.9121, 2.2989, 2.4686, 2.353 ])
```
murrayrm, line 62: Same question as above.  Just want to make sure I understand the need for the change.

control/timeresp.py
```diff
@@ -458,8 +462,9 @@ def step_response(sys, T=None, X0=0., input=None, output=None,
     sys: StateSpace, or TransferFunction
         LTI system to simulate
 
-    T: array-like object, optional
-        Time vector (argument is autocomputed if not given)
+    T: array-like or number, optional
+        Time vector, or simulation time duration if a number (time vector is 
+        autocomputed if not given)
```
murrayrm, line 467: Should we document someplace (perhaps in `doc/`) how this is autocomputed?

control/timeresp.py
```diff
@@ -472,6 +477,10 @@ def step_response(sys, T=None, X0=0., input=None, output=None,
     output: int
         Index of the output that will be used in this simulation. Set to None
         to not trim outputs
+    
+    T_num: number, optional
+        Number of time steps to use in simulation if T is not provided as an 
+        array (autocomputed if not given); ignored if sys is discrete-time.
```
murrayrm, line 483: Document someplace how these are autocomputed?

control/tests/sisotool_test.py
sawyerbfuller: The function in scipy that used to be used to generate the time vector, `scipy.signal.ltisys._default_response_times(A, n)` uses `numpy.linspace` without the keyword arg `endpoint=False` so it returns n+1 data points instead of n. The new function does, so it generates only n data points, which is consistent with how other functions in python-control seem to do it. (thus the slightly changed values). Let me know if you have a preference to go back to n+1 datapoints. My thinking is, if you ask for n datapoints, that ought to be how many you get.

control/timeresp.py
sawyerbfuller: I couldn't find an obvious place in doc about where to put it so I put in the docstring of step_response and referred back to that docstring in impuse, initial, and step_info.

Issue #422, opened 2020-06-22
use dt=0 instead of dt=None (unspecified timebase) as default option for lti systems?


sawyerbfuller:
I was wondering if somebody could tell me the motivation behind the current convention for default initialization of LTI systems to have dt=None? Using dt=None means an unspecified time base, allowing the system to be interpreted either as cont or discrete time. (I couldn't find any discussion of this but please let me know if I missed it). 

A system with an unspecified dt seems to be an unusual object:

- printing out its transfer function shows 's' as the complex variable, suggesting it is a cont-time system
- but isdtime(sys) evaluates to true (but not if strict=True is specified) -- I thought this was a bug at first
- cont and discrete time systems have substantially different behavior, which seems to be at odds with the notion that an object should know how to handle itself

Have you thought about using a default of dt=0 (continuous time), and leave the dt=None case be a special case, to be instantiated only when needed?  (Issue #419 made me think of this and is related)

ilayn:
It's a relic from Matlab with some very old historic reasons but survived under backwards-compatibility flow. Then apparently adopted also here. Internally matlab secretly inserts some default value (that I forgot) whenever dt is needed anyways. My take on this is that user should insert some random value if they don't care because they don't care or dt=None should map to dt=0.01 to be slightly realistic.

murrayrm:
I think I was the person who used dt=None.  The motivation was a bit different: if someone didn't specify whether the system was continuous or discrete, we wanted to allow as much flexibility as possible.  So, for example, if you combine a system without a timebase (dt=None) with one that has a time base (dt=0, dt=True, dt=0.1, etc) then the first system will assume the time base of the other.

This probably doesn't matter that much, though it is handy for systems that are constants (no states).

The current approach is documented [here](https://python-control.readthedocs.io/en/0.8.3/conventions.html#discrete-time-systems).

I would like to see a bit more discussion of this before we make a change, since it may break people's code.

sawyerbfuller:
Hi Richard, makes sense that dt=None that could be very useful for systems that are a static gain. 

Could dt=None be a special case that is used by default when constructing such systems? 

Looks like in matlab that the default is dt=0 (continuous time). with an option to set it to -1 to be unspecified if desired: https://www.mathworks.com/help/control/ref/tf.html#f4-209775_sep_mw_ec86807c-5450-4be4-ae9a-e85042c27676

I couldn’t find any examples of systems using an unspecified time base online after a quick search.

[...]

Issue #244, opened 2018-11-14
StateSpace constructor should not call _remove_useless_states


MaxGaukler:
State space representations should, in my opinion, not be reduced automatically with StateSpace._remove_useless_states(), similar to the fact that we don't call minreal() automatically.

Unlike transfer functions, state space models can model non-observable or non-controllable behaviour, and cannot always be considered equivalent to their minimized form. Additionally, it is rather confusing to have `ss(A,B,C,D).A !== A`  and e.g. `ss(A,B,C,D).sample(T).A !== expm(A*T)`. For a more object-oriented interface, it would also be great to have member functions like stateSpace.isControllable(), which do not make sense if the system has already been minimized.

To give an example: The autonomous system x'=0, y=x  is a signal model for the constant output y=x(0). It is not equivalent to the system y=0 if you consider the initial response for x(0) !== 0.

```
​import control
print(control.__version__)
print(control.ss(0,0,1,0)) # expected: A,B,C,D unchanged from input
print(control.ss(0,0,1,0).sample(1)) # expected: ss(1,0,1,0, dt=1)
```
output:
```
0.8.0

A = []
B = []
C = []
D = [[0]]

A = []
B = []
C = []
D = [[0]]
dt = 1
```

Is there a reason for the current behavior, or would it be okay to change it?

murrayrm:
I agree with this comment.

One thing to think through is whether anything "breaks" if we stop calling `_remove_useless_states()`.  Eg, are there operations that create spurious states that were getting reduced out?

Perhaps create a branch, try removing that call to `_remove_useless_states()` and see if unit tests pass?

MaxGaukler:
Lots of tests break if the call is removed, mainly blockdiagram algebra (LTI system interconnection) and some others which may be caused by that (rlocus, LQR). It seems that some more effort is required.
https://github.com/MaxGaukler/python-control/tree/remove-useless-states
https://travis-ci.org/MaxGaukler/python-control/jobs/457349504#L4054

DapRock:
I'm having the same problem with the system order reduction. I've tried to use the library to work with a Luenberger observer which assumes a state as invariant in time and I'm getting this same state being reduced in the problem. I have better description here: https://stackoverflow.com/questions/57484563/why-some-a-state-is-removed-from-the-model-while-using-control-ss-in-python

Do you have any tip to workaround in this case? Thanks for your time!

murrayrm:
[...]

PR #424, opened 2020-06-26
add capability to switch to legacy defaults

sawyerbfuller:
For example, `config.use_legacy_defaults('0.8.3')`  
* as suggested by murrayrm in #422 

I added a few more defaults that might change in the future, but left them as they currently are for this PR: 
	* statesp.remove_useless_states (issue #244) (currently True; future is False? )
	* statesp and xferfcn.default_dt (issue #422) (currently None, future is 0?)
	* statesp.use_numpy_matrix (currently True, future is False)

### DIFF
--- a/control/config.py
+++ b/control/config.py
@@ -161,6 +161,13 @@ def use_numpy_matrix(flag=True, warn=True):
     set_defaults('statesp', use_numpy_matrix=flag)
 
 def use_legacy_defaults(version):
+    """ Sets the defaults to whatever they were in a given release.
+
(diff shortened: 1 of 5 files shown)


murrayrm:
Changes look good.  I think the docstrings for the various functions need to be updated to indicate the default values come from config variables.  We should also add the new variables to the documentation in `doc/`.

murrayrm:
Looks ready to merge.  Will let sit for a few days in case others want to comment.

Issue #432, opened 2020-07-18
Interpretation of argument for use_legacy_defaults is not clear


murrayrm:
My sense of the way that `use_legacy_defaults` should work is that if I had code that was working in some version and it stops working in a future version, I should be able to use `use_legacy_defaults` to return things to the version that was working.

So, for example, if I had things working in 0.8.2 and I upgrade to 0.8.4 or 0.9.0 and things stop working, I should be able to say
```
use_legacy_defaults('0.8.2')
```
and my code will work again.

This particular call definitely won't work (`use_legacy_defaults` only recognizes '0.8.3') and it is not clear to me how it could work.  Basically, it seems like we need to track when a change in default behavior occurred and then "revert" any defaults that are different between the version I was using and the current version.

@sawyerbfuller Does my issue make sense?  This is related to PR #431, I think.

sawyerbfuller:
I don’t have much familiarity with backwards-incompatible changes before 0.8.3 but I assumed that we could use this function to add such functionality as needed from feedback from users. 

My take on it is that the function should be available as a “good enough” solution that allows users to avoid the thorniest issues that come up whenever there are big changes in the code (a few of which I anticipated in that PR). But not something that will include every version number and provide compatibility with minor revision of how the code works. 

In short, we should be a little bit “opinionated” about how things ought to work rather than to aim to maintain perfect backwards compatibility at all times. 

So, if I update to the latest version of python-control and something breaks in my code, the first thing I should try is use-legacy-defaults on various previous version numbers, one of which might fix things, but maybe not everything.

sawyerbfuller:
We could alternatively include every recent release in that function, most of which would do the same thing.

murrayrm:
I agree with the philosophy, especially the summary sentence:

> So, if I update to the latest version of python-control and something breaks in my code, the first thing I should try is use-legacy-defaults on various previous version numbers, one of which might fix things, but maybe not everything.

But this means that `use_legacy_defaults` needs to take an arbitrary version number and do the "right thing", which I don't think it currently does.  Eg, if I were using 0.8.2 and things were working, then I got to 0.8.4 and things break, in the current implementation I would need to know to call `use_legacy_defaults('0.8.3')` even though I never personally used 0.8.3.

I think a bit of clever parsing of the version number and lexicographical comparison would allow an implementation where if I enter a version number I get the legacy defaults for anything that changed after that version number.

PR #435, opened 2020-07-23
quick and dirty parser for version number in legacy-defaults

sawyerbfuller:
follows issue #432

### DIFF
--- a/control/config.py
+++ b/control/config.py
@@ -168,8 +168,9 @@ def use_legacy_defaults(version):
     version : string
         version number of the defaults desired. ranges from '0.1' to '0.8.4'. 
     """
-    first_digit = int(version[0])
-    second_digit = int(version[2])
+    numbers_list = version.split(".")
+    first_digit = int(numbers_list[0])
+    second_digit = int(numbers_list[1].strip('abcdef')) # remove trailing letters
     if second_digit < 8:
         # TODO: anything for 0.7 and below if needed
         pass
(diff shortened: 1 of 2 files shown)


control/config.py
```diff
@@ -166,9 +166,16 @@ def use_legacy_defaults(version):
     Parameters
     ----------
     version : string
-        version number of the defaults desired. Currently only supports `0.8.3`. 
+        version number of the defaults desired. ranges from '0.1' to '0.8.4'. 
     """
-    if version == '0.8.3': 
+    first_digit = int(version[0])
+    second_digit = int(version[2])
+    if second_digit < 8:
```
bnavigator, line 173: Will break with >=0.10.0

Maybe use setuptool's [`packaging.version.parse`](https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python)?

sawyerbfuller: I found I could make a quick change that addresses this concern - does what I came up with look ok? alternatively, are you sure a dependency on the module you proposed ok to have? (is it always included in python?)

bnavigator: It is an edge case. setuptools should not really be needed after install, or if you get the package from something else. The other option mentioned in that SO question is undocumented but standard lib `distutils.version`. `python-control` already uses that within the test suite.

Although your quick solution works, too. I would be fine with it. (*Not to imply that I have any authority. I am only here to contribute unwarranted suggestions.*)

PR #447, opened 2020-08-10
fix impulse size for discrete-time impulse response

sawyerbfuller:
The input impulse used to drive the impulse response of a discrete-time system should be unit *area* rather than unit *height*. With this change, the discrete-time system's impulse response matches the cont-time systems's, if the discrete time system was sampled with 'impulse-equivalent' approximation according to:

``sysd = sys.sample(0.1, 'impulse')``

A new unit test checks this. 

(this is consistent with MATLAB behavior)

Update: rewrote summary, last version was written after a long work session and it shows : )

### DIFF
--- a/control/tests/timeresp_test.py
+++ b/control/tests/timeresp_test.py
@@ -204,14 +204,15 @@ class TestTimeresp(unittest.TestCase):
         np.testing.assert_array_almost_equal(
             yy, np.vstack((youttrue, np.zeros_like(youttrue))), decimal=4)
 
-        # discrete time
-        self.siso_tf1
+    def test_discrete_time_impulse(self):
+        # discrete time impulse sampled version should match cont time
         dt = 0.1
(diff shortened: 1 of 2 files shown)


control/tests/timeresp_test.py
```diff
@@ -204,6 +204,14 @@ def test_impulse_response(self):
         np.testing.assert_array_almost_equal(
             yy, np.vstack((youttrue, np.zeros_like(youttrue))), decimal=4)
 
+        # discrete time
+        self.siso_tf1
```
bnavigator, line 208:
```suggestion
        sys = self.siso_tf1
```

Should go into separate test function. It is independent of the previous tests.

sawyerbfuller: Thanks for catching that test bug. fixed, and moved unit test to a new function

PR #452, opened 2020-08-17
PEP8 updated semicolons are removed

joaomh:
I just removed the unnecessary semicolons in phaseplot_test.py and other PEP8 adjustment


bnavigator:
This is good, but would have been fixed by #438 at the latest. Please don't bother fixing more style issues with current test files. 
Thank you for your contribution anyway :)

PR #376, opened 2020-02-29
Complex matrices

lytex:
Now I've written new tests that use complex matrices without touching the old ones

PR #468, opened 2020-11-12
fixed StateSpace for complex inputs

IguanaAzul:
The problem was that when trying to simulate a system with complex values in the state space matrices it was by default casting to float, which generated wrong results.

I had to create a pull request to scipy as well as the dlsim function has a similar issue:
https://github.com/scipy/scipy/pull/13075

I also created an example script where this problem showed up and was fixed.

### DIFF
--- a/control/statesp.py
+++ b/control/statesp.py
@@ -70,7 +70,7 @@ __all__ = ['StateSpace', 'ss', 'rss', 'drss', 'tf2ss', 'ssdata']
 
 # Define module default parameter values
 _statesp_defaults = {
-    'statesp.use_numpy_matrix': True,
+    'statesp.use_numpy_matrix': False,
     'statesp.default_dt': None,
     'statesp.remove_useless_states': True,
     }
@@ -98,11 +98,11 @@ def _ssmatrix(data, axis=1):
     # Convert the data into an array or matrix, as configured
     # If data is passed as a string, use (deprecated?) matrix constructor
     if config.defaults['statesp.use_numpy_matrix']:
-        arr = np.matrix(data, dtype=float)
+        arr = np.matrix(data, dtype=complex)
     elif isinstance(data, str):
-        arr = np.array(np.matrix(data, dtype=float))
+        arr = np.array(np.matrix(data, dtype=complex))
     else:
-        arr = np.array(data, dtype=float)
+        arr = np.array(data, dtype=complex)
     ndim = arr.ndim
     shape = arr.shape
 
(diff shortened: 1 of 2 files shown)


bnavigator:
Hi,

thanks for the contribution. Unfortunately the change is not straight forward. Are you aware of the discussion in #371, #372 and #376?

control/statesp.py
```diff
@@ -70,7 +70,7 @@
 
 # Define module default parameter values
 _statesp_defaults = {
-    'statesp.use_numpy_matrix': True,
+    'statesp.use_numpy_matrix': False,
```
bnavigator, line 73: Please don't change the default setting here. We have a road map in mind when this should happen and it should be independent of the support for complex types.

bnavigator: That said, actually #438 takes [complex inputs into account](https://github.com/python-control/python-control/pull/438/files#diff-b5aac033a6dd07653e82d9f5a40afaa775010c2d2d5af8beeb159487b566df72)  already!

IguanaAzul: Ok, I will change this back

IguanaAzul:
> Hi,
> 
> thanks for the contribution. Unfortunately the change is not straight forward. Are you aware of the discussion in #371, #372 and #376?

Oh, I wasn`t aware of those discussions.
I think the fix I did conjugated with the fix on the scipy PR will work for input matrices with complex values.
https://github.com/scipy/scipy/pull/13075

control/statesp.py
IguanaAzul: > That said, actually #438 takes [complex inputs into account](https://github.com/python-control/python-control/pull/438/files#diff-b5aac033a6dd07653e82d9f5a40afaa775010c2d2d5af8beeb159487b566df72) already!

Oh, I see, should I close my PR?

bnavigator:
Travis is again not running, but without testing, I think your change will create a lot of casting warnings for cases, when users use control and expect real valued results. Given that #438 takes care of the correct type depending on the input, it should also resolve your use-case. That PR has merge-conflicts with current master right now, but are you able to check it out and test it anyway?

bnavigator:
I closed it because the tests fail and the issue is presumably resolved in another PR. Please feel free to open a new one if you think there is a better way than the approach in #438 or that your example script is a useful enhancement of the existing examples. But I honestly fail to see the point of the example beyond the fact that it demonstrates that python-control can't deal with complex state-space matrices yet.

Issue #421, opened 2020-06-20
A bug in bdalg.connect function


SeungKil-Paek:
Dear all,

I found a bug in the connect function in the bdalg.py:

335:            elif outp < 0 and -outp >= -sys.outputs:

if outp < 0, the index meant is -outp because the negative sign is to tell it is negative feedback.
(-outp) must be in the range [1, sys.outputs]. As outp < 0, -outp > 0 already. Hence, the remaining thing is that -outp <= sys.outputs. The code must be corrected like this:

335:            elif outp < 0 and -outp **_<= sys.outputs_**:

Regards.

Seung-Kil

murrayrm:
If you can generate a simple example that illustrates the bug, this would help us create a unit test to go along with the fix.

SeungKil-Paek:
It looks to me that the intention of the current code is to neglect interconnections out of range .
Hence, the result will be the same as that only with valid interconnections.

I made test code like this:

```python
import numpy as np

from control.statesp import StateSpace
from control.bdalg import append, connect

class Test(object):
    def __init__(self):
        self.sys1 = StateSpace([[1., -2], [3., -4]], [[5.], [7]], [[6, 8]], [[9.]])
        self.sys2 = StateSpace([[-1.]], [[1.]], [[1.]], [[0.]])

    def test_wrong_Q_arg(self):
        sys = append(self.sys1, self.sys2)
        
        # wrong negative feedback interconnection will be neglected
        Q1 = [[1, 2], [2, -3]] 
        ans1 = connect(sys, Q1, [2], [1, 2])

        Q2 = [[1, 2]]  
        ans2 = connect(sys, Q2, [2], [1, 2])
        np.testing.assert_array_almost_equal(ans1.A, ans2.A)
        np.testing.assert_array_almost_equal(ans1.B, ans2.B)
        np.testing.assert_array_almost_equal(ans1.C, ans2.C)
        np.testing.assert_array_almost_equal(ans1.D, ans2.D)

Test().test_wrong_Q_arg()
```

However, the test ends with IndexError.

```

Volumes/ssd1/Applications/anaconda3/lib/python3.7/site-packages/control/bdalg.py in connect(sys, Q, inputv, outputv)
    334                 K[inp,outp-1] = 1.
    335             elif outp < 0 and -outp >= -sys.outputs:
--> 336                 K[inp,-outp-1] = -1.
    337     sys = sys.feedback(np.array(K), sign=1)
    338 

[...]

PR #474, opened 2020-12-19
Fix bdalg.connect

sawyerbfuller:
Fixes bug in index check in `bdalg.connect` described in #421. 

Also added a few more checks, helpful error messages, and unit tests.

### DIFF
--- a/control/bdalg.py
+++ b/control/bdalg.py
@@ -303,10 +303,10 @@ def connect(sys, Q, inputv, outputv):
         System to be connected
     Q : 2D array
         Interconnection matrix. First column gives the input to be connected.
-        The second column gives the index of an output that is to be fed into 
-        that input. Each additional column gives the index of an additional 
+        The second column gives the index of an output that is to be fed into
+        that input. Each additional column gives the index of an additional
         input that may be optionally added to that input. Negative
-        values mean the feedback is negative. A zero value is ignored. Inputs 
+        values mean the feedback is negative. A zero value is ignored. Inputs
         and outputs are indexed starting at 1 to communicate sign information.
     inputv : 1D array
         list of final external inputs, indexed starting at 1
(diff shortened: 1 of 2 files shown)


roryyorke:
Looks good to me.  One minor stylistic comment: you don't need `as context` if you're not going to use `context`.

control/bdalg.py
```diff
@@ -325,15 +327,34 @@ def connect(sys, Q, inputv, outputv):
...
+        raise IndexError( 
+            "outputv index %s out of bounds"%outputv[np.where(index_errors)])
+    index_errors = (Q[:,0:1] - 1 > sys.inputs) | (Q[:,0:1] < 1)
+    if np.any(index_errors):  
+        raise IndexError( 
+            "Q input index %s out of bounds"%Q[np.where(index_errors)])
+    index_errors = (np.abs(Q[:,1:]) - 1 > sys.outputs)
+    if np.any(index_errors):  
+        raise IndexError( 
+            "Q output index %s out of bounds"%Q[np.where(index_errors)])
```
bnavigator, line 347: - Style issue: no white space between format strings and argument. 
- %s is not the correct format. Indices are integers.

control/tests/bdalg_test.py
```diff
@@ -270,6 +272,46 @@ def test_feedback_args(self):
         sys = ctrl.feedback(1, frd)
         self.assertTrue(isinstance(sys, ctrl.FRD))
 
+    def testConnect(self):
```
bnavigator, line 275: These tests only look for failures. What about tests that successfully connect systems?

Obviously I will pytest.parametrize this for #438

bnavigator (changes requested):
Looks good, thanks! One step closer to 0.8.4 :tada: 

Some more comments about style and the tests.

control/tests/bdalg_test.py
sawyerbfuller: > These tests only look for failures. What about tests that successfully connect systems?

good idea; added. 
 
> Obviously I will pytest.parametrize this for #438

thx!

control/bdalg.py
sawyerbfuller: fixed style issue.

re print: it to prints out an array of failing indices. is there a better way than just casting to a string like this?

control/bdalg.py
```diff
@@ -302,14 +302,16 @@ def connect(sys, Q, inputv, outputv):
...
-        Interconnection matrix. First column gives the input to be connected
-        second column gives the output to be fed into this input.  Negative
-        values for the second column mean the feedback is negative, 0 means
-        no connection is made.  Inputs and outputs are indexed starting at 1.
+        Interconnection matrix. First column gives the input to be connected.
+        The second column gives the index of an output that is to be fed into 
+        that input. Each additional column gives the index of an additional 
+        input that may be optionally added to that input. Negative
+        values mean the feedback is negative. A zero value is ignored. Inputs 
+        and outputs are indexed starting at 1 to communicate sign information.
```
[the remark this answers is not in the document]
sawyerbfuller: found it i think

control/bdalg.py
bnavigator: You're right. According to the [doc](https://docs.python.org/3/library/stdtypes.html#old-string-formatting), `%s` in old style formatting converts any argument using `str()`, which should always work for an array of integers.

bnavigator (approved):

PR #476, opened 2020-12-22
Documentation update

murrayrm:
This PR has a number of small documentation updates the I collected over time.  All minor.


bnavigator (approved):

control/statefbk.py
```diff
@@ -373,9 +374,9 @@ def lqr(*args, **keywords):
 
     Returns
     -------
-    K: 2D array
+    K: 2D array (or matrix)
```
bnavigator, line 377: numpydoc:

```suggestion
    K : 2D array (or matrix)
```
Same for all the other functions in this file, which are a mix of numpydoc and non-numpydoc right now.

control/statefbk.py
```diff
@@ -389,6 +390,12 @@ def lqr(*args, **keywords):
     --------
     lqe
 
+    Notes
+    -----
+    The return type for `K` and `S` depends on the default class set for
+    state space operations.  By default, this is the Numpy `matrix`
+    class in this release, but this can be reconfigured using the
+    :func:`~control.use_numpy_matrix` function.
```
bnavigator, line 398: This is also true for most of the other functions in this file. Maybe a general note referenced here and in those functions would be preferable. Cf. #439

Issue #395, opened 2020-04-22
control.matlab.markov does not work


nrgopalrao:
Hi,
I am new to this and it might be a silly doubt but since i could not find an answer anywhere, I am opening an issue here.

My problem is that I need to estimate initial states for my state space model. This requires that I find the markov parameters. I checked the documentation and tried the following example.
`import control.matlab as mt
inp=np.array([1,2])
outp=np.array([2,4])
mrk= mt.markov(outp,inp,1)`

I just took one input and one output as shown above and tried to run the code. It always gives me the following error

"Traceback (most recent call last):

  File "<ipython-input-10-c31db07b63e1>", line 3, in <module>
    mrk= mt.markov(outp,inp,1)

  File "C:\Users\Gopal123\Anaconda3\lib\site-packages\control\modelsimp.py", line 421, in markov
    Ulast = np.vstack((0, np.reshape(UU[0:n-1, m-2], (-1, 1))))

IndexError: too many indices for array
"
I am not able to understand the reason for this. I'd appreciate it if anyone can help me out. Thanks.

murrayrm:
Here's the code from the unit tests for that function:
```
U = np.array([[1.], [1.], [1.], [1.], [1.]])
Y = U
M = 3
H = markov(Y,U,M)
```
This seems to work, though with a warning message.

bnavigator:
I think you found a bug in the `markov` function. Contrary to [python-control convention](https://python-control.readthedocs.io/en/0.8.3/conventions.html#time-series-data), this function expects the time series of input and output data as a 2D array, where the time samples are by rows not columns.

```
Python 3.8.2 (default, Mar 05 2020, 18:58:42) [GCC]
Type "copyright", "credits" or "license" for more information.

IPython 7.13.0 -- An enhanced Interactive Python.

In [1]: import numpy as np

In [2]: from control.modelsimp import markov

In [3]: inp = np.array([[1,2]])

In [4]: outp = np.array([[2,4]])

In [5]: inp.T
Out[5]: 
array([[1],
       [2]])

In [6]: outp.T
Out[6]: 
array([[2],
       [4]])

In [7]: markov(outp,inp,1)
Traceback (most recent call last):

  File "<ipython-input-7-a693851d44f5>", line 1, in <module>
    markov(outp,inp,1)

  File "/usr/lib/python3.8/site-packages/control/modelsimp.py", line 424, in markov
    UU = np.hstack((UU, Ulast))

[...]

PR #478, opened 2020-12-24
updated markov() to add tranpose keyword + default warning

murrayrm:
This PR partially addresses issue #395 by adding a `transpose` keyword to the `markov` function that allows transposed inputs.  In the current release, the `markov` function is not consistent with the normal time-series convention (see discussion in #395) so a warning message is generated if `transpose` is not set explicitly to `True`.

These files will need to be updated again in the 0.9.0 release so that the default for the `transpose` keyword changes from `True` to `False`.

### DIFF
--- a/control/modelsimp.py
+++ b/control/modelsimp.py
@@ -45,17 +45,20 @@ from __future__ import print_function
 
 # External packages and modules
 import numpy as np
-from .exception import ControlSlycot
+import warnings
+from .exception import ControlSlycot, ControlMIMONotImplemented
 from .lti import isdtime, isctime
 from .statesp import StateSpace
 from .statefbk import gram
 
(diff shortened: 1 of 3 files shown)


bnavigator:
This PR itself LGTM. But are we sure that the algorithm actually works for real data? The data used in the test is worthless to determine any markov parameters. (https://github.com/python-control/python-control/issues/395#issuecomment-618418178)

murrayrm:
Good point on whether the calculation was right.  Looks like it wasn't...

In this latest push, I have:

* Updated the markov() algorithm to compute the correct coefficients (the old code is there for comparison, along with a new algorithm + reference).
* Reverted to the original `markov_test.py` (to make sure tests are backward compatible) and moved all of the "enhanced" testing into `markov_test_array.py`.
* Added a new unit test in `markov_test_array.py` to make sure the calculation is correct.

bnavigator (approved):

PR #480, opened 2020-12-28
Switch default state space matrix type to 'array' (instead of 'matrix')

murrayrm:
This is the initial commit to the 0.9.0 release, which changes the default value of state space matrices to be `numpy.array` (really `ndarray`) intead of `numpy.matrix`.

I also updated the `use_legacy_defaults` code to be a bit more general, so that as we make changes we can allow backward compatibility.


bnavigator (approved):
More work created for the necessary rebase of #438, but that's okay.

PR #481, opened 2020-12-29
Stop testing Python 2.7 and scipy 0.19

bnavigator:
#447, which did not fail in the PR tests, started to fail in master after the merge, because `cont2discrete(..., 'impulse')` fails.

Following NEP 29 we could even drop 3.6. But that is still active in distributions like openSUSE Leap 15.x, so I would like to keep that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,11 @@ cache:
     - $HOME/.local
 
 python:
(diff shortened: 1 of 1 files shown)


bnavigator:
Let's defer this to another point in time.

Issue #467, opened 2020-11-10
Phase shift in bode plot


SebastienDeriaz:
Given the transfer function
G = 1/s^3
We should see a constant negative phase of -270°, which is confirmed in matlab

![image](https://user-images.githubusercontent.com/71486197/98642118-2c613280-232d-11eb-8987-59973bff4dcb.png)

However, both **control.bode** and **control.matlab.bode** give 90° instead of -270°. Although technically the same value, it does get confusing.

```
import control
import control.matlab
s = control.tf('s')

G = 1/s**3

control.bode(G);
```
![image](https://user-images.githubusercontent.com/71486197/98642553-eb1d5280-232d-11eb-85d0-d9d6ebfa73d1.png)

```
control.matlab.bode(G);
```
![image](https://user-images.githubusercontent.com/71486197/98642584-f7091480-232d-11eb-9ad1-d16e6431346d.png)

What i've tried so far
- [x] using different transfer functions (1/s^4 or 1/s^2 * 1/(s+10))
- [x] specifying "omega" with a frequency range (as suggested here : https://ch.mathworks.com/matlabcentral/answers/99531-why-is-the-phase-that-is-returned-by-the-bode-function-wrapping-in-the-control-system-toolbox-5-1-r )
- [x] extending the "omega" range

Is it possible to get the "absolute" phase of the transfer function ? If not, i would suggest that `control.matlab.bode` be corrected as to reflect the behavior of matlab's bode

Thanks for your help !

SebastienDeriaz:
My (limited) understanding of the function is that **bode** uses **freqresp** which uses **evalfr** and then **numpy.angle**. Since **evalfr** returns a complex number it isn't directly possible to get the "absolute" angle so the issue i'm reporting seems to require a dedicated code to determine a "reference phase" of some sort.
My excuses for my current level of python (or gitHub) doesn't lead me much further than that

bnavigator:
Thanks, for bringing this up.

Related issue: #388 

> My excuses for my current level of python (or gitHub) doesn't lead me much further than that

You are nevertheless very welcome to practice your Python and GitHub skills by trying to fix the issue and submitting a pull request :student:

ilayn:
The phase correction is only possible when you have the pole-zero structure and the behavior `w->0`, so you have to compute the initial phase to get the right amount of shift which is typically a minus sign attached to the `atan2` computation. 

[...]

PR #494, opened 2020-12-31
implementation of initial_phase, wrap_phase keywords for bode()

murrayrm:
This PR adds `initial_phase` and `wrap_phase` keywords to `bode`, which allow you to tune the behavior of the phase portion of a Bode plot.  The default behavior should more or less match previous behavior, except that the default initial phase will be between 0 and -360 degrees (so 1/s^3 will start with phase -270).

Resolves issue #467.

### DIFF
--- a/control/freqplot.py
+++ b/control/freqplot.py
@@ -135,7 +135,8 @@ def bode_plot(syslist, omega=None,
     initial_phase : float
         Set the reference phase to use for the lowest frequency.  If set, the
         initial phase of the Bode plot will be set to the value closest to the
-        value specified.  Default is 180 if wrap_phase is False, 0 if
+        value specified.  Units are in either degrees or radians, depending on
+        the `deg` parameter. Default is -180 if wrap_phase is False, 0 if
         wrap_phase is True.
     wrap_phase : bool or float
         If wrap_phase is `False`, then the phase will be unwrapped so that it
--- a/control/freqplot.py
+++ b/control/freqplot.py
@@ -78,6 +78,7 @@ _bode_defaults = {
     'bode.deg': True,           # Plot phase in degrees
     'bode.Hz': False,           # Plot frequency in Hertz
     'bode.grid': True,          # Turn on grid for gain and phase
+    'bode.wrap_phase': False,   # Wrap the phase plot at a given value
 }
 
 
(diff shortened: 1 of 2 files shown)


control/freqplot.py
```diff
@@ -131,7 +132,18 @@ def bode_plot(syslist, omega=None,
     grid : bool
         If True, plot grid lines on gain and phase plots.  Default is set by
         `config.defaults['bode.grid']`.
-
+    initial_phase : float
+        Set the reference phase to use for the lowest frequency.  If set, the
+        initial phase of the Bode plot will be set to the value closest to the
+        value specified.  Default is 180 if wrap_phase is False, 0 if
+        wrap_phase is True.
```
bnavigator, line 139: Mention, that this is in degrees or radians depending on the deg parameter?

control/tests/freqresp_test.py
```diff
@@ -271,3 +272,68 @@ def test_options(editsdefaults):
     # Make sure we got the right number of points
     assert numpoints1 != numpoints3
     assert numpoints3 == 13
+
+@pytest.mark.parametrize(
+    "TF, initial_phase, default_phase, expected_phase",
+    [pytest.param(ctrl.tf([1], [1, 0]),
```
bnavigator, line 278: This initializes the systems in the collection phase of pytest. During that phase, the filter for the matrix warnings with `use_numpy_matrix` is not active yet:
https://travis-ci.org/github/python-control/python-control/jobs/752332682#L2031

Instead of `TF` you can parametrize two parameters `num`, `den` (or just `den` in this case) and then call `ctrl.tf()` inside the test functuin

control/tests/freqresp_test.py
```diff
@@ -271,3 +272,68 @@ def test_options(editsdefaults):
...
+    # Check initial phase of standard transfer functions
+    mag, phase, omega = ctrl.bode(TF)
+    assert(abs(phase[0] - default_phase) < 0.1)
+
+    # Now reset the initial phase to +180 and see if things work
+    mag, phase, omega = ctrl.bode(TF, initial_phase=initial_phase)
+    assert(abs(phase[0] - expected_phase) < 0.1)
+
+    # Make sure everything works in rad/sec as well
+    # Turn off plotting since that seems to slow things down a lot (?)
```
bnavigator, line 307: What is causing this? deg=False?

bnavigator (changes requested):

control/tests/freqresp_test.py
bnavigator: Hmm, the warnings are already in master, and I can not reproduce them on my local machine with master or your branch. Maybe I am wrong here.

bnavigator: Okay, it was a different recently merged PR. See #496. The warnings are about `StateSpace` matrices, not something from `tf()`

control/tests/freqresp_test.py
murrayrm: I tried debugging, but wasn't able to figure out what was making the difference.  I looked through all of the processing that happens if deg=False and couldn't see anything that would make the difference.  Everything *looks* fine, it just adds ~6 seconds to the run.  ??

control/freqplot.py
murrayrm: Fixed.

control/tests/freqresp_test.py
bnavigator: Maybe it is not `deg=False` but the existing of the previously created figures in the same test?

murrayrm: Indeed, that turned out to be the problem!

bnavigator (approved):

PR #497, opened 2021-01-02
I/O system improvements: linearize, interconnect, docstrings

murrayrm:
This PR implements a couple of enhancements for the I/O systems package, mainly related to linear systems:

* The `linearize` command now preserves signal names for the linearized system, so that you can use signal names when forming and `InterconnectedSystem`.

* Interconnections of `LinearIOSystem`s are now treated as `LinearICSystems` (a subclass of `InterconnectedSystem` and `LinearIOSystem`) so that `StateSpace` functions will still work on them and they still keep track of the underlying interconnection structure.

* There is now an `interconnect` function that allows the construction of an `LinearICSystems ` or a `InterconnectedSystem` depending on whether the subsystems are all linear or not.  This basically serves as a replacement for the `connect` function for LTI systems and allows named interconnection of linear or nonlinear systems.

In addition, I cleaned up the description for how to specify interconnections so that tuples are only used for the lowest level signal descriptions; everything else is a list.  (Tuples will still work, but there are some ambiguities when you use tuples since `outlist=(3, 2, 1)` could be interpreted as a signal description for a single signal (subsystem 3, signal 2, with gain 1) or as a list of signals for SISO systems (the single outputs from subsystems 3, 2, and 1).

Additional small changes:

* The `rss` and `drss` functions now allow the `strictly_proper` keyword, which sets the D matrix to zero.  (I needed this to avoid occasional errors in unit tests for the `interconnect` function.

* Added some configuration variables that allow you to change the way that states in interconnected systems are named (default was `sys.state` but is `sys_state` for 0.9+) and the way that duplicate systems are named (default was `copy of sys` but is `sys$copy` for 0.9+).

* Added unit tests for all of the above.

* Updated examples to match the newly documented conventions for signal list specification.

### DIFF
--- a/control/tests/iosys_test.py
+++ b/control/tests/iosys_test.py
@@ -219,8 +219,11 @@ class TestIOSys:
 
         # If we copy signal names w/out a system name, append '_linearized'
         ct.use_legacy_defaults('0.8.4')
+        ct.config.use_numpy_matrix(False)       # get rid of warning messages
         linearized = kincar.linearize([0, 0, 0], [0, 0], copy=True)
         assert linearized.name == kincar.name + '_linearized'
+
+        ct.reset_defaults()
         ct.use_legacy_defaults('0.9.0')
         linearized = kincar.linearize([0, 0, 0], [0, 0], copy=True)
         assert linearized.name == kincar.name + '$linearized'
--- a/control/tests/iosys_test.py
+++ b/control/tests/iosys_test.py
@@ -1293,7 +1293,7 @@ def test_linear_interconnection():
             ['sys2.y[1]'],
             ['sys2.u[1]']])
     assert isinstance(nl_connect, ios.InterconnectedSystem)
-    assert not isinstance(nl_connect, ios.LinearInterconnectedSystem)
+    assert not isinstance(nl_connect, ios.LinearICSystem)
 
     # Now take its linearization
     ss_connect = nl_connect.linearize(0, 0)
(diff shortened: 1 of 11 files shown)


control/tests/iosys_test.py
```diff
@@ -195,6 +204,36 @@ def kincar_output(t, x, u, params):
...
+        assert linearized.find_state('theta') == 2
+
+        # If we copy signal names w/out a system name, append '_linearized'
+        ct.use_legacy_defaults('0.8.4')
+        ct.config.use_numpy_matrix(False)       # get rid of warning messages
+        linearized = kincar.linearize([0, 0, 0], [0, 0], copy=True)
+        assert linearized.name == kincar.name + '_linearized'
+
+        ct.reset_defaults()
+        ct.use_legacy_defaults('0.9.0')
```
bnavigator, line 227: See below

bnavigator:
https://travis-ci.org/github/python-control/python-control/jobs/752920076#L2029

```
_______________ TestIOSys.test_linearize_named_signals[arrayout] _______________
....
E       PendingDeprecationWarning: the matrix subclass is not the recommended way to represent matrices or deal with linear algebra (see https://docs.scipy.org/doc/numpy/user/numpy-for-matlab-users.html). Please adjust your code to use regular ndarray.
...
FAILED control/tests/iosys_test.py::TestIOSys::test_linearize_named_signals[arrayout]
FAILED control/tests/iosys_test.py::TestIOSys::test_sys_naming_convention[arrayout]
FAILED control/tests/iosys_test.py::TestIOSys::test_signals_naming_convention_0_8_4[arrayout]
FAILED control/tests/iosys_test.py::TestIOSys::test_duplicates[arrayout] - Pe...
```

Autouse fixture `matarrayout` defines `matrixerrorfilter`, which is not overridden by `matrixfilter`. You need to locally override `matarrayout`.

Issue #459, opened 2020-08-22
numpy.ndarray * control.TransferFunction gives array object


roryyorke:
`TransferFunction` doesn't have the `__array_priority = 11` class attribute that `StateSpace` and `FrequencyResponseData` do.  Is this on purpose?  See below for results of `array * tf`, which results in an array (dtype object), and `array * ss`, which results in a SS object.  I found this while investigating the behaviour of `bdalg.connect` in relation with #421 

Should multiplication of scalar transfer functions by arrays be special-cased so that, e.g., with scalar tf `k`, `k * eye(5)` gives a 5x5 TF with identical entries on the diagonal?  I think this sort of convention is used in literature, but it might be tricky to get something consistent here (e.g., should 1-input 1-output state-space systems do something analogous?; should one do general broadcasting?)

```
(control-dev) rory@rory-latitude:~/src/python-control$ ipython
Python 3.8.5 | packaged by conda-forge | (default, Jul 31 2020, 02:39:48) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.17.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: !git describe                                                                                                  
0.8.3-62-g6ede92e

In [2]: import control                                                                                                 

In [3]: import numpy as np                                                                                             

In [4]: np.array([[1]]) * control.tf([1], [1,1])                                                                       
Out[4]: array([[TransferFunction(array([1.]), array([1, 1]))]], dtype=object)

In [5]: np.array([[1]]) * control.ss([[-1]], [[1]], [[1]], [[0]])                                                      
Out[5]: StateSpace(array([[-1.]]), array([[1.]]), array([[1.]]), array([[0.]]))
```

roryyorke:
For the purposes of being able to use `bdalg.connect` with TransferFunctions, it's probably sufficient just to add the `__array_priority__` entry to the class in question.  However, I wondered about general type conversion in python-control; see below.

For the operations +, -, *, /, I would expect the following priority of "type conversion", where conversions are always to earlier entries (so, e.g., FRD + StateSpace should give type FRD).  For FRD, this is based on the idea that it's possible, and least in principle, to go from any of the others to FRD type, but the opposite is not possible (or, at least, not straightforward).  

  1. FRD
  2. LinearIOSystem
  3. StateSpace
  4. TransferFunction
  5. numpy.ndarray / Python float

[...]

PR #498, opened 2021-01-04
TransferFunction array priority plus system type conversion checking

murrayrm:
(This PR builds on #497, which is required for some of the functionality.)

This PR updates the `__array_priority__` variable in the TransferFunction class that is used by `numpy`, as described in issue #459.  With this change, multiplying a transfer function by an `ndarray` object gives a transfer function.

I also created a new unit test (`control/tests/type_converstion_test.py`) that goes through the matrix of interconnections in #497 and confirms that output type for each operation.  In this current PR, no new functionality is included, but I did mark those entries that should (eventually) be updated.  Depending on what people think about the PR, we can either leave those for another time or put them into this PR.


bnavigator:
So, wait on #497?

> I also created a new unit test (`control/tests/type_converstion_test.py`)

On first glance, the test file looks good.

murrayrm:
I'll update #497 tonight and also do one more small change here (putting in the final table of conversations for later use).  Once #497 is integrated, I'll make sure this is rebased against that (should be OK, I think, since I started with that as my initial branch) and then we can merge this one as well.

control/tests/type_conversion_test.py
```diff
@@ -117,3 +122,66 @@ def test_xferfcn_ndarray_precedence(opname, ltype, rtype, expected, sys_dict):
...
+#   * When combining state space or transfer with linear I/O systems, the
+#   * output should be of type Linear IO system, since that maintains the
+#   * underlying state space attributes.
+#
+# Note: tfx = non-proper transfer function, order(num) > order(den)
+#
+
+type_list = ['ss', 'tf', 'tfx', 'frd', 'lio', 'ios', 'arr', 'flt']
+conversion_table = [
+    # L / R ['ss',  'tf',  'tfx', 'frd', 'lio', 'ios', 'arr', 'flt']
```
bnavigator, line 152:
```suggestion
    # L \ R ['ss',  'tf',  'tfx', 'frd', 'lio', 'ios', 'arr', 'flt'] 
```

Issue #464, opened 2020-11-03
root_locus() plot figure size issue - ipynotebook


RicardoSQueen:
Hello,
I'm currently running python-control 0.8.3 on python 3.8.5.
When plotting 'root_locus' in a interactive python notebook, I get a plot way offset to the right, according to the image below:
![image](https://user-images.githubusercontent.com/64474885/98043988-ef64df80-1e04-11eb-85d1-f0f8922f8f20.png)
The unexpected for me, though, is that when the exact same code was executed directly from the terminal, the plot was perfect.
The plot issues appears on every interactive notebook environment: jupyter notebook, jupyter lab and visual studio code's jupyter implementation. I'm running these versions of jupyter:
notebook                       6.1.4
jupyter_client                 6.1.7
jupyter_contrib_core     0.3.3
jupyter_core                  4.6.3
jupyter_latex_envs        1.4.6
jupyterlab                      2.2.8
jupyterlab_pygments     0.1.2
jupyterlab_server          1.2.0

here is the code for reproducing:
```
from control import *
import matplotlib.pyplot as plt

num = [25]
den = [1,6,25]
x1 = tf(num, den)
rlist, klist = root_locus(x1)
plt.show()
```

sawyerbfuller:
Hi Ricardo, thanks for pointing out the issue. I think this has been fixed in #413 and is now in the dev branch (see screenshot below running the dev branch). Release 0.8.4 is right around the corner and includes these fixes. 

<img width="530" alt="image" src="https://user-images.githubusercontent.com/58706249/98150675-e2d39c00-1e83-11eb-9072-961d683ddcb2.png">

RicardoSQueen:
Hey Sawyer,
Sorry, I tried finding an issue related to my issue but I didn't find any.. oh well. Awesome to know that!
I did install the dev branch over here and it seems that it's still not ok. Here are some screenshots:
![image](https://user-images.githubusercontent.com/64474885/99159441-15b73480-26bb-11eb-8cb3-c2631e64a713.png)
![image](https://user-images.githubusercontent.com/64474885/99159445-1bad1580-26bb-11eb-8f56-370d9227a887.png)
![image](https://user-images.githubusercontent.com/64474885/99159448-21a2f680-26bb-11eb-83ed-93ebf4832411.png)
The plot is all white to me.. And it took a little while (longer than usual)
Running your code worked out great:
[...]

PR #503, opened 2021-01-09
fix rlocus plotting problem in Jupyter notebooks + comments, PEP8

murrayrm:
This PR resolves issue #464.  I added some comments to the code along the way as I was trying to sort out what was going wrong.  No easy unit tests for this one, I think, since it seems like the issue is restricted to Jupyter notebooks and has to do with the way they generate plots (I guess).

Before (need to zoom *way* in to see the plot)
jupyter_bug-before

After:
jupyter_bug-after

### DIFF
--- a/control/rlocus.py
+++ b/control/rlocus.py
@@ -222,6 +222,14 @@ def root_locus(sys, kvect=None, xlim=None, ylim=None,
         ax.set_xlabel('Real')
         ax.set_ylabel('Imaginary')
 
+        # Set up the limits for the plot
+        # Note: need to do this before computing grid lines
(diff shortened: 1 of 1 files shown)


bnavigator:
LGTM.

Could you post before-after pictures here?

The coveralls increase by 11% is a lie, because the last Travis run on master failed with server errors.

Uncovered line: https://coveralls.io/builds/36205925/source?filename=control%2Frlocus.py#L247. At least call the plot function with a dtime sys once?

Maybe https://matplotlib.org/devel/testing.html#writing-an-image-comparison-test is also worth a try.

Issue #446, opened 2020-08-04
Move from legacy Travis CI integration to a Github Actions based system


bnavigator:
The link to Travis CI seems broken. Maybe related to https://travis-ci.community/t/github-status-not-posted-on-commits-on-repositories-using-legacy-service-integration/7798/4

murrayrm:
I remember this was going to go away, but wasn't sure when.  We need to set up a GitHub action instead.  I'll try to do that in the next few days.

bnavigator:
I agree that a move to Github actions will be necessary at some time.

Travis still reports the new run after the most recent merge. And the PR runs are still there just not reported on the PR pages.

bnavigator:
The PR run checks came back. Not sure. why. Probably a refresh of Travis' API tokens and rate limits.

murrayrm:
You (@bnavigator) apparently put in some sort of request to set up the GitHub/Travis CI integration.  I got an e-mail asking for approval of the permissions and I approved.  So presumably that turned things back on (and hopefully didn't break anything!).

bnavigator:
I think they even came back before the request. I tried to follow the migration advice from [this guide](https://docs.travis-ci.com/user/migrate/legacy-services-to-github-apps-migration-guide/#migrating-and-activating-all-repositories).

We are still on legacy though:

travis-ci.**org**
![Screenshot_20200818_165015](https://user-images.githubusercontent.com/4623504/90528463-eef86900-e172-11ea-9d4b-ff15b9dc5ed2.png)

travis-ci.**com**
![Screenshot_20200818_165356](https://user-images.githubusercontent.com/4623504/90528969-7d6cea80-e173-11ea-890c-19f66ae77bd6.png)



Nothing broken yet. Please tell me when, I am moving to fast!

bnavigator:
Note that this banner is currently popping up on travis-ci.org:

> Please be aware travis-ci.org will be shutting down in several weeks, with all accounts migrating to travis-ci.com. Please stay tuned here for more information.

sawyerbfuller:
are they dropping free service for open source projects? 

https://news.ycombinator.com/item?id=25338983

ilayn:
They already did. But you can request for more free credits as they expire from Travis. For SciPy, we had to switch to GitHub Actions and Azure Pipelines for most of the CI work. We can't leave completely because we are building wheels for ppc64 and arm but python-control can. It's not that difficult to setup but certainly wastes some maintainer time

[...]

PR #504, opened 2021-01-10
CI tests via GitHub actions and conda

murrayrm:
This is a fairly minimal version of continuous integration (CI) tests using GitHub actions consisting of two workflows:

* `python-package-conda`: This workflow uses a conda-based install for everything and tests only against the latest scipy and Python 3.6 and 3.9.  So compared with our current Travis CI testing, we lose scipy-0.19.1 testing and Python 2.7.  Coverage is implemented, but not sure (yet) whether it properly connects to coveralls.
* `control-slycot-src`: This workflow installs slycot from source, which allows testing against the latest version of slycot (in case we need it). No coverage testing.

This workflow users coveralls and pushes results up to coveralls.io.

Note: I think that as soon as this PR is merged then we should see the workflows show up in the Actions tab on GitHub.  This is working correctly in my [personal fork](https://github.com/murrayrm/python-control/tree/github-actions-conda).

### DIFF
--- a/.github/workflows/python-package-conda.yml
+++ b/.github/workflows/python-package-conda.yml
@@ -1,0 +1,34 @@
+name: Python Package using Conda
+
+on: [push]
+
+jobs:
+  build-linux:
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 5
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Add conda to system path
+      run: |
+        # $CONDA is an environment variable pointing to the root of the miniconda directory
+        echo $CONDA/bin >> $GITHUB_PATH
+    - name: Install dependencies
+      run: |
+        conda env update --file environment.yml --name base
+    - name: Lint with flake8
+      run: |
+        conda install flake8
+        # stop the build if there are Python syntax errors or undefined names
+        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+    - name: Test with pytest
+      run: |
+        conda install pytest
+        pytest
--- a/.github/workflows/python-package-conda.yml
+++ b/.github/workflows/python-package-conda.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       max-parallel: 5
       matrix:
-        python-version: [2.7, 3.6, 3.9]
+        python-version: [3.6, 3.9]
         slycot: ["", "conda"]
         array-and-matrix: [0]
         include:
@@ -19,34 +19,28 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Python
-      uses: actions/setup-python@v2
-    - name: Set up conda
-      run: |
-        echo $CONDA/bin >> $GITHUB_PATH
-        conda create -q -n test-environment python=${{matrix.python-version}} 
     - name: Install dependencies
       run: |
+        # Set up conda
+        echo $CONDA/bin >> $GITHUB_PATH
+        conda create -q -n test-environment python=${{matrix.python-version}} 
         source $CONDA/bin/activate test-environment
-        conda install pip coverage pytest
+
+        # Set up (virtual) X11
+        sudo apt install -y xvfb
+
+        # Install test tools
+        conda install pip coverage pytest 
+        conda install -c conda-forge pytest-xvfb pytest-cov
+
+        # Install python-control dependencies
         conda install numpy matplotlib scipy
-        if [[ '${{matrix.python-version}}' == '2.7' ]]; then
-          # matplotlib for Python 2.7 seems to require X11 to run correctly
-          sudo apt install -y xvfb
+        if [[ '${{matrix.slycot}}' == 'conda' ]]; then
+          conda install -c conda-forge slycot
         fi
-    - name: Install slycot via conda (optional)
-      if: ${{ matrix.slycot == 'conda' }}
-      run: |
-        source $CONDA/bin/activate test-environment
-        conda install -c conda-forge slycot
     - name: Test with pytest
       env:
         PYTHON_CONTROL_ARRAY_AND_MATRIX: ${{ matrix.array-and-matrix }}
       run: |
         source $CONDA/bin/activate test-environment
-        if [[ '${{matrix.python-version}}' == '2.7' ]]; then
-          # Run within (virtual) X11 environment for Python 2.7/matplotlib
-          xvfb-run pytest control/tests
-        else
-          coverage run -m pytest control/tests
-        fi
+        xvfb-run --auto-servernum pytest control/tests
--- a/.github/workflows/python-package-conda.yml
+++ b/.github/workflows/python-package-conda.yml
@@ -3,7 +3,7 @@ name: Conda-based pytest
 on: [push, pull_request]
 
 jobs:
-  build-linux:
+  test-linux:
     runs-on: ubuntu-latest
 
     strategy:
@@ -19,6 +19,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+
     - name: Install dependencies
       run: |
         # Set up conda
@@ -31,18 +32,36 @@ jobs:
 
         # Install test tools
         conda install pip coverage pytest
-        conda install -c conda-forge pytest-cov
+        pip install coveralls
 
         # Install python-control dependencies
         conda install numpy matplotlib scipy
         if [[ '${{matrix.slycot}}' == 'conda' ]]; then
           conda install -c conda-forge slycot
         fi
+
     - name: Test with pytest
       env:
         PYTHON_CONTROL_ARRAY_AND_MATRIX: ${{ matrix.array-and-matrix }}
       run: |
         source $CONDA/bin/activate test-environment
         # Use xvfb-run instead of pytest-xvfb to get proper mpl backend
+        # Use coverage instead of pytest-cov to get .coverage file
         # See https://github.com/python-control/python-control/pull/504
-        xvfb-run --auto-servernum pytest control/tests
+        xvfb-run --auto-servernum coverage run -m pytest control/tests
+
+    - name: Coveralls parallel
+      # https://github.com/coverallsapp/github-action
+      uses: AndreMiras/coveralls-python-action@develop
+      with:
+        parallel: true
+
+  coveralls:
+    name: coveralls completion
+    needs: test-linux
+    runs-on: ubuntu-latest
+    steps:
+    - name: Coveralls Finished
+      uses: AndreMiras/coveralls-python-action@develop
+      with:
+        parallel-finished: true
--- a/.github/workflows/python-package-conda.yml
+++ b/.github/workflows/python-package-conda.yml
@@ -20,14 +20,8 @@ jobs:
         echo $CONDA/bin >> $GITHUB_PATH
     - name: Install dependencies
       run: |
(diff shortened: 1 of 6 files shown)


bnavigator:
I am not done with Slycot yet (coming closer), but see https://github.com/bnavigator/Slycot/blob/3c0cb727d5371996c038c94e9bd68621a5430b8a/.github/workflows/slycot-build-and-test.yml for an example how to deal with coveralls.

control/tests/rlocus_test.py
```diff
@@ -48,6 +50,7 @@ def test_without_gains(self, sys):
         roots, kvect = root_locus(sys, plot=False)
         self.check_cl_poles(sys, roots, kvect)
 
+    @X11only
```
bnavigator, line 53: The failures are not due to X11 but because the figure manager has no toolbar. I guess the matplorlib version/setup is different than on the Travis runners.

Same problem has always been present for the openSUSE rpm package, which I maintain. Forcing the PyQt5 backend instead of Agg helped there, but my tries (short) for Slycot on GitHub worklfows did not succeed. I just skip the two tests in python-control/Slycot#140

control/tests/conftest.py
```diff
@@ -26,7 +26,8 @@
                                           "PendingDeprecationWarning")
 matrixerrorfilter = pytest.mark.filterwarnings("error:.*matrix subclass:"
                                                "PendingDeprecationWarning")
-
+X11only = pytest.mark.skipif(os.getenv("DISPLAY") is None,
+                             reason="requires X11")
```
bnavigator, line 30: Would not work outside of a Unix system. Is DISPLAY even set on macOS?

.github/workflows/python-package-conda.yml
```diff
@@ -0,0 +1,52 @@
...
+            array-and-matrix: 1
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+    - name: Set up conda
+      run: |
+        echo $CONDA/bin >> $GITHUB_PATH
+        conda create -q -n test-environment python=${{matrix.python-version}}
```
bnavigator, line 27: There is an action for that: https://github.com/marketplace/actions/setup-miniconda

.github/workflows/python-package-conda.yml
```diff
@@ -0,0 +1,52 @@
...
+        array-and-matrix: [0]
+        include:
+          - python-version: 3.9
+            slycot: conda
+            array-and-matrix: 1
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
```
bnavigator, line 23: Conflicts with the conda setup

.github/workflows/python-package-conda.yml
```diff
@@ -0,0 +1,52 @@
...
+        conda create -q -n test-environment python=${{matrix.python-version}} 
+    - name: Install dependencies
+      run: |
+        source $CONDA/bin/activate test-environment
+        conda install pip coverage pytest
+        conda install numpy matplotlib scipy
+        if [[ '${{matrix.python-version}}' == '2.7' ]]; then
+          # matplotlib for Python 2.7 seems to require X11 to run correctly
+          sudo apt install -y xvfb
+        fi
```
bnavigator, line 36: Why still bother with Python2? Time to ditch it for the next release.

Also, pytest-xvfb is the way to go.

.github/workflows/python-package-conda.yml
```diff
@@ -0,0 +1,52 @@
...
+      env:
+        PYTHON_CONTROL_ARRAY_AND_MATRIX: ${{ matrix.array-and-matrix }}
+      run: |
+        source $CONDA/bin/activate test-environment
+        if [[ '${{matrix.python-version}}' == '2.7' ]]; then
+          # Run within (virtual) X11 environment for Python 2.7/matplotlib
+          xvfb-run pytest control/tests
+        else
+          coverage run -m pytest control/tests
+        fi
```
bnavigator, line 52: use pytest-cov

.github/workflows/python-package-conda.yml
bnavigator: ... actually I see that you use a github preinstalled miniconda. Missed that. That might be even better than the action.

control/tests/rlocus_test.py
murrayrm: Thanks for the extra info. I'd prefer not to have to factor these out.  Let me see if I can use pytest-xvfb or something similar to get it to work.

control/tests/conftest.py
murrayrm: It is set if you are using X11, so the mark is probably OK.  But as noted above, this may not be the actual thing to check for.  I'll see if there is a better way to trap the issue.

.github/workflows/python-package-conda.yml
murrayrm: I think I might be able to remove this completely.  I meant to do that at one point but lost track of it while playing around to just get something to work.  I'll see if I can remove it and if things still work correctly.

.github/workflows/python-package-conda.yml
murrayrm: Thanks for the suggestion.  Will try shifting to pytest-xvfb.  I figure we may as well test against Python 2.7 for as long as it continues to work.  Amazingly, it is still the default Python on MacOS 11 (Big Sur).

.github/workflows/python-package-conda.yml
murrayrm: Thanks.  Will give it a shot.

.github/workflows/python-package-conda.yml
bnavigator: I tried this method on the Slycot workflow, but it does not work with Windows.

.github/workflows/python-package-conda.yml
bnavigator: If we don't actively remove it, every PR with a failing Py2 test will try to workaround it. I think we should save contributors from this torture.

control/tests/rlocus_test.py
murrayrm: I ended up leaving in these marks since they are useful if you are testing without X11, but also got pytest-xvfb working, so these tests aren't skipped in the GitHub Actions CI testing.

.github/workflows/python-package-conda.yml
```diff
@@ -0,0 +1,46 @@
...
+        conda install numpy matplotlib scipy
+        if [[ '${{matrix.slycot}}' == 'conda' ]]; then
+          conda install -c conda-forge slycot
+        fi
+    - name: Test with pytest
+      env:
+        PYTHON_CONTROL_ARRAY_AND_MATRIX: ${{ matrix.array-and-matrix }}
+      run: |
+        source $CONDA/bin/activate test-environment
+        xvfb-run --auto-servernum pytest control/tests
```
bnavigator, line 46: pytest-xvfb lets you call plain pytest and it starts xvfb-run inside the test suite.

.github/workflows/python-package-conda.yml
```diff
@@ -0,0 +1,46 @@
...
+        echo $CONDA/bin >> $GITHUB_PATH
+        conda create -q -n test-environment python=${{matrix.python-version}} 
+        source $CONDA/bin/activate test-environment
+
+        # Set up (virtual) X11
+        sudo apt install -y xvfb
+
+        # Install test tools
+        conda install pip coverage pytest 
+        conda install -c conda-forge pytest-xvfb pytest-cov
```
bnavigator, line 34: coverage not reported to coveralls yet

control/tests/conftest.py
bnavigator: This disables the tests on systems without X11 like Windows (and macOS?). And what about Linux with Wayland instead of X11?

```suggestion
hastoolbarmode = pytest.mark.skipif(
    hasattr(plt.get_current_fig_manager().toolbar, 'mode'),
    reason="incompatible matplotlib backend")
```

.github/workflows/python-package-conda.yml
murrayrm: I tried that but it was still causing problems with the noX11 tests (they were skipped if I left the current mark in place, and generated an error if I got rid of the mark).  This version seems to work (?).

.github/workflows/python-package-conda.yml
murrayrm: I think we need to merge the PR before this is run and (hopefully) at that point things start to show up in coveralls?

control/tests/conftest.py
murrayrm: I'll see if I can find another way around this, but getting rid of this causes the test to fail under GitHub actions.

.github/workflows/python-package-conda.yml
bnavigator: Probably causes the selection of a different mpl backend then

bnavigator: Then you don't use and need to install pytest-xvfb

.github/workflows/python-package-conda.yml
bnavigator: `coverage` != `coveralls`.

See https://coveralls-python.readthedocs.io/en/latest/usage/configuration.html#github-actions-gotcha and my previously linked workflow draft for Slycot. (Much more complicated than yours! :( )

control/tests/conftest.py
murrayrm: I ended up not needing the mark to get things to work (since we dropped Python 2.7), so ended up reverting this change.

.github/workflows/python-package-conda.yml
murrayrm: Removed pytest-xvfb.

.github/workflows/python-package-conda.yml
```diff
@@ -31,18 +32,36 @@ jobs:
...
           conda install -c conda-forge slycot
         fi
+
     - name: Test with pytest
       env:
         PYTHON_CONTROL_ARRAY_AND_MATRIX: ${{ matrix.array-and-matrix }}
       run: |
         source $CONDA/bin/activate test-environment
         # Use xvfb-run instead of pytest-xvfb to get proper mpl backend
+        # Use coverage instead of pytest-cov to get .coverage file
```
bnavigator, line 49: pytest-cov produces one. See python-control/Slycot#140. Use the --cov parameter.

.github/workflows/python-package-conda.yml
```diff
@@ -31,18 +32,36 @@ jobs:
 
         # Install test tools
         conda install pip coverage pytest
-        conda install -c conda-forge pytest-cov
+        pip install coveralls
```
bnavigator, line 35: coveralls is on conda-forge nowadays

murrayrm:
I think this PR is good to go.  There are some small changes that @bnavigator suggests, but I don't these these substantially add/enhance functionality => we can probably stick with this for now (and see if it all works when merged).

bnavigator:
Sorry, for being so pedantic here. I went through many of these hoops for Slycot and the openSUSE rpm packages so I want to share the "proper" solutions.

- The coveralls jobs should get a proper name tag so that the listing looks nicer. (coveralls-python FAQ previously linked)
  Compare https://coveralls.io/builds/36275777 to https://coveralls.io/builds/36276758
- The coveralls link on the master commit :heavy_check_mark: is still from Travis. Do we remove Travis eventually or do we try to merge those two coveralls reports per PR/commit, Otherwise we will always have to subsequent datapoints in the history, possibly with different values.
- The way miniconda is set up this limits everything to Unix. Maybe even Linux only.

murrayrm:
I'll move these comments to issue #446 so that we can continue to track these improvements. I mainly wanted to get shifted over to some new test infrastructure since Travis CI is getting *so* frustrating. (GitHub actions clears commits in 2-3 minutes, versus what was often 30+ min for Travis).

I think we can leave Travis and GitHub Actions running for a few days, as we see whether new PRs and things are properly handled, but then I think we shift away from Travis completely.

Issue #500, opened 2021-01-07
gangof4_plot() is not covered by unit tests.


bnavigator:
It's a failry complex plot function. Do we know, that it still works? And will it continue to work with future code changes? Better write some unit tests!

PR #505, opened 2021-01-13
Unit tests for gangof4_plot

murrayrm:
This PR adds a unit test for the gang of 4 plots, resolving issue #500.

It also tests whether the new GitHub Actions CI interface is working...


control/tests/baseline_images/freqresp_test/gangof4-pvtol-expected.png
```diff
@@ -0,0 +1 @@
+/Users/murray/Dropbox/macosx/src/python-control/murrayrm/control/tests/baseline_images/freqresp_test/gangof4-pvtol.png
```
bnavigator, line 1: ?

pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
```
bnavigator, line 1: we already have a pytest section in setup.cfg

control/tests/baseline_images/freqresp_test/gangof4-pvtol-expected.png
murrayrm: That's very odd.  Should have been a copy of the image...

pytest.ini
murrayrm: Got it.  Will move to there.

control/tests/freqresp_test.py
```diff
@@ -345,3 +347,26 @@ def test_phase_wrap(TF, wrap_phase, min_phase, max_phase):
...
+    # Transfer functions for dynamics
+    Pi = ctrl.tf([r], [J, 0, 0])        # inner loop (roll)
+
+    k, a, b = 200, 2, 50                # control parameters
+    Ci = k * ctrl.tf([1, a], [1, b])    # lead compensator
+
+    return (Pi, Ci)
+
+# Regression test for Gang of 4 plots
+@nopython2
```
bnavigator, line 368: Why install nose in python2 then? Just for the import?

bnavigator:
GitHub workflow results look good. And so fast! :rocket: :smile:

control/tests/baseline_images/freqresp_test/gangof4-pvtol-expected.png
murrayrm: Symbolic link set up by pytest that apparently used an absolute path.  Fixed.

control/tests/freqresp_test.py
murrayrm: The problem I was having is that I need to import `image_comparison` from `matplotlib.testing.decorators` and this requires `nose` under python2 (no idea why).  And I think there was some problem with getting the comparison to work in Python2 (at some point), so I skipped the unit test.

We can probably clean it up, but since it might break Travis and since we are deprecating Python 2, I'm going to leave for now.

bnavigator: I hope we can get rid of Python 2 soon.

bnavigator:
- I don't think the symlinked file is needed at all.

- The comparison test fails on my machine:

expected:
gangof4-pvtol-expected
actual:
gangof4-pvtol
diff:
gangof4-pvtol-failed-diff

Comes from float data, needs float tolerance?

murrayrm:
It looks like bitmap comparison is not going to work here (the problem is that small shifts are causing large differences in the bitmap).  I'll have to write something that extracts individual elements and make sure they are in the right places.

bnavigator:
I am not sure if it is really worth the effort. My suggestion was just an idea.

Would it suffice to
- remove the axis text by the appropriate keyword of the tester,
- switching off the grid, and
- a large enough resolution, (which would not increase the file size too much because the png compression should work good enough)?

control/tests/freqresp_test.py
henklaak: Thanks for the effort, but... in my experience image comparisons with matplotlib outputs are very brittle test.
Whenever matplotlib decides to change their defaults (which they have done in the past), or a font is ever so slightly different, the test will fail. It's a never ending maintenance hassle.

Probably better to mock the matplotlib interface and only check for the right interface calls.
I'm willing to help.

henklaak: To help you on your way:

def test_gangof4_pvtol(pvtol_inner):
    with patch('control.freqplot.plt') as mockplt:
        ctrl.gangof4(*pvtol_inner)
        print(mockplt.method_calls)

bnavigator: Comparing with a fixed set of samples with values of arbitrary precision is equally brittle.

bnavigator: The issue to test here is not about the correct value of the curves but that the plot functions produce a non-crashing and legible plot.

henklaak: @bnavigator 
Respectfully disagree, we are not testing matplotlib, we must make sure that we send the right calls to matplotlib. 

And we wouldn't compare the calls with a hard coded array, but with the data from within the gangof4 function itself, using float compares with tolerances.

I'll support your decision.

bnavigator: Sorry, I still don't see a point in testing that the calls which are defined in the source code are the same calls which we retype again in the test code. There is nothing meaningful to **compare**.

murrayrm:
Getting rid of this (very old) PR.  There is now a script `examples/plot_gallery.py` that generates a set of "standard" plots that can be used to identify changes (by eye).  Probably OK for now.

PR #509, opened 2021-01-15
change default value of statesp.remove_useless_states to False

murrayrm:
This PR resolves issue #244.  It looks like the only issue was that converting a constant to a state space object was creating an extra state and this state had to then be removed.  By creating a state space system with no internal state (A = [[]] instead of A=[[0]]), this action is no longer required to get sensible results from block diagram algebra functions.

The PR makes that change and also changes the default value of `statesp.remove_useless_states` to `False`.

### DIFF
--- a/control/tests/statesp_test.py
+++ b/control/tests/statesp_test.py
@@ -395,9 +395,6 @@ class TestStateSpace:
         D0 = 0
         D1 = np.ones((2,1))
         assert StateSpace(A0, B0, C1, D1).is_static_gain()
-        # TODO: fix this once remove_useless_states is false by default
-        # should be False when remove_useless is false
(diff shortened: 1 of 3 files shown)


sawyerbfuller (approved):

control/tests/statesp_test.py
```diff
@@ -395,9 +395,6 @@ def test_is_static_gain(self):
         D0 = 0
         D1 = np.ones((2,1))
         assert StateSpace(A0, B0, C1, D1).is_static_gain()
-        # TODO: fix this once remove_useless_states is false by default
-        # should be False when remove_useless is false
-        # print(StateSpace(A1, B0, C1, D1).is_static_gain())
```
bnavigator, line 400: `assert not StateSpace(A1, B0, C1, D1).is_static_gain()` ?

PR #514, opened 2021-01-19
MIMO impulse and step response

murrayrm:
This PR implements MIMO impulse and step responses, mirroring the style and functionality of frequency response functions for MIMO systems.  It resolves issue #512 and is consistent with the discussion issue #453. Specifically:

* If a system is MIMO, calling `step_response` or `impulse_response` will generate an array of responses (indexed by output, input, and time).
* Consistent with frequency response methods, if a system is SISO then by default the output will just be indexed by time.
* Input and output axes that are single dimensional can be removed by using the `squeeze=True` keyword.
* SISO systems can be tried as MIMO systems (retaining the output and input axes in the response) using the `squeeze=False` keyword.

Note that the response for `initial_response` and `forced_response` is *not* indexed by the input (since there is no input in one case and a specifically applied input in the other).  The processing of the `squeeze` keyword is the same: by default SISO systems will be returned without the output axis, use `squeeze=True` to remove single dimensional axes from MIMO systems, use `squeeze=False` to force SISO systems to be treated like MIMO systems.

Other (small changes):

* Updated unit tests to make sure all of the squeezing rules are properly implemented with the new MIMO functionality.
* Found a bug in the way that the `transpose` keyword was working: it changed [output, input, time] to [time, input, output] instead of [time, output, input].  This was caught when adding unit tests.
* Found some errors in the MATLAB unit tests where the return arguments were in the wrong order (didn't show up with SISO only responses).

### DIFF
--- a/control/timeresp.py
+++ b/control/timeresp.py
@@ -500,10 +500,10 @@ def _process_time_response(
         config.defaults['control.squeeze_time_response'].
 
     input : int, optional
-        If present, the response represents ony the listed input.
+        If present, the response represents only the listed input.
 
     output : int, optional
-        If present, the response represents ony the listed input.
+        If present, the response represents only the listed output.
 
     Returns
     -------
--- a/control/timeresp.py
+++ b/control/timeresp.py
@@ -91,8 +91,7 @@ __all__ = ['forced_response', 'step_response', 'step_info', 'initial_response',
 # Helper function for checking array-like parameters
 def _check_convert_array(in_obj, legal_shapes, err_msg_start, squeeze=False,
                          transpose=False):
-    """
-    Helper function for checking array_like parameters.
+    """Helper function for checking array_like parameters.
 
     * Check type and shape of ``in_obj``.
     * Convert ``in_obj`` to an array if necessary.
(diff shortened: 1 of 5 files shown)


control/timeresp.py
```diff
@@ -456,36 +457,122 @@ def forced_response(sys, T=None, U=0., X0=0., transpose=False,
...
+        output even if the system is not SISO. If squeeze=False, keep the
+        output as a 3D array (indexed by the output, input, and time) even if
+        the system is SISO. The default value can be set using
+        config.defaults['control.squeeze_time_response'].
+
+    input : int, optional
+        If present, the response represents ony the listed input.
+
+    output : int, optional
+        If present, the response represents ony the listed input.
```
sawyerbfuller, line 506:
```suggestion
        If present, the response represents only the listed output.
```

sawyerbfuller (approved):
Looks good to me with one small doc change. 

also: not changed in this PR but maybe the right place to fix it: on line 20 in `config.py`  is a duplicated setting:
``'control.squeeze_time_response': True,``

Issue #493, opened 2020-12-31
sumblk doesnt exists for connection in matlab compatibility module


k10blogger:
From upgrading from 0.7.0 to 0.8.0 of python control library. I cannot find the sumblk function in matlab compatibility module in the version 0.8.0

Here is a link to documentation of previous release.
https://python-control.readthedocs.io/en/0.7.0/matlab_strings.html#system-interconnections

I cannot seem to find this in new release. 
I am also not able see the reason for this exclusion.

sawyerbfuller:
Hi siddarth, that feature did not have an asterisk (*) next to it, indicating that it wasn’t yet implemented. 

I agree it would be a very nice feature to have and have I have used it myself, created as follows: 

sumblk = control.ss([[0,0.],[0,0]], [[0,0.],[0,0]], [[0,0.],[0,0]], [[1,  0],[1,  1]])

murrayrm:
A few comments on this:

* A less "stateful" version of @sawyerbfuller's subblk function would be ```sumblk = control.ss([], [], [], [[1, 0], [1, 1]])``` which avoids the creation of any additional states.  Note that this creates an "unnamed" system, since `StateSpace` systems do not have signal names associated with that.

* For systems with named signals (`InputOutputSystem`'s), this functionality is also available in the `interconnect` function "for free" by simply interconnecting multiple signals to the input of another block (they are automatically summed together and can also have optional gains assigned to them).  For example, the following code will create a feedback loop (without creating an explicit summing block for creating of the error term):
```
P = ct.LinearIOSystem(ct.tf2ss(ct.tf([1], [1, 1])), name='P')
C = ct.LinearIOSystem(ct.tf2ss(ct.tf([2], [1, 0])), name='C')
S = ct.interconnect([P, C],
    connections = [ ['P.u[0]', 'C.y[0]'], ['C.u[0]', '-P.y[0]'] ],
    inplist=['C.u[0]'], outlist=['P.y[0]'])
```

It would be possible to create a function that looks more like the MATLAB `sumblk` function, but if the only usage is for `InputOutputSystem`'s, then I am not sure whether it is needed (?).

k10blogger:
For named system, usage of sumblk as below is something that i feel is better readable:
```
 S1.InputName = 'r'
 S1.OutputName = 'ug1'
 S2.InputName = 'e'
 S2.OutputName = 'ug2'
 S3.InputName = 'ym'
 S3.OutputName = 'y'
[...]

PR #517, opened 2021-01-23
add summing junction + implicit signal interconnection

murrayrm:
This PR addresses the enhancement describe in issue #516 where signals with the same name are automatically interconnected by the `interconnect` function.  It also defines a new function `summation_block` that addresses the request in issue #493.  Together, these capabilities allow the following code to work for setting up a simple feedback system:
```
P = ct.ss2io(
    ct.rss(2, 1, 1, strictly_proper=True),
    inputs='u', outputs='y', name='P')
kp = ct.tf(random.uniform(1, 10), [1])
ki = ct.tf(random.uniform(1, 10), [1, 0])
C = ct.tf2io(kp + ki, inputs='e', outputs='u', name='C')
sumblk = ct.summation_block(inputs=['r', '-y'], output='e', name="sum")
Tio_sum = ct.interconnect(
    (C, P, sumblk), inplist=['r'], outlist=['y'])
```

Other minor changes:
* Added code to allow the `iosys.feedback` function to accept a float_like for the feedback system (so that `feedback(P * C, 1)` now works.

### DIFF
--- a/control/iosys.py
+++ b/control/iosys.py
@@ -3,16 +3,11 @@
 # RMM, 28 April 2019
 #
 # Additional features to add
-#   * Improve support for signal names, specially in operator overloads
-#       - Figure out how to handle "nested" names (icsys.sys[1].x[1])
-#       - Use this to implement signal names for operators?
 #   * Allow constant inputs for MIMO input_output_response (w/out ones)
 #   * Add support for constants/matrices as part of operators (1 + P)
 #   * Add unit tests (and example?) for time-varying systems
 #   * Allow time vector for discrete time simulations to be multiples of dt
 #   * Check the way initial outputs for discrete time systems are handled
-#   * Rename 'connections' as 'conlist' to match 'inplist' and 'outlist'?
-#   * Allow signal summation in InterconnectedSystem diagrams (via new output?)
 #
 
 """The :mod:`~control.iosys` module contains the
@@ -44,7 +39,7 @@ from . import config
 __all__ = ['InputOutputSystem', 'LinearIOSystem', 'NonlinearIOSystem',
            'InterconnectedSystem', 'LinearICSystem', 'input_output_response',
            'find_eqpt', 'linearize', 'ss2io', 'tf2io', 'interconnect',
-           'summation_block']
+           'summing_junction']
 
 # Define module default parameter values
 _iosys_defaults = {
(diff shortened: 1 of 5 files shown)


sawyerbfuller:
Very cool! 

I particularly like that it appears possible with this to leave out the ‘name’ keyword for simple systems like this. That keyword is somewhat redundant and prone to errors when code gets modified. 

If you give the sumblock more input names (eg 3 instead of 2), does it sum those three signals?

murrayrm:
> it appears possible with this to leave out the ‘name’ keyword for the constitutive systems for simple interconnections like this. Am I correct in that?

Correct.  The following works just fine (from #516):
```
P = ct.tf2io(ct.tf(1, [1, 0]), inputs='u', outputs='y')
C = ct.tf2io(ct.tf(10, [1, 1]), inputs='e', outputs='u')
sumblk = ct.summation_block(inputs=['r', '-y'], output='e')
T = ct.interconnect((P, C, sumblk), inplist='r', outlist='y')
```
(`P`, `C`, and `sumblk` all have names automatically assigned, but not needed here).

> If you give the sumblock more input names (eg 3 instead of 2), does it sum those three signals?

Yup.  Any number of inputs can be summed together.  The `summation_block` function does not require named signals, so you can also do `ct.summation_block(7)` if you want a `LinearIOSystem` that has 7 inputs and 1 output (sum of the inputs), for example).

Some possible tweaks:
* I think I may use the term `summing_junction` (instead of `summation_block`) to line up with the term we use in FBS.
* I will add a bit more documentation, both in docstrings and in the manual
* I want to switch all examples using `InterconnectedSystem` to use `interconnect` since using the class constructor since the function provides a slightly higher level interface (eg, the class constructor still requires the explicit signal name).
*

sawyerbfuller:
Awesome. 

New name for summer sounds better. 

Hoping this will make creating simulink-like interconnected system simulations much easier. I will let you know how it goes once it’s been merged in.

bnavigator:
Very nice!

(Now we will have users demanding a graphical editor next :grin:. Implemented as jupyter notebook widget or the like.)

sawyerbfuller:
I think the first step in that direction might be to add an option in
interconnect to draw a block diagram. This could be done using matplotlibs
pretty good capabilities  for this.


-- 
Sawyer Fuller
Assistant Professor of Mechanical Engineering
Adjunct, Paul G Allen School of Computer Science
University of Washington
http://faculty.washington.edu/~minster/

(Typed with my thumbs)

control/iosys.py
```diff
@@ -1893,8 +1894,18 @@ def interconnect(syslist, connections=[], inplist=[], outlist=[],
         and the special form '-sys.sig' can be used to specify a signal with
         gain -1.
 
-        If omitted, the connection map (matrix) can be specified using the
-        :func:`~control.InterconnectedSystem.set_connect_map` method.
+        If omitted, all the `interconnect` function will attempt to create the
+        interconneciton map by connecting all signals with the same base names
+        (ignoring the system name).  Specifically, for each input signal name
+        in the list of systems, if that signal name corresponds to the output
+        signal in any of the systems, it will be connected to that input (with
```
sawyerbfuller, line 1901: Nice that this summation effect is stated explicitly.

sawyerbfuller (approved):

PR #518, opened 2021-01-24
Rory/minor fixes

roryyorke:
Some minor fixes, mostly to examples.

I guess the examples aren't run as part of CI?  examples/run_examples.sh was supposed to help with that.

### DIFF
--- a/control/tests/convert_test.py
+++ b/control/tests/convert_test.py
@@ -155,7 +155,7 @@ class TestConvert:
     def testConvertMIMO(self):
         """Test state space to transfer function conversion.
 
-        Do a MIMO conversation and make sure that it is processed
+        Do a MIMO conversion and make sure that it is processed
         correctly both with and without slycot
 
         Example from issue gh-120, jgoppert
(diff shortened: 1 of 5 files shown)


bnavigator:
> I guess the examples aren't run as part of CI? examples/run_examples.sh was supposed to help with that.

We should add an appropriate examples_test.py [like we do in Slycot](https://github.com/python-control/Slycot/blob/master/slycot/tests/test_examples.py)

murrayrm:
We have a script `examples/run_examples.py` that will run all examples and check for errors.  We just need to put that at the end of workflow.  Perhaps this can/should go in the `slycot-from-src` workflow just as an independent check that we haven't accidentally broken anything?

Changes all look fine.  I think fixed a few of these in some of my pending PR's (I remember a conversation → conversion conversion at some point), so we may get some small merge conflicts.

@roryyorke Do you want to add the `run_examples.py` check in this PR or hold for a different one (it is note quite related, but since you pointed it out...).

roryyorke:
I'm happy to do it, though I'll only get to it the weekend starting 30 Jan.

PR #521, opened 2021-01-28
Describing functions

murrayrm:
This PR adds a basic describing function capability, mainly captured in the new `describing_function_plot` command.  Almost all of the new code is in the `descfcn.py` file, with a few changes elsewhere to support some of the patterns.  There is a Jupyter notebook demonstrating the use of the code.

Additional changes:
* Added a `mirror` keyword to `nyquist_plot` that sets the line style used for plotting the complementary portion of the Nyquist plot.  This defaults to `--`, which matches the FBS convention.
* Added an `_isstatic` method to the `NonlinearIOSystem` class that allows passing a nonlinear I/O object as the argument for computing describing functions.
* Added a `__call__` method to the `NonlinearIOSystem` class that direct calling of a static nonlinear input/output function (returning the value of the function at the given input).


control/iosys.py
```diff
@@ -450,6 +450,10 @@ def issiso(self):
         """Check to see if a system is single input, single output"""
         return self.ninputs == 1 and self.noutputs == 1
 
+    def isstatic(self):
```
sawyerbfuller, line 453: statespace and xferfcn have methods `is_static_gain()` I added. should we choose a standard conventiuon? I would lean toward isstatic because it is shorter and matches `issiso()`,

murrayrm: I agree we should be uniform.  I'm not sure that `is_static_gain`() would make sense for a nonlinear system, so that is another reason to opt for `isstatic()`.  

Also: these are mainly internally used functions, so I think it is OK to stick with `isstatic`/`issiso` rather than `is_static`/`is_siso`.

murrayrm: Converted to `_isstatic` in latest push since this really is an internal function, not intended to be seen by the user.

@sawyerbfuller: when we update `is_static_gain`, we can either change to `_isstatic` if all usage is internal or change to `isstatic` (and update `descfcn.py` for consistency).

sawyerbfuller: renamed in #524

control/descfcn.py
```diff
@@ -0,0 +1,482 @@
...
+    sin_theta = np.sin(theta)
+    cos_theta = np.cos(theta)
+
+    # See if this is a static nonlinearity (assume not, just in case)
+    if not hasattr(F, '_isstatic') or not F._isstatic():
+        # Initialize any internal state by going through an initial cycle
+        [F(x) for x in np.atleast_1d(A).min() * sin_theta]
+
+    # Go through all of the amplitudes we were given
+    df = []
```
roryyorke, line 166: It would be better to preallocate `df` with `np.empty`; maybe
```
    retdf = np.empty(np.shape(A), dtype=complex)
    df = retdf
    df.shape = (-1,)
```
the for loop would be `for i, a in enumerate(np.atleast_1d(A)):`, assignments would be to `df[i]`, and the function would return `retdf`.

control/descfcn.py
```diff
@@ -0,0 +1,482 @@
...
+    # and then integrate the product against \sin\theta and \cos\theta to obtain
+    #
+    #   \int_0^{2\pi} F(a\sin\theta) \sin\theta d\theta = M_1 \pi \cos\phi
+    #   \int_0^{2\pi} F(a\sin\theta) \cos\theta d\theta = M_1 \pi \sin\phi
+    #
+    # From these we can compute M1 and \phi.
+    #
+
+    # Evaluate over a full range of angles
+    theta = np.linspace(0, 2*np.pi, num_points)
```
roryyorke, line 155: this is not something I'm not at all sure about: should this omit the final point, so it's more like a discrete Fourier transform?

control/descfcn.py
```diff
@@ -0,0 +1,482 @@
...
+                raise ValueError("function must evaluate to zero at zero")
+            df.append(1.)
+            continue
+        elif a < 0:
+            raise ValueError("cannot evaluate describing function for A < 0")
+
+        # Save the scaling factor for to make the formulas simpler
+        scale = dtheta / np.pi / a
+
+        # Evaluate the function (twice) along a sinusoid (for internal state)
```
roryyorke, line 181: I don't see how this is evaluating the function twice over the sinusoid (assume that means two periods of sinusoid).

control/descfcn.py
```diff
@@ -0,0 +1,482 @@
...
+    # evaluating the nonlinearity over a sinusoid.  The Fourier series for a
+    # static nonlinear function evaluated on a sinusoid can be written as
+    #
+    # F(A\sin\omega t) = \sum_{k=1}^\infty M_k(A) \sin(k\omega t + \phi_k(A))
+    #
+    # The describing function is given by the complex number
+    #
+    #    N(A) = M_1(A) e^{j \phi_1(A)} / A
+    #
+    # To compute this, we compute F(\theta) for \theta between 0 and 2 \pi,
```
roryyorke, line 139:
```suggestion
    # To compute this, we compute F(A\sin\theta) for \theta between 0 and 2 \pi,
```

control/descfcn.py
```diff
@@ -0,0 +1,482 @@
...
+    # use the identities
+    #
+    #   \sin(\theta + \phi) = \sin\theta \cos\phi + \cos\theta \sin\phi
+    #   \int_0^{2\pi} \sin^2 \theta d\theta = \pi
+    #   \int_0^{2\pi} \cos^2 \theta d\theta = \pi
+    #
+    # and then integrate the product against \sin\theta and \cos\theta to obtain
+    #
+    #   \int_0^{2\pi} F(a\sin\theta) \sin\theta d\theta = M_1 \pi \cos\phi
```
roryyorke, line 148:
```suggestion
    #   \int_0^{2\pi} F(A\sin\theta) \sin\theta d\theta = M_1 \pi \cos\phi
```

control/descfcn.py
```diff
@@ -0,0 +1,482 @@
...
+    #   \int_0^{2\pi} F(a\sin\theta) \cos\theta d\theta = M_1 \pi \sin\phi
```
roryyorke, line 149:
```suggestion
    #   \int_0^{2\pi} F(A\sin\theta) \cos\theta d\theta = M_1 \pi \sin\phi
```

roryyorke (commented):
I'll try to review from `describing_function_plot` on later this weekend.

roryyorke (commented):
Searching for the intersections will be O(m * n).  

We could look libraries we can pull in to find the intersection of H(jw) and N(A).  I googled "numpy polyline intersection" and found pointers to https://pypi.org/project/Shapely/, but that may be over the top.   

We could also look at using `scipy.optimize.root`, though I don't see an obvious way to bound the domain of search.  And we'd have to handle the multiple roots ourselves.

When N(A) is strictly real (saturation, deadzone, relay), we could probably specialize the solver.

control/descfcn.py
```diff
@@ -0,0 +1,482 @@
...
+class DescribingFunctionNonlinearity():
+    """Base class for nonlinear functions with a describing function
+
+    This class is intended to be used as a base class for nonlinear functions
+    that have a analytically defined describing function (accessed via the
+    :meth:`describing_function` method).  Objects using this class should also
+    implement a `call` method that evaluates the nonlinearity at a given point
+    and an `_isstatic` method that is `True` if the nonlinearity has no
+    internal state.
+
```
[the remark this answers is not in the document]
murrayrm: I've updated the documentation a bit to address the comments above.

In terms of calling `DescribingFunctionNonlinearity.description_function` with an array: the intent is that this is a low level function that just evaluates the describing function at a scalar value.  The `describing_function` function has the code to evaluate on an array.

control/descfcn.py
```diff
@@ -0,0 +1,482 @@
...
+        corresponding describing function values is returned.
+
+    Raises
+    ------
+    TypeError
+        If A < 0 or if A = 0 and the function F(0) is non-zero.
+
+    """
+    # If there is an analytical solution, trying using that first
+    if try_method and hasattr(F, 'describing_function'):
```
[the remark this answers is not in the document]
murrayrm: Good catch. Fixed.

control/descfcn.py
murrayrm: Good catch.  We are approximating the integral by a discrete sum with the assumption that the function is constant on each small interval.  But this means that we don't want to include the value at 2pi (since we already captured the segment just before 2pi in the previous evaluation).

control/descfcn.py
murrayrm: Left over from an earlier iteration (before I implemented `_isstatic`).  Fixed.

control/freqplot.py
```diff
@@ -508,7 +508,7 @@ def gen_zero_centered_series(val_min, val_max, period):
 
 def nyquist_plot(syslist, omega=None, plot=True, label_freq=0,
                  arrowhead_length=0.1, arrowhead_width=0.1,
-                 color=None, *args, **kwargs):
+                 mirror='--', color=None, *args, **kwargs):
```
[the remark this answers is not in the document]
murrayrm: Updated documentation.  Should either be a linestyle or False if you don't want to plot the mirror image at all.  Setting it to (explicitly) to None will generate an error (from matplotlib).

control/descfcn.py
```diff
@@ -0,0 +1,482 @@
...
+
+    def describing_function(self, A):
+        # Check to make sure the amplitude is positive
+        if A < 0:
+            raise ValueError("cannot evaluate describing function for A < 0")
+
+        if self.lb <= A and A <= self.ub:
+            return 1.
+        else:
+            alpha, beta = math.asin(self.ub/A), math.asin(-self.lb/A)
```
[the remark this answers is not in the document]
murrayrm: It's equivalent (I checked, but also the unit tests check this).

control/descfcn.py
```diff
@@ -0,0 +1,482 @@
...
+            return np.nan
+
+        df_real = 4 * self.b * math.sqrt(1 - (self.c/A)**2) / (A * math.pi)
+        df_imag = -4 * self.b * self.c / (math.pi * A**2)
+        return df_real + 1j * df_imag
+
+
+# Backlash nonlinearity (#48 in Gelb and Vander Velde, 1968)
+class backlash_nonlinearity(DescribingFunctionNonlinearity):
+    """Backlash nonlinearity for use in describing function analysis
```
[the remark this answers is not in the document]
murrayrm: I saw that, but couldn't figure out what the "friction-controlled" had to do with it exactly.  It seemed to me that you just had backlash as in a gear and so the output just followed the input (modulo the backlash action).

control/descfcn.py
```diff
@@ -0,0 +1,482 @@
...
+        super(backlash_nonlinearity, self).__init__()
+
+        self.b = b              # backlash distance
+        self.center = 0         # current center position
+
+    def __call__(self, x):
+        # Convert input to an array
+        x_array = np.array(x)
+
+        y = []
```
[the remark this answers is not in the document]
murrayrm: Good catch.

control/freqplot.py
```diff
@@ -606,11 +606,12 @@ def nyquist_plot(syslist, omega=None, plot=True, label_freq=0,
...
 
-                plt.plot(x, -y, '-', color=c, *args, **kwargs)
-                ax.arrow(
-                    x[-1], -y[-1], (x[-1]-x[-2])/2, (y[-1]-y[-2])/2,
-                    fc=c, ec=c, head_width=arrowhead_width,
-                    head_length=arrowhead_length)
+                if mirror is not False:
+                    plt.plot(x, -y, mirror, color=c, *args, **kwargs)
+                    ax.arrow(
+                        x[-1], -y[-1], (x[-1]-x[-2])/2, (y[-1]-y[-2])/2,
```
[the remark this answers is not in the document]
murrayrm: Not related to this PR.  We should generate a new issue to check this.

control/descfcn.py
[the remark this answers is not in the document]
roryyorke: Maybe it's not important, but I remember a colleague always talking about the two backlash types (friction-controlled and inertia-controlled), and I thought it was standard nomenclature.  The inertia-controlled case is discussed in the reference above immediately after friction-controlled.

control/freqplot.py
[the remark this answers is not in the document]
roryyorke: I meant the sign of the `dy` arg in the new `if mirror is not False` branch; you've change  `y[-1]` to `-y[-1]`, but haven't changed (y[-1]-y[-2])/2`.  I'd have thought if it were wrong it would be obvious on the figure, though.

control/descfcn.py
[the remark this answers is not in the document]
murrayrm: Sorry, I was looking at something different (probably just the table in Appendix B, which didn't have the explanation).  I'll update.

PR #522, opened 2021-01-28
Freqplot improvements

sawyerbfuller:
fixes to freqplot.py:
* empty glyph in margin plot
* evalfr(sys,s)->sys(s)
* changed to a reasonable number of frequency points (1000) in bode plots (up from a very coarse 50)
* unify frequency range specification between bode and nyquist 
edit:
* mag/phase data in bode plot is now plotted along with vertical nyquist freq line for DT systems so that it is convenient to call `legend(('sys1', 'sys2'))`
* unit tests for frequency ranges

### DIFF
--- a/control/bdalg.py
+++ b/control/bdalg.py
@@ -54,6 +54,7 @@ $Id$
 """
 
 import numpy as np
+from scipy.signal.ltisys import StateSpace
 from . import xferfcn as tf
 from . import statesp as ss
 from . import frdata as frd
@@ -280,7 +281,10 @@ def append(*sys):
     >>> sys = append(sys1, sys2)
 
     """
-    s1 = sys[0]
+    if not isinstance(sys[0], StateSpace):
+        s1 = ss._convert_to_statespace(sys[0])
+    else:
+        s1 = sys[0]
     for s in sys[1:]:
         s1 = s1.append(s)
     return s1
--- a/control/bdalg.py
+++ b/control/bdalg.py
@@ -54,7 +54,6 @@ $Id$
 """
 
 import numpy as np
-from scipy.signal.ltisys import StateSpace
 from . import xferfcn as tf
 from . import statesp as ss
 from . import frdata as frd
@@ -175,7 +174,7 @@ def negate(sys):
     >>> sys2 = negate(sys1) # Same as sys2 = -sys1.
 
     """
-    return -sys;
+    return -sys
 
 #! TODO: expand to allow sys2 default to work in MIMO case?
 def feedback(sys1, sys2=1, sign=-1):
@@ -281,10 +280,7 @@ def append(*sys):
     >>> sys = append(sys1, sys2)
 
     """
-    if not isinstance(sys[0], StateSpace):
-        s1 = ss._convert_to_statespace(sys[0])
-    else:
-        s1 = sys[0]
+    s1 = ss._convert_to_statespace(sys[0])
     for s in sys[1:]:
         s1 = s1.append(s)
     return s1
--- a/control/freqplot.py
+++ b/control/freqplot.py
@@ -220,18 +220,17 @@ def bode_plot(syslist, omega=None,
             raise ControlMIMONotImplemented(
                 "Bode is currently only implemented for SISO systems.")
         else:
-            omega_sys = np.array(omega)
-            if sys.isdtime(True):
+            omega_sys = np.asarray(omega)
+            if sys.isdtime(strict=True):
                 nyquistfrq = 2. * math.pi * 1. / sys.dt / 2.
                 omega_sys = omega_sys[omega_sys < nyquistfrq]
                 # TODO: What distance to the Nyquist frequency is appropriate?
             else:
                 nyquistfrq = None
 
-            # Get the magnitude and phase of the system
-            mag_tmp, phase_tmp, omega_sys = sys.frequency_response(omega_sys)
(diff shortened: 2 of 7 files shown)


sawyerbfuller:
added a commit because changes were not necessary and to avoid a merge conflict with #521

control/freqplot.py
```diff
@@ -518,29 +518,36 @@ def nyquist_plot(syslist, omega=None, plot=True, label_freq=0,
...
         List of linear input/output systems (single system is OK)
-    omega : freq_range
-        Range of frequencies (list or bounds) in rad/sec
-    Plot : boolean
+    plot : boolean
         If True, plot magnitude
+    omega : array_like
+        Range of frequencies in rad/sec
+    omega_limits : array_like of two values
+        Limits of the to generate frequency vector.
```
sawyerbfuller, line 526:
```suggestion
        Set of frequencies to be evaluated in rad/sec.
```

control/freqplot.py
```diff
@@ -518,29 +518,36 @@ def nyquist_plot(syslist, omega=None, plot=True, label_freq=0,
     ----------
     syslist : list of LTI
         List of linear input/output systems (single system is OK)
-    omega : freq_range
-        Range of frequencies (list or bounds) in rad/sec
-    Plot : boolean
+    plot : boolean
         If True, plot magnitude
+    omega : array_like
+        Range of frequencies in rad/sec
```
sawyerbfuller, line 524:
```suggestion
        Set of frequencies to be evaluated in rad/sec.
```

control/tests/sisotool_test.py
```diff
@@ -78,8 +78,8 @@ def test_sisotool(self, sys):
 
         # Check if the bode_mag line has moved
         bode_mag_moved = np.array(
-            [111.83321224, 92.29238035, 76.02822315, 62.46884113, 51.14108703,
-             41.6554004, 33.69409534, 27.00237344, 21.38086717, 16.67791585])
+            [674.0242, 667.8354, 661.7033, 655.6275, 649.6074, 643.6426,
```
sawyerbfuller, line 81: Had to Change this because new number of sampling points.

control/bdalg.py
```diff
@@ -54,6 +54,7 @@
 """
 
 import numpy as np
+from scipy.signal.ltisys import StateSpace
```
murrayrm, line 57: Shouldn't this be importing control.StateSpace?

control/bdalg.py
```diff
@@ -280,7 +281,10 @@ def append(*sys):
     >>> sys = append(sys1, sys2)
 
     """
-    s1 = sys[0]
+    if not isinstance(sys[0], StateSpace):
```
murrayrm, line 284: With the current code at the top of the file, this is checking to see if `sys[0]` is a `sp.signal.ltisys.StateSpace` object, not a `control.StateSpace` object.  It should still work, but basically is almost always getting called (?).

control/bdalg.py
```diff
@@ -280,7 +281,10 @@ def append(*sys):
...
+        s1 = ss._convert_to_statespace(sys[0])
```
murrayrm, line 285: FYI, this function already checks if the argument is a `StateSpace` object and just returns if so => you can replace lines 284 to 287 with `s1 = ss._convert_to_statespace(sys[0])`.

control/freqplot.py
```diff
@@ -565,55 +572,61 @@ def nyquist_plot(syslist, omega=None, plot=True, label_freq=0,
...
-                            num=50, endpoint=True, base=10.0)
-
-    for sys in syslist:
-        if sys.ninputs > 1 or sys.noutputs > 1:
-            # TODO: Add MIMO nyquist plots.
-            raise NotImplementedError(
-                "Nyquist is currently only implemented for SISO systems.")
+        if omega_limits is None:
+            # Select a default range if none is provided
+            omega = default_frequency_range(syslist, Hz=False,
```
murrayrm, line 582: Should this be an internal function `_default_frequency_range`?

control/freqplot.py
```diff
@@ -565,55 +572,61 @@ def nyquist_plot(syslist, omega=None, plot=True, label_freq=0,
...
-            # Compute the primary curve
-            x = np.multiply(mag, np.cos(phase))
-            y = np.multiply(mag, np.sin(phase))
+            omega_limits = np.asarray(omega_limits)
+            if len(omega_limits) != 2:
+                raise ValueError("len(omega_limits) must be 2")
+            if omega_num:
+                num = omega_num
+            else:
+                num = config.defaults['freqplot.number_of_samples']
```
murrayrm, line 591: Another way to do this is
```
omega_num = ct.config._get_param('freqplot', 'number_of_samples', omega_num)
```

control/bdalg.py
sawyerbfuller: good point. fixed in next commit.

control/bdalg.py
sawyerbfuller: hmm, yes, not sre how that got in there. removed.

control/freqplot.py
sawyerbfuller: probably. its counter part, `_default_time_vector` is. I'll change it.

control/freqplot.py
sawyerbfuller: done.

murrayrm (approved):

PR #531, opened 2021-02-04
sisotool small visual cleanup, new feature to show step response of different input-output than loop

sawyerbfuller:
This PR adds a new feature to sisotool that allows it to show the step response of a system that is not the loop transfer function L. Some visual cleanup including reducing oversized axis tick labels. 

* To see a the step response of a more complicated system that includes a feedforward controller, or any other input-output pair, you can do so by passing sisotool a 2 input, 2-output system. The gain you choose when you click the rlocus plot gets inserted between output[0] and input[0], and then the step response is between input[1] and output[1]. 
* Doubled the click threshold size in rlocus because I had a hard time aiming the mouse well enough to get successful click events.

### DIFF
--- a/control/tests/sisotool_test.py
+++ b/control/tests/sisotool_test.py
@@ -22,6 +22,11 @@ class TestSisotool:
         return TransferFunction([1000], [1, 25, 100, 0])
 
     @pytest.fixture
+    def sysdt(self):
+        """Return a generic SISO transfer function"""
+        return TransferFunction([1000], [1, 25, 100, 0], True)
+
+    @pytest.fixture
     def sys222(self):
         """2-states square system (2 inputs x 2 outputs)"""
         A222 = [[4., 1.],
@@ -45,7 +50,7 @@ class TestSisotool:
         D221 = [[1., -1.]]
         return StateSpace(A222, B222, C221, D221)
 
-    def test_sisotool(self, sys, sys222, sys221):
+    def test_sisotool(self, sys, sysdt, sys222, sys221):
         sisotool(sys, Hz=False)
         fig = plt.gcf()
         ax_mag, ax_rlocus, ax_phase, ax_step = fig.axes[:4]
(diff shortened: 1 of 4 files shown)


bnavigator:
:+1: 

* Pictures please?
* Please cover the 2-in, 2-out case and the MIMO Exception with tests: https://coveralls.io/builds/36879506/source?filename=control%2Fsisotool.py#L162

sawyerbfuller:
Here's a current screenshot: 

image

bnavigator:
Nice. Let's go all the way and cover these two lines as well:

https://coveralls.io/builds/36884037/source?filename=control%2Fsisotool.py#L170
https://coveralls.io/builds/36884037/source?filename=control%2Fsisotool.py#L172

control/tests/sisotool_test.py
```diff
@@ -88,7 +119,21 @@ def test_sisotool(self, sys):
...
+        # test discrete-time
+        sisotool(sysdt, tvect=5)
+
+        # test MIMO compatibility
+        # sys must be siso or 2 input, 2 output
+        with pytest.raises(ControlMIMONotImplemented):
+            sisotool(sys221)
+        # does not raise an error:
+        sisotool(sys222)
+
```
bnavigator, line 137: I would move these to separate tests.

control/tests/sisotool_test.py
```diff
@@ -88,7 +119,21 @@ def test_sisotool(self, sys):
         step_response_moved = np.array(
             [0., 0.0072, 0.0516, 0.1554, 0.3281, 0.5681, 0.8646, 1.1987,
              1.5452, 1.875])
-        # old: array([0., 0.0239, 0.161 , 0.4547, 0.8903, 1.407,
-        #             1.9121, 2.2989, 2.4686, 2.353])
         assert_array_almost_equal(
             ax_step.lines[0].get_data()[1][:10], step_response_moved, 4)
+
+        # test supply tvect
+        sisotool(sys, tvect=np.arange(0, 1, .1))
```
bnavigator, line 126: Somehow this still does not cover https://coveralls.io/builds/36911686/source?filename=control%2Fsisotool.py#L168

control/tests/sisotool_test.py
sawyerbfuller: sure, thanks for the suggestions. still getting the hang of best practices for testing

control/tests/sisotool_test.py
sawyerbfuller: Let's see if that works

PR #534, opened 2021-02-06
Nyquist plot improvements

murrayrm:
This PR provides improvements to Nyquist plots:

* The `nyquist_plot` command now returns the number of encirclements of the -1 point and (optionally) the contour used to generate the plot instead of the magnitude, phase, and frequency values.  (The system response can be generated from the contour by simply evaluating the LTI system along the contour).
* Nyquist plots now automatically create a small indentation around poles on the imaginary access, along Nyquist plots to be closed curves.  Options are available to change the size of the indentation (`indent_radius`) and the direction of the indentation for the case of poles exactly on the imaginary axis (set `indent_direction` to 'left' or 'right').  Indentation can be turned off by setting `indent_direction` to `none`).
* Arrows are now generated using `FancyArrowPatch`, which provides a fixed size and aspect ratio in display coordinates (so zooming in doesn't not change the size and shape of arrows).  The default arrow style can be overridden using the option `arrow_size` parameter or the `arrow_style` parameters.
* The number (or specific location) of arrows can be specified using the `arrows` keyword, which allows you to specify the number of arrows or the specific locations (via a list of relative positions on the primary segment).
* The line style for mirror image of the primary segment can be set using the `mirror_style` parameter and defaults to a dashed line.  The mirror image segment can be omitted by setting `mirror_style=False`.
* The MATLAB version of `nyquist` retains its original call signature and return values.

Before:
 

After:
 

Other notes:
* The `mirror_style` parameter will conflict with the `mirror` parameter in PR #521 (describing functions).  The functionality is basically the same, but there will be a merge conflict in whichever gets merged second.  (I think we should use `mirror_style`).
* The various default settings can now be set using `config.defaults` with module name `nyquist`.
* The `control/tests/nyquist_test.py` unit test file serves as a pretty complete unit test, but can also be run in python/ipython to generate plots that you can see (and make sure they look OK).
* I pulled the code for processing arguments for the MATLAB version of `bode` into a separate function (`_parse_freqplot_args`) that is used for both `bode` and `nyquist`.
* Added unit tests, docstrings, some PEP8 cleanup.


control/freqplot.py
```diff
@@ -520,85 +524,186 @@ def gen_zero_centered_series(val_min, val_max, period):
...
+        'nyquist', 'arrow_size', kwargs, _nyquist_defaults, pop=True)
+    arrow_style = config._get_param('nyquist', 'arrow_style', kwargs, None)
+    indent_radius = config._get_param(
+        'nyquist', 'indent_radius', kwargs, _nyquist_defaults, pop=True)
+    indent_direction = config._get_param(
+        'nyquist', 'indent_direction', kwargs, _nyquist_defaults, pop=True)
+
     # If argument was a singleton, turn it into a list
-    if not hasattr(syslist, '__iter__'):
+    isscalar = not hasattr(syslist, '__iter__')
```
sawyerbfuller, line 691: IsScalar sounds like it could be referring to the system. ‘Is_Single_system’?

control/matlab/wrappers.py
```diff
@@ -53,19 +56,71 @@ def bode(*args, **keywords):
...
+    imag : ndarray (or list of ndarray if len(syslist) > 1))
+        imaginary part of the frequency response array
+    omega : ndarray (or list of ndarray if len(syslist) > 1))
+        frequencies in rad/s
+
+    """
+    from ..freqplot import nyquist_plot
+
+    # If first argument is a list, assume python-control calling format
+    if (getattr(args[0], '__iter__', False)):
```
sawyerbfuller, line 100: I have seen this style elsewhere in the library so I am curious: isn’t hasattr(args[0], ‘__iter__’) both more readable and compact?

sawyerbfuller (approved):

sawyerbfuller:
Very nice! Plots look great and I think the new return values make sense.

control/matlab/wrappers.py
bnavigator: `hasattr` also returns True if `args[0]` is an iterable sequence, but empty. This code does only branch if `args[0]` is not iterable or the sequence does not cast to False (is empty).

control/tests/nyquist_test.py
```diff
@@ -0,0 +1,292 @@
...
+# Utility function for counting unstable poles of closed loop (Z in FBS)
+def _Z(sys):
+    return (sys.feedback().pole().real >= 0).sum()
+
+# Decorator to close figures when done with test (to avoid matplotlib warning)
+@pytest.fixture(scope="function")
+def figure_cleanup():
+    plt.close('all')
+    yield
+    plt.close('all')
```
bnavigator, line 41: Did you try to use the existing `mplcleanup` (automatically loaded from `conftest.py`)?
https://github.com/python-control/python-control/blob/d75c395083d38244fcb1ebb568c0ef61860284ad/control/tests/conftest.py#L111-L124

Now that we don't test python2 anymore, we could directly map it to [matplotlib.testing.decorators.cleanup](https://matplotlib.org/3.3.3/api/testing_api.html#matplotlib.testing.decorators.cleanup)

control/matlab/wrappers.py
sawyerbfuller: I see, thanks!

murrayrm: `getattr([], '__iter__', False)` returns ` <method-wrapper '__iter__' of list object ...`.  So looks like both of them would follow the branch?

bnavigator:
Hmm, I wonder why the nichols_test xlim warning is back here. The test is run **before** nyquist test and already uses the mplcleanup fixture.

Issue #533, opened 2021-02-06
Root locus timeout on Windows: no SIGALRM


roryyorke:
See https://github.com/python-control/Slycot/runs/1844313808?check_suite_focus=true#logs

`SIGALRM` is Unix only: `https://docs.python.org/3/library/signal.html#signal.SIGALRM`

murrayrm:
I'll put in a decorator to filter out this test if SIGALRM isn't available.

PR #535, opened 2021-02-06
skip timeout test if SIGALRM is not defined

murrayrm:
Quick fix for issue #533: if SIGALRM is not defined, don't bother with the test.


bnavigator:
I propose two alternatives: murrayrm/python-control#5 or #537

murrayrm:
Closing in favor of #537.

Issue #532, opened 2021-02-06
Inconsistent evaluation of tf/ss systems with pole at the origin


murrayrm:
Systems that have a pole at the origin return inconsistent values:
```
import control as ct
sys_tf = ct.tf([1], [1, 0])
sys_ss = ct.tf2ss(sys_tf)

sys_tf(0)        # returns (inf+0j) with divide by zero warning
sys_tf(0j)       # returns (inf+nanj) with divide by zero warning
sys_ss(0)        # LinAlgError: Singular matrix
sys_tf.dcgain()  # returns array(inf)
sys_ss.dcgain()  # returns array(nan)
```
I think all of these should return `inf` (or `array(inf)`).

If you have a pole and zero at the origin, you get `nan` or 1:
```
sys_tf = ct.tf([1, 0], [1, 0])
sys_ss = ct.tf2ss(sys_tf)

sys_tf(0)        # returns (nan+0j) with invalid value warning
sys_ss(0)        # returns (1+0j) [state space system is empty]
sys_tf.dcgain()  # returns array(nan)
sys_ss.dcgain()  # returns (1+0j) [state space system is empty]
```
I think all of these cases should produce `nan`.

I have a partial fix in a branch, but want to see what people think the "right" answer is before committing it.

roryyorke:
Would this apply to evaluating a system at any s-value where it has a pole?

For the second set of examples, for TFs one objection is that `\lim_{s->0} (s/s) = 1`.  For a Bode or Nyquist plot, for example, wouldn't one want this limit, and not a NaN ?

For state-space objects, why would you want a static gain to evaluate to NaN?

Separarately for state-space, the MIMO case might be tricky: consider a diagonal system with g_11 = 1/(s+1), g_22 = 1/s; would evaluating this at s = 0 be expected to give [[1, 0], [0, inf]] ?  I wonder how easy this will be to arrange:

```
n [37]: np.eye(2) @ np.diag([1, np.inf]) @ np.eye(2)
<ipython-input-37-7a3bfddc9905>:1: RuntimeWarning: invalid value encountered in matmul
  np.eye(2) @ np.diag([1, np.inf]) @ np.eye(2)
Out[37]: 
array([[nan, nan],
       [nan, inf]])
```

bnavigator:
+1 for consistency. `nan` looks like a good choice to me.

[...]

PR #542, opened 2021-02-11
Fix inconsistencies in frequency response at poles

murrayrm:
This PR addresses issue #532 regarding inconsistencies in the return values for various frequency response functions.  The changes fixes a number of previous inconsistencies, where state space and transfer function representations could return different values and evaluating the frequency response at the "origin" gave different values depending on whether you evaluated at 0 or 0j.

The changes here only affect results when evaluating the system at a pole of the system.

In this PR:
* State space systems and transfer functions return the same values when evaluated at poles (including poles at the origin):
  * Evaluation at 0 with pole at the origin: returns (inf + nanj)
  * Evaluation at non-origin pole with nonzero numerator: returns (inf + infj)
  * Evaluation at pole with cancellation by a zero: returns (nan + nanj)
* State space systems and transfer functions interpret argument as a complex number, so that G(0) and G(0j) return the same value.
* Computing the DC gain is (exactly) the same as evaluating at 0

A few other small changes:
* Added a `warn_infinity` flag to all functions used to evaluate frequency response.  By default this is set to `True`, in which case you get a divide by zero warning or singular matrix warning when evaluating at a pole.  Setting it to `False` suppresses this warning.  (Also, `dcgain` sets `warn_infinity` to False by default).
* Updated docstrings and unit tests

### DIFF
--- a/control/lti.py
+++ b/control/lti.py
@@ -644,8 +644,9 @@ def dcgain(sys):
     Returns
     -------
     gain : ndarray
-        The zero-frequency gain, or np.nan if the system has a pole
-        at the origin
+        The zero-frequency gain, or np.inf if the system has a pole at the
+        origin, np.nan if there is a pole/zero cancellation at the origin.
+
     """
     return sys.dcgain()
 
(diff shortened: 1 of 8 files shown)


sawyerbfuller (approved):
looks good to me! Nice to have these corner cases doing the right thing

PR #547, opened 2021-02-20
Small documentation updates

murrayrm:
There were some small glitches in docstrings that were causing errors in sphinx and also a few older changes to `margin` docstrings that were hanging around in an unnoticed branch.

No changes in code.


sawyerbfuller (approved):
looks good to me

PR #549, opened 2021-02-21
Optimal control module

murrayrm:
This PR adds a new optimal control module, `control.optimal`, that implements finite horizon optimal control problems with constraints, including rudimentary model predictive control (MPC).  The underlying algorithms are not super-efficient, so this is more of a "reference implementation" than something that you could use on a large problem, but it does allow you to explore ideas around tradeoffs in various types of cost functions, constraints, and other concepts.  The PR includes unit tests and documentation on the use of the module.

(The motivation for this PR is that I'll be teaching an optimal control class next year, and I'd like to have some tools around that students can use to get a feel for the concepts.  The hope is to implement most of the concepts that are in my (very incomplete) notes on "[Optimization-Based Control](https://fbswiki.org/wiki/index.php/Supplement:_Optimization-Based_Control)".)

A few notes (for feedback):

* I have called the module `control.optimal` and it is _not_ loaded by default.  So, like the `scipy.optimize` package, you have to load the module separately if you want to use it (`control.flatsys` is also like this, so there is a precedent).  I'm initially called the module `obc` (for optimization-based control`, but decided that `optimal` was probably better (and matched `optimize`, used in SciPy).
* The module is basically just a wrapper around the `scipy.optimize.minimize` function: it essentially takes the element of an optimal control problem and creates an optimization problem for SciPy to solve.  For this reason I have tried to make things consistent with `scipy.optimize` when possible (eg, the form of constraints, the way results are returned).
* The unit tests are a bit finicky and seem to depend on what system they are tested on.  Everything is working in GitHub Actions, but you'll see some conditional checks in the code for things that work differently there versus my local machine.  For this reason, it would be useful if people can try things out on other platforms and let me know what happens.
* Finally, as noted already above, the code is not super efficient, particularly for continuous time systems.  You'll see this if you run the `examples/steering-optimal.py` script, which takes about 30 seconds to do solve some pretty straightforward problems.  <s>It also takes about 15 seconds for `control/tests/optimal_test.py` to run on my Mac, which can get annoying (a full `pytest` run takes about a 45 seconds, so this is 30% for just one module).</s> 

Other changes along the way:
* Added in some functionality to check for unrecognized keywords in the `config.py` parsing function `_get_param`.
* 28 Feb 2021: Added a new 'bezier' basis function in `flatsys`, though only a partial implementation (needed for examples of optimizing over a set of basis functions).
* 28 Feb 2021: Added a `benchmark` directory with some airspeed velocity (asv) benchmarks for optimal control.  These are mainly for development purposes, but might be something we use more generally at a future date.

Comments and advice welcome!


sawyerbfuller:
This is super cool. Having these features conveniently in the library (rather than having to roll your own optimal controller) seems like it could facilitate incorporating these more advanced controls concepts into the digital controls class I teach, whcih is something I've been wanting to do. 

I probably won't be able to try your version before it's merged with the master branch, but I'll let you know if I encounter any issues once it is and I've gotten a chance to try it.

Issue #337, opened 2019-08-21
Any plan for step_info to support negative step responses?


hgducharme:
I have step responses that go negative and sometimes stay negative, and it breaks the settling time calculation and rise time calculation. I've been messing around with it and might open a pull request if this is something y'all would be interested in. 

For example, here is my code, the step response plot, and the error

```python
from control import tf, step_response, step_info
import matplotlib.pyplot as plt
import numpy as np

transfer_function = tf( [3.32, 0, -162.8], [1, 24.56, 186.5, 457.8, 116.2] )
time, response = step_response(transfer_function)

fig, ax = plt.subplots(1, 1)
ax.plot(time, response)
plt.show()

step_info(transfer_function)
```

![step_response](https://user-images.githubusercontent.com/8628581/63395443-5ecdeb00-c389-11e9-963a-24804b81aada.png)

`step_info()` gives the following error

```
Traceback (most recent call last):
  File "bug_test.py", line 14, in <module>
    step_info(transfer_function)
  File "/Users/hgducharme/Programming/pendulum/env/lib/python3.7/site-packages/control/timeresp.py", line 536, in step_info
    SettlingTime = T[i + 1]
IndexError: index 1000 is out of bounds for axis 0 with size 1000
```

When I run this code in MATLAB, I get the following result

<br>

<img width="644" alt="Screen Shot 2019-08-20 at 8 36 44 PM" src="https://user-images.githubusercontent.com/8628581/63395721-427e7e00-c38a-11e9-8ca3-25cd7bb7ce7d.png">

murrayrm:
@joaoantoniocardoso Any interest in taking a look at this?

joaoantoniocardoso:
@murrayrm Yes, I think I can take a look at this tomorrow. 

Also I think this error is not something that happens only for negative step responses... Months ago I did a pull-request that fixes some other cases with the same message.

juanodecc:
Hi, think I solved this problem and the problem of systems with non settling time, I mean when dcgain isinf or isnan, doing:

```
    # Steady state value
    InfValue = sys1.dcgain()
    if InfValue < 0.0:
        yout = -yout
        InfValue = np.abs(InfValue)




    # SettlingTime
    sup_margin = (1. + SettlingTimeThreshold) * np.abs(InfValue)
    inf_margin = (1. - SettlingTimeThreshold) * np.abs(InfValue)
    # find Steady State looking for the first point out of specified limits
[...]

PR #555, opened 2021-03-04
fixed bug in calculation of percent overshoot 

sawyerbfuller:
Now it gives the correct result even if the transfer function is not strictly proper.

Other changes: 
* step_info now calculates overshoot with respect to dcgain rather than the final (edit: numeric) value of the (computed) step response.
* adjusted the length of the step response to go for a little longer, to within 0.1% rather than 1% of steady-state, to improve precision of `step_info` (and updated unit tests)

a remark:  `sys.dcgain()` now always returns a complex number because it evaluates at a frequency of zero (1 for DT systems) (as a result of #542?). This returns a real-valued complex number. I think there is a reasonable argument that it should be cast to a `float` because currently we assume system parameters are real-valued.

### DIFF
--- a/control/tests/sisotool_test.py
+++ b/control/tests/sisotool_test.py
@@ -68,8 +68,8 @@ class TestSisotool:
 
         # Check the step response before moving the point
         step_response_original = np.array(
-            [0.    , 0.0366, 0.2032, 0.4857, 0.82  , 1.1358, 1.3762, 1.507 ,
-             1.5206, 1.4326])
+            [0.    , 0.021 , 0.124 , 0.3146, 0.5653, 0.8385, 1.0969, 1.3095,
+             1.4549, 1.5231])
         assert_array_almost_equal(
             ax_step.lines[0].get_data()[1][:10], step_response_original, 4)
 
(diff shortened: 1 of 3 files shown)


murrayrm:
Is #337 also addressed, by chance?

sawyerbfuller:
sorry, didn't have a chance to get to that one on this PR

bnavigator:
LGTM

juanodecc:
Dear @sawyerbfuller, the solution could have problem with systems with initial condition different of zero. I want to contribute , but at the moment, I don't now how test the last version.

I maked this example, to explain:


```
import numpy as np
import control as ctrl

A = [[0,1,0],[0,0,1],[-1,-2,-3]]
B = [[0],[0],[1]]
C = [1,0,0]
D = 0

sys=ctrl.ss(A,B,C,D)

t= np.linspace(0,20,1000)
t1,y1 =ctrl.step_response(sys,t,[-1,0,2])

plt.plot(t1,y1)
ctrl.step_info(sys)

new_overshoot = 100. * (y1.max() - sys.dcgain()) / sys.dcgain()
print('new_overshoot =', new_overshoot)
```
{'RiseTime': 2.616832783582626,
 'SettlingTime': 9.555404558233526,
 'SettlingMin': 0.8926922608222024,
 'SettlingMax': 1.1446150537879327,
 **'Overshoot': 15.77739060863902,**
 'Undershoot': 0.0,
 'Peak': 1.1446150537879327,
 'PeakTime': 6.06629418012336,
 'SteadyStateValue': 0.9886343506022361}

**new_overshoot = 38.34913836051592**

bnavigator:
But `ctrl.step_info(sys)` does not have any initial conditions. It does know nothing of the `step_response()` call.

PR #556, opened 2021-03-04
Matched plurality in the documentation.

dapperfu:
Code variable and 'return's documentation both indicate multiple poles can be returned.


sawyerbfuller:
Good catch, thanks for the contribution!

PR #558, opened 2021-03-11
new dynamics and output functions for statespace and iosystems

sawyerbfuller:
revised version of #546. 
* makes `t` argument required following discussion there. not sure yet that this is the best, but I was almost there already so I just submitted this PR. 
* benchmark results in the discussion of that PR that showed that creating a separate callable that skips argument processing had only a small performance advantage, so I decided to leave `dynamics` as a method.  
* leaves code and usage of `_rhs` and `_out` unchanged in `iosys`. The idea is to leave them as fast internal methods and to have `dynamics` and `output` serve as more user-friendly versions where speed is of less importance.

### DIFF
--- a/control/iosys.py
+++ b/control/iosys.py
@@ -756,8 +756,17 @@ class LinearIOSystem(InputOutputSystem, StateSpace):
         if params and warning:
             warn("Parameters passed to LinearIOSystems are ignored.")
 
-    _rhs = StateSpace.dynamics
-    _out = StateSpace.output
+    def _rhs(self, t, x, u):
+        # Convert input to column vector and then change output to 1D array
+        xdot = np.dot(self.A, np.reshape(x, (-1, 1))) \
+            + np.dot(self.B, np.reshape(u, (-1, 1)))
+        return np.array(xdot).reshape((-1,))
+
+    def _out(self, t, x, u):
+        # Convert input to column vector and then change output to 1D array
+        y = np.dot(self.C, np.reshape(x, (-1, 1))) \
+            + np.dot(self.D, np.reshape(u, (-1, 1)))
+        return np.array(y).reshape((-1,))
 
 class NonlinearIOSystem(InputOutputSystem):
     """Nonlinear I/O system.
--- a/control/iosys.py
+++ b/control/iosys.py
@@ -370,7 +370,7 @@ class InputOutputSystem(object):
         NotImplemented("Evaluation not implemented for system of type ",
                        type(self))
 
-    def dynamics(self, t, x, u, params={}):
+    def dynamics(self, t, x, u):
         """Compute the dynamics of a differential or difference equation.
 
         Given time `t`, input `u` and state `x`, returns the value of the
@@ -401,7 +401,7 @@ class InputOutputSystem(object):
         -------
         dx/dt or x[t+dt] : ndarray
         """
-        return self._rhs(t, x, u, params)
+        return self._rhs(t, x, u)
 
     def _out(self, t, x, u, params={}):
         """Evaluate the output of a system at a given state, input, and time
@@ -415,7 +415,7 @@ class InputOutputSystem(object):
         # If no output function was defined in subclass, return state
         return x
 
-    def output(self, t, x, u, params={}):
+    def output(self, t, x, u):
         """Compute the output of the system
 
         Given time `t`, input `u` and state `x`, returns the output of the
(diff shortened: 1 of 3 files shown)


control/iosys.py
```diff
@@ -358,26 +358,88 @@ def _update_params(self, params, warning=False):
...
+
+        Parameters
+        ----------
+        t : float
+            the time at which to evaluate
+        x : array_like
+            current state
+        u : array_like
+            input
+
```
murrayrm, line 399: Optional keyword `params` is not defined in docstring.

control/iosys.py
```diff
@@ -358,26 +358,88 @@ def _update_params(self, params, warning=False):
...
```
murrayrm, line 436: `params` is not documented.

control/iosys.py
```diff
@@ -706,7 +768,6 @@ def _out(self, t, x, u):
             + np.dot(self.D, np.reshape(u, (-1, 1)))
         return np.array(y).reshape((-1,))
 
-
```
murrayrm, line 709: I think this line removal will generate a PEP8 error (check with pycodestyle).  Two blank lines are required before function and class definitions.

control/statesp.py
```diff
@@ -1227,12 +1227,110 @@ def dcgain(self, warn_infinite=False):
...
+            raise ValueError("len(x) must be equal to number of states")
+
+        if len(args) == 2: # received t and x, ignore t
+            return self.A.dot(x).reshape((-1,)) # return as row vector
+        else: # received t, x, and u, ignore t
+            u = np.reshape(args[2], (-1, 1)) # force to a column in case matrix
+            if np.size(u) != self.ninputs:
+                raise ValueError("len(u) must be equal to number of inputs")
+            return self.A.dot(x).reshape((-1,)) \
+                 + self.B.dot(u).reshape((-1,)) # return as row vector
```
murrayrm, line 1278: Is there a reason to do it this way compared to defining the call signature as `dynamics(self, t, x, u=0)` and then having the function be something like
```
return self.A @ x if u == 0 else self.A @ x + self.B @ u
```
(perhaps with some dimension checking for user friendliness).

control/statesp.py
```diff
@@ -1227,12 +1227,110 @@ def dcgain(self, warn_infinite=False):
...
+
+    def output(self, *args):
```
murrayrm, line 1280: Same comment as above regarding call signature. 

If you keep this call signature, you might update the docstring to include the call signature as the first line so that that it won't appear in `readthedocs` as `output(*args)` but rather as `output(t, x, u)`.  See sphinx `autodoc_docstring_signature` documentation [here](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html).

control/tests/statesp_test.py
```diff
@@ -751,6 +763,58 @@ def test_horner(self, sys322):
...
+        [[1, 1], [[1], [1]], np.atleast_2d([1,1]).T])
+    @pytest.mark.parametrize('u', [0, 1, np.atleast_1d(2)])
+    def test_dynamics_and_output_siso(self, x, u, sys121):
+        assert_array_almost_equal(
+            sys121.dynamics(0, x, u),
+            sys121.A.dot(x).reshape((-1,)) + sys121.B.dot(u).reshape((-1,)))
+        assert_array_almost_equal(
+            sys121.output(0, x, u),
+            sys121.C.dot(x).reshape((-1,)) + sys121.D.dot(u).reshape((-1,)))
+
```
murrayrm, line 776: Add a test for the case where you only pass two arguments?  Could just be added to test_dynamics_and_output_siso as a new case (where you ignore u => a bit wasteful, but easy to add).

control/statesp.py
sawyerbfuller: good idea ... i think that would be much cleaner

control/statesp.py
sawyerbfuller: switch call signature as suggested above

control/tests/statesp_test.py
sawyerbfuller: wanted to do something like that, just couldn't think of how to do it after a few hours of coding : ) Thanks for the suggestion.

control/iosys.py
sawyerbfuller: removed. can add it back in later when there is a system to pass params.

murrayrm (approved):

Issue #523, opened 2021-01-28
stability_margins doesn't find gain=1 crossover frequency correctly for DT systems


sawyerbfuller:
stability_margins doesn't find gain=1 crossover frequency correctly for DT systems, but it seems to mostly do so for CT systems. I think the probelm may be in how `margins.poly_z_mag1_crossing` is finding roots, but there were a few  steps in that function I couldn't follow, e.g. why H(z)*H(1/z)=1 implies |H(z)| = 1. 

Example: 
```python
import control as ct
omegan = 2*np.pi
zeta = 0.2
plant = 1.1*ct.tf(omegan**2, [1, 2*zeta*omegan, omegan**2])
Ts = .05
plantdisc = ct.c2d(plant,Ts,'zoh')
#_ = ct.bode_plot(plant, margins=True) # gives correct gain=1 crossover freq
_ = ct.bode_plot(plantdisc, margins=True) # no gain=1/no phase margin given
```
<img width="407" alt="image" src="https://user-images.githubusercontent.com/58706249/106208565-6f86b100-6178-11eb-874a-809bd05a899b.png">

As an example of what it should look like, here is Matlab:
```matlab
omegan = 2*pi;
zeta = 0.2;
plant = 1.1*tf(omegan^2, [1, 2*zeta*omegan, omegan^2])
Ts = .05;
plantdisc = c2d(plant,Ts,'zoh')
margin(plantdisc)
```
<img width="476" alt="image" src="https://user-images.githubusercontent.com/58706249/106208756-bb395a80-6178-11eb-8bb2-95ee6d928a3d.png">

bnavigator:
> why H(z)*H(1/z)=1 implies |H(z)| = 1

If I am not mistaken, it follows from `|H(z)|²  = H(z) * conj(H(z)) = 1`. [Octave uses the same formula](https://octave.sourceforge.io/control/function/margin.html)

But the algorithm had a bug. Thanks for noticing and reporting.

sawyerbfuller:
I see, Makes sense if we are restricting the root search to the unit circle, where 1/z = conj(z).

 And it has been a little while but I guess it must be that h(conj(z)) = conj(h(z))

sawyerbfuller:
Reopening I found another example that seems to confuse the current version: a 4th order system consisting of two slightly-damped resonances in series: 

```python
import control as ct
omegan = 1
zeta = 0.5
[...]

PR #566, opened 2021-03-12
new method keyword in stability_margins and auto-fallback back to FRD ..

sawyerbfuller:
... when it is detected that there may be numerical inaccuracy in dt polynomial calculation. 
* this is intended to resolve numerical issue observed in DT margin calculation when there is a large difference between the size of numerator and denominator coefficients #523. I am not sure how good the method is I used to detect this issue is any good. In fact I am sure it is not. comments welcome here. 
* can use `method='poly'` to override and force polynomial evaluation, or `method='FRD'` to force to using interpolated FRD to calculate gain and phase crossover frequencies.

### DIFF
--- a/control/margins.py
+++ b/control/margins.py
@@ -300,6 +300,7 @@ def stability_margins(sysdata, returnall=False, epsw=0.0, method='best'):
     determined by the frequency of maximum sensitivity (given by the
     magnitude of 1/(1+L)).
     """
+    # TODO: FRD method for cont-time systems doesn't work
     try:
         if isinstance(sysdata, frdata.FRD):
             sys = frdata.FRD(sysdata, smooth=True)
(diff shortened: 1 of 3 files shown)


control/margins.py
```diff
@@ -300,6 +321,27 @@ def stability_margins(sysdata, returnall=False, epsw=0.0):
...
 
+    if method == 'frd':
+        # convert to FRD if we got a transfer function
+        if isinstance(sys, xferfcn.TransferFunction):
+            omega_sys = freqplot._default_frequency_range(sys)
+            if sys.isctime():
+                sys = frdata.FRD(sys, omega_sys)
+            else:
+                omega_sys = omega_sys[omega_sys < np.pi / sys.dt]
+                sys = frdata.FRD(sys(np.exp(1j*sys.dt*omega_sys)), omega_sys,
```
bnavigator, line 333: Are you trying to hide the fact, that `FRD(sys, omega)` does not handle discrete time systems correctly? ;)

bnavigator: See #568.

murrayrm:
LGTM

Issue #565, opened 2021-03-12
step_info undershoot must be "Undershoot: Percentage of undershoot"


juanodecc:
The function doc:

> https://github.com/python-control/python-control/blob/92de12d48cadd62b7428f1868ef7508930c9a9d5/control/timeresp.py#L774

the code calculate undershoot as:
>https://github.com/python-control/python-control/blob/92de12d48cadd62b7428f1868ef7508930c9a9d5/control/timeresp.py#L818

An example:

```
import numpy as np              # NumPy library
import matplotlib.pyplot as plt
import control as ctrl  # Name: control Version: 0.8.4

sys3 = ctrl.tf([-1, 1],[1,1,1])
ctrl.step_info(sys3)
time,yout=ctrl.step_response(sys3)
plt.figure()
plt.plot(time,yout)

```
{'RiseTime': 1.197344248356904,
 'SettlingTime': 6.631445067822853,
 'SettlingMin': 0.8878785981432115,
 'SettlingMax': 1.2087112656708903,
 'Overshoot': 23.04582827419207,
 **'Undershoot': -0.2791703847438639,**
 'Peak': 1.2087112656708903,
 'PeakTime': 4.236756571109045,
 'SteadyStateValue': 0.9823260833983173}

![Figure_1](https://user-images.githubusercontent.com/21680839/110949000-d7581c00-8320-11eb-93c3-643a9f554d3c.png)

Issue #564, opened 2021-03-12
step_info problem with non stationary systems response


juanodecc:
Systems with non stationary response to step input, step_info could give wrong information:

```
import numpy as np              # NumPy library
import matplotlib.pyplot as plt
import control as ctrl  # Name: control Version: 0.8.4


sys2 = ctrl.tf(1,[1,1,0]) 
time,yout=ctrl.step_response(sys2)
plt.figure()
plt.plot(time,yout)
info=ctrl.step_info(sys2)

```
{'RiseTime': 19.0,
 'SettlingTime': 24.5,
 'SettlingMin': 21.500000000169212,
 'SettlingMax': 23.750000000017877,
 'Overshoot': 0.0,
 'Undershoot': 0.0,
 'Peak': 23.750000000017877,
 'PeakTime': 24.75,
 'SteadyStateValue': 23.750000000017877}

![Figure_16](https://user-images.githubusercontent.com/21680839/110946654-e7223100-831d-11eb-93aa-b071d825b191.png)

The step_info expected result for this system could be:

 tupla with fields:   
 
 RiseTime         :  nan 
 RiseTimeInit     :  nan 
 SettlingTime     :  nan 
 SettlingMin      :  nan 
 SettlingMax      :  nan 
 Overshoot        :  nan 
 Undershoot       :  nan 
 Peak             :  inf 
 PeakTime         :  inf 
 SteadyStateValue :  nan

PR #567, opened 2021-03-12
Step info improve jpp

juanodecc:
This modification improve step_info results: 

Solve the issue Any plan for step_info to support negative step responses? #337
solve the issue step_info problem with non stationary systems response #564
solve the issue step_info undershoot must be "Undershoot: Percentage of undershoot" #565

I try to fullfill pep 8
Add some aditional tests to check the results

### DIFF
--- a/control/tests/timeresp_test.py
+++ b/control/tests/timeresp_test.py
@@ -339,8 +339,8 @@ class TestTimeresp:
                                    [Strue[k] for k in Sktrue],
                                    rtol=rtol)
 
- # tolerance for all parameters could be wrong for some systems ej: discrete systems time parameters tolerance
-    # could be +/-dt
+    # tolerance for all parameters could be wrong for some systems
+    # discrete systems time parameters tolerance could be +/-dt
     @pytest.mark.parametrize(
         "tsystem, info_true, tolerance",
         [("tf1_matlab_help", {
--- a/control/tests/timeresp_test.py
+++ b/control/tests/timeresp_test.py
@@ -193,6 +193,35 @@ class TestTimeresp:
     def no_pole_cancellation(self):
         return TransferFunction([1.881e+06],
                                 [188.1, 1.881e+06])
+    
+    @pytest.fixture
+    def siso_tf_type1(self):
+        # System Type 1 - Step response not stationary:  G(s)=1/s(s+1)
+        return TransferFunction(1, [1, 1, 0])
+
+    @pytest.fixture
+    def siso_tf_kpos(self):
+        # SISO under shoot response and positive final value G(s)=(-s+1)/(s²+s+1)
+        return TransferFunction([-1, 1], [1, 1, 1])
+
+    @pytest.fixture
+    def siso_tf_kneg(self):
+        # SISO under shoot response and negative final value k=-1 G(s)=-(-s+1)/(s²+s+1)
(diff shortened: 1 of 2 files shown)


control/tests/timeresp_test.py
```diff
@@ -303,6 +339,73 @@ def test_step_info(self):
                                    [Strue[k] for k in Sktrue],
                                    rtol=rtol)
 
+ # tolerance for all parameters could be wrong for some systems ej: discrete systems time parameters tolerance
+    # could be +/-dt
```
bnavigator, line 343:
```suggestion
    # tolerance for all parameters could be wrong for some systems
    # discrete systems time parameters tolerance could be +/-dt
```

control/timeresp.py
```diff
@@ -794,34 +794,67 @@ def step_info(sys, T=None, T_num=None, SettlingTimeThreshold=0.02,
...
-            break
+     # TODO: this could be a function step_info4data(t,y,yfinal)
+    rise_time: float = np.NaN
+    settling_time: float = np.NaN
+    settling_min: float = np.NaN
+    settling_max: float = np.NaN
+    peak_value: float = np.Inf
+    peak_time: float = np.Inf
+    undershoot: float = np.NaN
+    overshoot: float = np.NaN
```
bnavigator, line 805: The first use of typing in python-control?

control/timeresp.py
```diff
@@ -794,34 +794,67 @@ def step_info(sys, T=None, T_num=None, SettlingTimeThreshold=0.02,
...
+            tr_lower_index = (np.where(yout >= RiseTimeLimits[0] * InfValue)[0])[0]
+            tr_upper_index = (np.where(yout >= RiseTimeLimits[1] * InfValue)[0])[0]
+            # SettlingTime
+            for i in reversed(range(T.size - 1)):
+                if (yout[i] <= inf_margin) | (yout[i] >= sup_margin):
+                    settling_time = T[i + 1]
+                    break
+            # Overshoot and Undershoot
+            overshoot = np.abs(100. * (yout.max() - InfValue) / InfValue)
+            undershoot = np.abs(100. * yout.min() / InfValue)
```
bnavigator, line 838: Hmm, this has a lot of duplicate code. Have you considered using a `sign` variable and the `abs()` function instead of to completly separate blocks? I think It won'Ttturn out much more convoluted than the current InfValue < 0 part.

control/timeresp.py
juanodecc: yes, is wrong?

control/tests/timeresp_test.py
juanodecc: Im just copy and paste an old test, but I am not satisfied with the solution, because use the same tolerance value for all the output, and this could be not right. If you have any suggestion are well come

control/timeresp.py
juanodecc: ok

control/timeresp.py
bnavigator: Not at all! Good thing we already ditched Python 2.

control/tests/timeresp_test.py
bnavigator: The comment/suggestion is just about the broken formatting

bnavigator:
Other than the MIMO handling (#574), this looks good to me. I'll merge this. Future PRs can improve it further.

PR #569, opened 2021-03-13
Add optimization to flat systems trajectory generation

murrayrm:
This PR adds the ability to generate trajectories for differential flat systems taking into account cost functions and constraints, in a manner that is consistent with the `control.optimal` module.  In particular, you can now call `flat.point_to_point` with `cost` and `constraints` keywords.  The code then uses any extra degrees of freedom in the flat trajectory to minimize cost and satisfy constraints at a list of time points.  For differentially flat systems, this method is _much_ faster than solving an optimal control problem with terminal constraints.

I also found and fixed some docstring glitches and naming inconsistencies in `control.optimal` along the way (these are all minor and don't change behavior.  Unit tests, examples, and updated documentation are included.


sawyerbfuller:
I am not in a position to evaluate this at the moment, but my inclination would be to merge and include with 0.9 so people can play with it/find rough edges sooner rather than later.

PR #580, opened 2021-03-20
change u == 0 test to avoid py3.8+ syntax warning

murrayrm:
PR #546 introduced a test of the form `u == 0` that generates a SyntaxWarning in Python 3.8+.  I changed this to set `u` to `None` by default and then check for `u is None`.  A good description of the reason for the warning is [here](https://adamj.eu/tech/2020/01/21/why-does-python-3-8-syntaxwarning-for-is-literal/).

### DIFF
--- a/control/statesp.py
+++ b/control/statesp.py
@@ -1238,7 +1238,7 @@ class StateSpace(LTI):
         """
         return self._dcgain(warn_infinite)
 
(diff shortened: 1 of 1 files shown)


sawyerbfuller:
good catch, thanks. just so easy to write 'is' by accident in such cases, I thought i had gotten it out of my system but I guess not. : )

PR #583, opened 2021-03-20
Step info improve jpp

juanodecc:
WIP: @bnavigator check this new feature... I tested it with a mimo system from matlab help online. This new version improve the undershoot calculus. The undershoot ocurr when the value have diferent sign respect the y_final... the system must have inverse response

For discusions:
1) In the test for MIMO systems, I had to modifie the result of SettlingTimeMin for pass the test, because I think Matlab result is wrong. (the script take max and min after rise time, and this orcurr at Tr = 0)
2) The Tr calculs have an error of one DT when compute the Tr for digital systems (like mimo system used for testing) because the script compute the Tr with the two points far. (this is easy to implement)

TODO (WIP):  make a new function to compute step_info for data from external simulations or stimulations: step_info_data(t,y,y_final)

### DIFF
--- a/control/tests/timeresp_test.py
+++ b/control/tests/timeresp_test.py
@@ -222,6 +222,15 @@ class TestTimeresp:
         D = [0.06]
         sys = StateSpace(A, B, C, D, 0.2)
         return sys
+    
+    @pytest.fixture
+    def mimo_matlab_help(self):
+        A = [[0.68, -0.34], [0.34, 0.68]];
+        B = [[0.18, -0.05], [0.04, 0.11]];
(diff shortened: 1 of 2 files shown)


bnavigator:
Your commits have a date from 2 days ago and they are based on an older commit. Please check #577, which is already merged. We essentially did very similar changes. Could you have a look and make further improvements such as the undershoot calculation based on current master please?

You can discard your unit tests and just change the existing references for the undershoot value, if necessary.

juanodecc:
ok, Ill try to solve this problem

Issue #592, opened 2021-03-25
Singular values plot missing


forgi86:
Among the frequency-domain analysis functions, a singular values plot is missing in the current version 
of python-control. Is there any plan to include it? I think it is very common and useful for MIMO system analysis.

bnavigator:
I think it depends on what exactly you want to see implemented.

A simple wrapper of feeding MIMO transfer functions or `FrequencyResponseData` objects into `numpy.linalg.svd` and then calling a Matplotlib plot function does not really add any value over calling those functions directly, does it?

forgi86:
The matlab equivalent (function sigma) does this plot on an automatically adjusted logarithmic frequency scale, with singular values in dB, etc. Eventually it should look like a bode plot, but with singular values instead. Sure you can build this plot from scratch, but it takes some time boilerplate code. Today for instance I needed it while working in Python, but I was lazy so I switched to the commercial software :) So I think wrapping it up in a nice way would add some value!

bnavigator:
Sounds reasonable. You are always welcome to submit a PR.

forgi86:
OK maybe I'll try to do it when I have time!

PR #593, opened 2021-03-27
a first implementation of the singular value plot as discussed in #592

forgi86:
- added singular values plot (function singular_values_plot in freqplot.py)
- added a test of the new function (test_singular_values_plot  in freqresp_test.py)
- added an example jupyter notebook (singular-values-plot.ipynb)


control/freqplot.py
```diff
@@ -1039,14 +1040,173 @@ def gangof4_plot(P, C, omega=None, **kwargs):
...
+            if not omega_range_given:
+                # limit up to and including nyquist frequency
+                omega_sys = np.hstack((
+                    omega_sys[omega_sys < nyquistfrq], nyquistfrq))
+        else:
+            nyquistfrq = None
+
+        mag, phase, omega = sys.frequency_response(omega)
+        fresp = mag * np.exp(1j * phase)
+        #fresp = evalfr(sys, 1j * omega_sys)
```
bnavigator, line 1162: avoid back and forth conversion.
```suggestion
        fresp = sys(1j*omega if sys.isctime() else np.exp(1j * omega * sys.dt))
```

control/tests/freqresp_test.py
```diff
@@ -508,3 +509,17 @@ def test_dcgain_consistency():
...
+    den = [75, 1]
+    sys = tf([[[87.8], [-86.4]],
+              [[108.2], [-109.6]]],
+             [[den, den],
+              [den, den]])
+    sigma, omega = singular_values_plot(sys, 0.0)
+    sys_dc = np.array([[87.8, -86.4],
+                       [108.2, -109.6]])
+
+    u, s, v = np.linalg.svd(sys_dc)
```
bnavigator, line 524: Use static reference values

control/tests/freqresp_test.py
```diff
@@ -508,3 +509,17 @@ def test_dcgain_consistency():
 
     sys_ss = ctrl.tf2ss(sys_tf)
     np.testing.assert_almost_equal(sys_ss.dcgain(), -1)
+
+
+def test_singular_values_plot():
```
bnavigator, line 514: Please parametrize with additional:
* StateSpace systems 
* FRD objects.
* SISO 
* ctime and dtime
* non-square input, output MIMO

bnavigator (commented):
Awesome!

I left some comments in the code.

The Notebook looks like omega exceeds the Nyquist frequency

bnavigator:
The test suite failure is unrelated and will be fixed by #588

control/freqplot.py
```diff
@@ -1039,14 +1040,173 @@ def gangof4_plot(P, C, omega=None, **kwargs):
...
+    Parameters
+    ----------
+    syslist : linsys
+        List of linear systems (single system is OK)
+    omega : array_like
+        List of frequencies in rad/sec to be used for frequency response
+    plot : bool
+        If True (default), plot magnitude and phase
+    omega_limits : array_like of two values
+        Limits of the to generate frequency vector.
```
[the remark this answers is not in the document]
forgi86: Thanks! Btw it was copy-pasted from bode_plot, it should also be fixed there

control/freqplot.py
```diff
@@ -1039,14 +1040,173 @@ def gangof4_plot(P, C, omega=None, **kwargs):
...
+    # Make a copy of the kwargs dictionary since we will modify it
+    kwargs = dict(kwargs)
+
+    # Check to see if legacy 'Plot' keyword was used
+    if 'Plot' in kwargs:
+        import warnings
+        warnings.warn("'Plot' keyword is deprecated in bode_plot; use 'plot'",
+                      FutureWarning)
+        # Map 'Plot' keyword to 'plot' keyword
+        plot = kwargs.pop('Plot')
```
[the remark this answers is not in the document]
forgi86: I removed this block of code

control/freqplot.py
```diff
@@ -1039,14 +1040,173 @@ def gangof4_plot(P, C, omega=None, **kwargs):
...
+    Plots a Singular Value plot for the system over a (optional) frequency range.
+
+    Parameters
+    ----------
+    syslist : linsys
+        List of linear systems (single system is OK)
+    omega : array_like
+        List of frequencies in rad/sec to be used for frequency response
+    plot : bool
+        If True (default), plot magnitude and phase
```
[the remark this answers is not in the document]
forgi86: fixed

control/tests/freqresp_test.py
forgi86: What do you mean? 
First I was thinking to use the reference mimo system from function ss_mimo(), but it is actually MISO (2 in, 1 out)

bnavigator:
```
            fresp = sys(1j*omega if sys.isctime() else np.exp(1j * omega * sys.dt))
    
>           fresp = fresp.transpose((2, 0, 1))
E           ValueError: axes don't match array
```

use `squeeze=False`

control/freqplot.py
```diff
@@ -1039,14 +1040,173 @@ def gangof4_plot(P, C, omega=None, **kwargs):
...
+    sigma : ndarray (or list of ndarray if len(syslist) > 1))
+        singular values
+    omega : ndarray (or list of ndarray if len(syslist) > 1))
+        frequency in rad/sec
+
+    Other Parameters
+    ----------------
+    grid : bool
+        If True, plot grid lines on gain and phase plots.  Default is set by
+        `config.defaults['bode.grid']`.
```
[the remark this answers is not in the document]
forgi86: Fixed. BTW, should I also add Hz and dB in Other Parameters?

control/tests/freqresp_test.py
forgi86: Should I define these different systems for this test only?
Perhaps it could be convenient to define some of them as @pytest.fixture?

forgi86:
Something strange about the return shape of the frequency response. For a square 2x2 system:

sys(np.array(0.0), squeeze=False).shape
(2, 2)

sys(0.0, squeeze=False).shape
(2, 2, 1)

Is it expected? Perhaps I should do an np.atleast_1d(omega) inside singular_values_plot anyways?

control/freqplot.py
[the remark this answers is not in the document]
bnavigator: ?

control/tests/freqresp_test.py
bnavigator:
```
s = np.array([197.20868123,   1.39141948])
```

parametrize together with the system (see e.g. timeresp_test.py)

control/tests/freqresp_test.py
bnavigator: Yes, please. See my comment above.

bnavigator:
> Is it expected? Perhaps I should do an np.atleast_1d(omega) inside singular_values_plot anyways?

Nope. Opened #594.

You can work around with `atleast_1d`.

control/freqplot.py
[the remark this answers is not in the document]
forgi86: The bode_plot function (which I took as example)  has a "Parameter" and "Other Parameter" docstring section.
I was wondering where should I put the docstring for the different function arguments of the singular_value_plot
function.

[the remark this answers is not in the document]
bnavigator: I was asking, because I didn't see the change in the commits, despite you wrote "Fixed."

Yes, please add all parameters to the docstring.

control/freqplot.py
```diff
@@ -1160,7 +1160,7 @@ def singular_values_plot(syslist, omega=None,
         else:
             nyquistfrq = None
 
-        fresp = sys(1j*omega if sys.isctime() else np.exp(1j * omega * sys.dt))
+        fresp = sys(1j*omega if sys.isctime() else np.exp(1j * omega * sys.dt)).reshape(sys.noutputs, sys.ninputs, len(omega))
```
bnavigator, line 1163: No. Please make sys return the correct shape by `squeeze=False` and `s` or `z` being a 1D array, which will work around #594 until #595 fixes it.

forgi86: Fixed according to the suggestion, thanks @bnavigator! 

BTW, I was wondering about the return shape of singular_value_plot:

>>> sigma_res, omega_res = ct.singular_values_plot(rss(2,2,2), np.linspace(-3, 2, 1000))
>>> sigma_res.shape
(1000, 2) # omega, singular_value

Is it OK? The general philosophy of python-control seems to be to have "channels first", both for time- and
frequency-domain functions. Perhaps I should transpose the result, then!

control/freqplot.py
```diff
@@ -1039,14 +1040,175 @@ def gangof4_plot(P, C, omega=None, **kwargs):
...
+        if sys.isdtime(strict=True):
+            nyquistfrq = math.pi / sys.dt
+            if not omega_range_given:
+                # limit up to and including nyquist frequency
+                omega_sys = np.hstack((
+                    omega_sys[omega_sys < nyquistfrq], nyquistfrq))
+        else:
+            nyquistfrq = None
+
+        fresp = sys(1j*omega if sys.isctime() else np.exp(1j * omega * sys.dt), squeeze=False)
```
bnavigator, line 1163: Generate s or z in the if branch from `omega_sys` above. `sys.isctime()` is also `True`  for sys.dt = None. Hence the `isdtime(strict=True)`

control/freqplot.py
```diff
@@ -54,9 +54,10 @@
 from .exception import ControlMIMONotImplemented
 from .statesp import StateSpace
 from .xferfcn import TransferFunction
+from .lti import evalfr
```
bnavigator, line 57: deprecated and not used here anymore.

control/tests/freqresp_test.py
```diff
@@ -508,3 +509,84 @@ def test_dcgain_consistency():
...
+
+@pytest.fixture
+def tsystem(request, ss_mimo_t, ss_miso_t, ss_simo_t):
+
+    systems = {"ss_mimo": ss_mimo_t,
+               "ss_miso": ss_miso_t,
+               "ss_simo": ss_simo_t,
+               "ss_siso": ss_simo_t
+               }
+    return systems[request.param]
```
bnavigator, line 583: I moved away from explicitly writing out every test system three times. See #588.

control/tests/freqresp_test.py
```diff
@@ -508,3 +509,84 @@ def test_dcgain_consistency():
...
+
+
+@pytest.mark.parametrize("tsystem", ["ss_mimo", "ss_miso", "ss_simo", "ss_siso"], indirect=["tsystem"])
```
bnavigator, line 586: I think you also want a sigma reference with more than one frequency point. Our whole discussion about squeezing omega illustrates the need to take special attention to different shapes of omega.

control/tests/freqresp_test.py
```diff
@@ -508,3 +509,84 @@ def test_dcgain_consistency():
...
+def test_singular_values_plot(tsystem):
+    sys = tsystem.sys
+    omega = tsystem.omega
+    sigma_check = tsystem.sigma
+    sigma, omega = singular_values_plot(sys, omega, plot=False)
```
bnavigator, line 591: Add a plot=True test. Needs the matptlotlib cleanup fixture.

bnavigator (changes requested):

control/tests/freqresp_test.py
forgi86: Sorry I don't understand this. What do you mean?

control/tests/freqresp_test.py
forgi86: I definitely agree, I would also add discrete-time examples. 
I was wondering if the format of the tests is OK. 

Should I define these "test systems" together with the expected values as TSys objects, like it was in the timeresp_test.py that you were mentioned? I followed this approach now but I cannot understand your previous comment 
"I moved away from explicitly writing out every test system three times" so I am a bit stuck on this point!

bnavigator: Sorry. I understand it might be a bit confusing. I first suggested the approach, which you implemented. It works well and you can leave it as is.

Howver, during my work for #588 I got the feeling, that creating so many separate test system fixtures which are all used through the proxy `tsystem` fixture anyway, introduces unnecessary extra code. You have to write out all the system names multiple times. So I moved everything into the `tsystem` fixture.

Yes, please attach the frequency vectors together with expected svd values as attributes to the TSys objects.

control/tests/freqresp_test.py
bnavigator: See below

control/freqplot.py
forgi86: removed

control/freqplot.py
forgi86: done

control/freqplot.py
bnavigator: Yes it should be consistent with `bode_plot()` and thus:

- with `frequency_response()`:
  ```
        Returns
        -------
        mag : ndarray
            The magnitude (absolute value, not dB or log10) of the system
            frequency response.  If the system is SISO and squeeze is not
            True, the array is 1D, indexed by frequency.  If the system is not
            SISO or squeeze is False, the array is 3D, indexed by the output,
            input, and frequency.  If ``squeeze`` is True then
            single-dimensional axes are removed.
   ```

- with `TransferFunction.__call__()` and `StateSpace.__call__()`:
  ```
        Returns
        -------
        fresp : complex ndarray
            The frequency response of the system.  If the system is SISO and
            squeeze is not True, the shape of the array matches the shape of
            omega.  If the system is not SISO or squeeze is False, the first
            two dimensions of the array are indices for the output and input
            and the remaining dimensions match omega.  If ``squeeze`` is True
            then single-dimensional axes are removed.
  ```

control/freqplot.py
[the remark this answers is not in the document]
forgi86: Fixed (now for real!)

control/freqplot.py
bnavigator: Of course SISO does not really make sense for SVD.

forgi86: Then I should definitely transpose the result before returning the result!
sigma = sigma.transpose()

Do you care about returning a c-contiguous result? In case I should  do something like
sigma = np.ascontiguousarray(sigma.transpose())

control/tests/freqresp_test.py
bnavigator: Sorry, my bad. It is used automatically in freqresp_test.py:

https://github.com/python-control/python-control/blob/8b900ca222490f985f4bfe4ff07df9afe8bcad36/control/tests/freqresp_test.py#L22

If that would not be the case, you would have to add it like this:
https://github.com/python-control/python-control/blob/8b900ca222490f985f4bfe4ff07df9afe8bcad36/control/tests/nichols_test.py#L21

It makes sure that any created plot does not stay open so it would interfere with subsequent unit tests.

Just make sure to call all the code branches of the new function. You can check the coveralls page, once a successful run is analyzed by it. `plot=True` would be one case for this.

control/freqplot.py
bnavigator: > Do you care about returning a c-contiguous result? In case I should do something like
> sigma = np.ascontiguousarray(sigma.transpose())

I don't think that is necessary.

forgi86: OK tks! 

BTW my function does accept SISO input, you get back as SVD the absolute value (aka magnitude).

control/freqplot.py
```diff
@@ -1039,14 +1039,183 @@ def gangof4_plot(P, C, omega=None, **kwargs):
...
+        if sys.isdtime(strict=True):
+            nyquistfrq = math.pi / sys.dt
+            if not omega_range_given:
+                # limit up to and including nyquist frequency
+                omega_sys = np.hstack((
+                    omega_sys[omega_sys < nyquistfrq], nyquistfrq))
+            omega_complex = np.exp(1j * omega * sys.dt)
+        else:
+            nyquistfrq = None
+            omega_complex = 1j*omega
```
bnavigator, line 1168: Shouldn't that be

```suggestion
            omega_complex = np.exp(1j * omega_sys * sys.dt)
        else:
            nyquistfrq = None
            omega_complex = 1j*omega_sys
```

?

forgi86: Yes, be bad!

forgi86:
> Whishlist:
> Can we (optionally) have the left and right singular vectors, too?

Sounds reasonable. From a control perspective they correspond to the output/input directions associated with the corresponding singular value. There's no way to plot them in a nice way, but they can indeed be optionally returned.

control/tests/freqresp_test.py
forgi86: I added more tests with variable-length frequency vectors, and also one with a discrete-time system.

Now I see what you mean by "explicitly writing out every test system three times", it is indeed a bit verbose and error-prone, but it works :)

control/freqplot.py
```diff
@@ -1039,14 +1040,173 @@ def gangof4_plot(P, C, omega=None, **kwargs):
...
+        else:
+            omega_range_given = True
+            omega_limits = np.asarray(omega_limits)
+            if len(omega_limits) != 2:
+                raise ValueError("len(omega_limits) must be 2")
+            if Hz:
+                omega_limits *= 2. * math.pi
+            omega = np.logspace(np.log10(omega_limits[0]),
+                                np.log10(omega_limits[1]), num=omega_num,
+                                endpoint=True)
```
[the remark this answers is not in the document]
sawyerbfuller: Very nice PR. I second @bnavigator's comment here. it would be great to have a common function to do what this code does, something analogous to the `_default_time_vector` private function in `timeresp.py` which would basically include this code. We could then also use it in the neglected `nichols` function as well (which should probably be put into `freqplot.py` IMO)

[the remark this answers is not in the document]
forgi86: Thanks @sawyerbfuller! Working on it :) BTW I am not sure whether the  _default_frequency_range as it is now is good for the svd plot also. On the distillation column example that I am testing the returned range seems to be too high-frequency!
Haven't looked inside this function yet. 

I'll do the refactoring anyways as as also  @bnavigator suggests (the rest of the logic is indeed the same for all the frequency-domain plots I guess)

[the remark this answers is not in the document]
forgi86: DONE! Now this logic is encapsulated in a private function:

omega_out, omega_range_given = _determine_frequency_range(syslist, omega_in, omega_limits, omega_num, Hz)

used by bode_plot and singular_values_plot. BTW, the _default_frequency_range function (which is invoked when the user
does not specify a range) it is perhaps not thought for a singular value plot and in some cases does not seem to produce a really "optimal" range. I think there is room for improvement in a future development...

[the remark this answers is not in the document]
sawyerbfuller: Sounds good. Perhaps a keyword could be added to `_default_frequency_range` in the future to request a range suitable for svd. 

Do you think it would make sense to change the name of the new function so that it is harder to confuse its purpose with that of `_default_frequency_range` ? One option would be something like `_default_omega_vector`, to parallel the equivalent function name in timeresp.py

control/tests/freqresp_test.py
```diff
@@ -508,3 +509,124 @@ def test_dcgain_consistency():
...
+@pytest.fixture
+def ss_siso_ct():
+    A = np.diag([-1 / 75.0])
+    B = np.array([[1.0]]) / 75.0
+    C = np.array([[87.8]])
+    D = np.zeros((1, 1))
+    T = TSys(ss(A, B, C, D))
+    T.omegas = [0.0]
+    T.sigmas = [np.array([[87.8]]),
+                np.array([[87.8, 70.24]])]
```
bnavigator, line 577: It's missing the second omega vector.

control/tests/freqresp_test.py
```diff
@@ -508,3 +509,124 @@ def test_dcgain_consistency():
...
+@pytest.mark.parametrize("tsystem",
+                         ["ss_mimo_ct", "ss_miso_ct", "ss_simo_ct", "ss_siso_ct", "ss_mimo_dt"], indirect=["tsystem"])
+def test_singular_values_plot(tsystem):
+    sys = tsystem.sys
+    for omega_ref, sigma_ref in zip(tsystem.omegas, tsystem.sigmas):
+        sigma, _ = singular_values_plot(sys, omega_ref, plot=False)
+        np.testing.assert_almost_equal(sigma, sigma_ref)
+
+
+def test_singular_values_plot_mpl(ss_mimo_ct):
```
bnavigator, line 618: Please also cover the code for the nyquistfreq line and check the correct location.

control/freqplot.py
```diff
@@ -190,35 +190,15 @@ def bode_plot(syslist, omega=None,
     grid = config._get_param('bode', 'grid', kwargs, _bode_defaults, pop=True)
     plot = config._get_param('bode', 'grid', plot, True)
     margins = config._get_param('bode', 'margins', margins, False)
-    wrap_phase = config._get_param(
-        'bode', 'wrap_phase', kwargs, _bode_defaults, pop=True)
-    initial_phase = config._get_param(
-        'bode', 'initial_phase', kwargs, None, pop=True)
+    wrap_phase = config._get_param('bode', 'wrap_phase', kwargs, _bode_defaults, pop=True)
+    initial_phase = config._get_param('bode', 'initial_phase', kwargs, None, pop=True)
+    omega_num = config._get_param('freqplot', 'number_of_samples', omega_num)
```
bnavigator, line 195: Please don't exceed 79 characters per line [(PEP8)](https://www.python.org/dev/peps/pep-0008/#maximum-line-length)

control/freqplot.py
```diff
@@ -1039,14 +1019,219 @@ def gangof4_plot(P, C, omega=None, **kwargs):
...
+            ax_sigma = plt.subplot(111, label='control-sigma')
+
+        color_cycle = plt.rcParams['axes.prop_cycle'].by_key()['color']
+
+    sigmas, omegas, nyquistfrqs = [], [], []
+    for idx_sys, sys in enumerate(syslist):
+        omega_sys = np.asarray(omega)
+        if sys.isdtime(strict=True):
+            nyquistfrq = math.pi / sys.dt
+            if not omega_range_given:
```
bnavigator, line 1123: Issue a warning if a given omega is above nyquistfrq?

control/freqplot.py
forgi86: Sorry, I just noticed that my IDE (PyCharm) uses 120 as default, which is not PEP!
Will change back :)

control/freqplot.py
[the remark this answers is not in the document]
forgi86: Name changed! Yes perhaps an option will be required if we come up with a better and specific logic for this plot.

control/tests/freqresp_test.py
forgi86: Added!

control/freqplot.py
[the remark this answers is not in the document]
forgi86: Yes a keyword may be the right way to do it if we find a better and specialized logic for the frequency range in 
the case of an svd plot. 

_default_omega_vector is also a bit misleading IMHO as the function computes the actually used range (the user-defined 
if specified, or the default auto range otherwise). The radix _default  is more appropriate for  the _default_frequency_range 
function I'd say.

I'd propose determine_omega_vector for this new function (or something else that gives the idea of a selection logic and not an autotuned range, I'm no mother-toungue)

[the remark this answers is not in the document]
sawyerbfuller: That name and reasoning sounds good to me.

control/tests/freqresp_test.py
forgi86: Added a test that checks this (and also checks the superposition of 2 singular values plot)

control/freqplot.py
forgi86: Don't have a strong opinion about it. I added a warning for the singular value plot now (but then also the bode plot should have it I guess)

control/freqplot.py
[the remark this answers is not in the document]
forgi86: BTW the logic for the user-provided vector is not consistent I'd say:  if Hz=True, a user-provided omega_limits is  interpreted as Hz and multiplied times 2pi, but a user-provided omega is not modified.

control/freqplot.py
murrayrm: Since the Nyquist frequency is plotted on the Bode/singular value plot, it seems like a warning may not be needed.  If you do add a warning, it would be nice to be able to turn it off (without having the user have to figure out how to filter warnings => keyword?).

control/freqplot.py
```diff
@@ -1039,14 +1021,239 @@ def gangof4_plot(P, C, omega=None, **kwargs):
...
+# Singular value plot
+#
+
+
+# Default values for Bode plot configuration variables
+_singular_values_plot_default = {
+    'singular_values_plot.dB': False,           # Plot singular values in dB
+    'singular_values_plot.deg': True,           # Plot phase in degrees
+    'singular_values_plot.Hz': False,           # Plot frequency in Hertz
+    'singular_values_plot.grid': True,          # Turn on grid for gain and phase
```
murrayrm, line 1034: It seems a bit odd to me to have to set these separately from the equivalent `bode` configuration variable.  Is someone going to use one set of defaults for Bode and a separate one for singular values?  (Also note that `gang_of_four` uses the Bode plot defaults.)

If we don't want to use `bode.dB`, etc, we could also change all of them to `freqplot.dB` (though perhaps with some backward compatibility??).

control/freqplot.py
```diff
@@ -1039,14 +1021,239 @@ def gangof4_plot(P, C, omega=None, **kwargs):
...
+
+    # Get values for params (and pop from list to allow keyword use in plot)
+    dB = config._get_param(
+        'singular_values_plot', 'dB', kwargs, _singular_values_plot_default, pop=True)
+    Hz = config._get_param(
+        'singular_values_plot', 'Hz', kwargs, _singular_values_plot_default, pop=True)
+    grid = config._get_param(
+        'singular_values_plot', 'grid', kwargs, _singular_values_plot_default, pop=True)
+    plot = config._get_param(
+        'singular_values_plot', 'grid', plot, True)
```
murrayrm, line 1098: 'grid' should be 'plot'.

examples/singular-values-plot.ipynb
```diff
@@ -0,0 +1,3277 @@
...
+   "metadata": {},
+   "outputs": [
+    {
+     "ename": "RuntimeError",
+     "evalue": "module compiled against API version 0xe but this version of numpy is 0xd",
+     "output_type": "error",
+     "traceback": [
+      "\u001B[0;31m---------------------------------------------------------------------------\u001B[0m",
+      "\u001B[0;31mRuntimeError\u001B[0m                              Traceback (most recent call last)",
+      "\u001B[0;31mRuntimeError\u001B[0m: module compiled against API version 0xe but this version of numpy is 0xd"
```
murrayrm, line 16: Store notebook output has errors.

control/freqplot.py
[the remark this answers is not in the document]
sawyerbfuller: I wrote some of that code and it’s very possible that I didn’t think about that. So, probably a bug.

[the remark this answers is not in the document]
forgi86: Just a small detail :). 
The returned vector is always in rad/sec instead, irrespectively of the Hz parameter. This is a feature I guess, not a bug right?

examples/singular-values-plot.ipynb
forgi86: Fixed

control/freqplot.py
forgi86: Fixed (also in bode_plot). BTW the plot option is an optional argument of the two functions with a default value set in the function, this line of code could also be removed right?

control/freqplot.py
```diff
@@ -1039,14 +1040,173 @@ def gangof4_plot(P, C, omega=None, **kwargs):
...
+    grid : bool
+        If True, plot grid lines on gain and phase plots.  Default is set by
+        `config.defaults['bode.grid']`.
+
+    Examples
+    --------
+    >>> den = [75, 1]
+    >>> sys = ct.tf([[[87.8], [-86.4]], [[108.2], [-109.6]]], [[den, den], [den, den]])
+    >>> omega = np.logspace(-4, 1, 1000)
+    >>> sigma, omega = singular_values_plot(sys)
```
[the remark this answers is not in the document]
forgi86: I have to admit I am not too familiar with doctest. I thought about something like this:

    >>> den = [75, 1]
    >>> sys = ct.tf([[[87.8], [-86.4]], [[108.2], [-109.6]]], [[den, den], [den, den]])
    >>> omega = np.logspace(-4, 1, 1000)
    >>> sigma, omega = singular_values_plot(sys)
    >>> singular_values_plot(sys, 0.0, plot=False)
    (array([[197.20868123],
        [  1.39141948]]),
     array([0.]))

Since it is a plotting function, I'd like to keep as first an example that produces a figure and does have text output 
(would be too long...) and as second example one that produces a small output (evaluation at one point) that can be checked. Makes sense?

control/freqplot.py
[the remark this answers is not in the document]
sawyerbfuller: I think the consensus plan (not currently complete) is to have something like the following, arranged in terms of a pure math to engineering heirarchy: 

Sys(s/z) - done 

Sys.frequency_response(omega) -> (currrently returns mag and phase and should return sys(j*omega)  ) 

Sys.bode(omega) -> returns mag and phase, and can give omega in hz instead of rad/sec. plot optional. 

From that perspective, my take is that if Hz is true, then omega should be returned as hz too?

control/freqplot.py
bnavigator: https://github.com/python-control/python-control/pull/593#discussion_r602856380

control/freqplot.py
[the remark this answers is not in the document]
forgi86: Ah, I thought it was intended! 
As a user, I would also perhaps expect the output frequency to be in Hz (or at least in cycles/[UNIT_OF_TIME]) if I specify the Hz=True option, but this is not the case in the current implementation of bode (and in the one of the singular_value_plot in this PR).  Yet, I would perhaps expect the dB option to be for the plot only (don't know why, just my feeling!) 

Shall we deal with this issue in this PR, or keep it for another one?

control/freqplot.py
forgi86: I'd add the properties dB, Hz, and grid to freqplot and use these for bode, singular_values_plot, and gang_of_four.
The property _singular_values_plot_default would then disappear, and _bode_defaults would still contain bode.deg and bode.wrap_phase (and maybe unused dB, Hz, grid for backward compatibility, at least for the moment).

Is this OK?

control/freqplot.py
forgi86: If everybody agrees, I would avoid issuing a warning at all to avoid additional logic that does not seem to add value.

control/freqplot.py
murrayrm: That seems fine. Make sure also to update the `use_matlab_defaults` and `use_fbs_defaults` functions in `config.py`.  I'm not sure about what to do with `bode.deg`.  On the one hand, it is true that this is the only function that plots phase, but at the same time seems odd to use `freqplot` for everything except that.  I would probably just use `freqplot` for everything.

control/freqplot.py
sawyerbfuller: ok with  me to leave out warning

control/freqplot.py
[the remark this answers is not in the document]
forgi86: I pushed a new version where the passed omega is interpreted in Hz if Hz=True.
Note that I had to modify certain tests in config_test.py which were assuming instead omega to be still in rad/sec 
(and thus, Hz assumed to be only for plotting).  
Not sure whether there is really consensus on this point!

control/freqplot.py
forgi86: I pushed a new version where all the settings for bode_plot and singular_values_plot are in _freqplot_defaults

bnavigator: I still see a `_singular_values_plot_default` with a comment for "Bode" above the definition and then used with_get_param('freqplot'). That's very confusing.

Also, personally, I would still prefer that we provide some backwards compatibility through an alias of config.defaults['bode'], so that old code does not break right away. Maybe including a warning message. I am not sure how to achieve this with a dict outside of `set_defaults` and `_get_param` in an elegant way.

control/freqplot.py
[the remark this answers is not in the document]
bnavigator: > Not sure whether there is really consensus on this point!
That's a really good question!

The docstring still says 

```
    omega : array_like
        List of frequencies in rad/sec to be used for frequency response

    Returns
    -------
   ...
    omega : ndarray (or list of ndarray if len(syslist) > 1))
        frequency in rad/sec
```

If we deviate from that, we also have to adjust the docstring.

`omega` as opposed to `frequency`, `freq` or `f` also implies natural frequencies in rad/sec. All other frequency domain functions with input parameter `omega` expect rad/sec.

I vote to keep it strictly rad/sec.

control/freqplot.py
forgi86: I tried to remove all references to the previous structure _singular_values_plot_default (new push).
Still have to handle backward compatibility in some way.

control/freqplot.py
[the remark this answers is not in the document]
forgi86: I would also prefer fixed units for input/outputs (for frequency, magnitude, and also for the phase in the  bode plot). 
Units options 'dB', 'Hz', and 'deg' then would only be used for the visualization. 
This interpretation is easier to document and less error-prone.

control/freqplot.py
```diff
@@ -1078,7 +1069,7 @@ def singular_values_plot(syslist, omega=None,
     ----------------
     grid : bool
         If True, plot grid lines on gain and phase plots.  Default is set by
-        `config.defaults['singular_values_plot.grid']`.
+        `config.defaults['freqplot.number_of_samples']`.
```
bnavigator, line 1072: ?

forgi86: was fixed with the latest push

forgi86: was fixed in the last commit

bnavigator: Still there.

bnavigator:
```suggestion
        `config.defaults['freqplot.grid']`.
```

bnavigator:
https://github.com/python-control/python-control/blob/8f3f76debd0cb95725d4070a29c158f6698ea6b8/doc/conventions.rst#package-configuration-parameters needs an update.

control/freqplot.py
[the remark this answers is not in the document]
sawyerbfuller: Ok sounds reasonable to me. I Agree that allowing input range to be given in hz would entail a lot of breaking changes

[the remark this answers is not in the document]
forgi86: New version pulled. The parameter 'Hz' is now only for the plot, and input/outputs are always in 
rad/sec.

Changing this behavior entaild indeed several changes and should be discussed separately.

forgi86:
Variables renamed in conventions.rst (bode -> freqplot)

sawyerbfuller (approved):

control/freqplot.py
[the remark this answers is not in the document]
bnavigator:
```
[ben@skylab:…python-control]% pytest --doctest-modules -k singular_values_plot control/freqplot.py                            [1]
====================================================== test session starts =======================================================
platform linux -- Python 3.8.8, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
Matplotlib: 3.4.1
Freetype: 2.6.1
rootdir: /home/ben/src/python-control, configfile: setup.cfg
plugins: timeout-1.4.2, mpl-0.12, xvfb-2.0.0, doctestplus-0.8.0, cov-2.11.1, hypothesis-6.8.1
collected 4 items / 3 deselected / 1 selected                                                                                    

control/freqplot.py F                                                                                                      [100%]

============================================================ FAILURES ============================================================
________________________________________ [doctest] control.freqplot.singular_values_plot _________________________________________
1068     Other Parameters
1069     ----------------
1070     grid : bool
1071         If True, plot grid lines on gain and phase plots.  Default is set by
1072         `config.defaults['freqplot.grid']`.
1073 
1074     Examples
1075     --------
1076     >>> den = [75, 1]
1077     >>> sys = ct.tf([[[87.8], [-86.4]], [[108.2], [-109.6]]], [[den, den], [den, den]])
UNEXPECTED EXCEPTION: NameError("name 'ct' is not defined")
  File "/usr/lib64/python3.8/doctest.py", line 1336, in __run
    exec(compile(example.source, filename, "single",
  File "<doctest control.freqplot.singular_values_plot[1]>", line 1, in <module>
NameError: name 'ct' is not defined
/home/ben/src/python-control/control/freqplot.py:1077: UnexpectedException
==================================================== short test summary info =====================================================
FAILED control/freqplot.py::control.freqplot.singular_values_plot
```

control/freqplot.py
```diff
@@ -1039,14 +1025,237 @@ def gangof4_plot(P, C, omega=None, **kwargs):
...
+    Examples
+    --------
+    >>> den = [75, 1]
+    >>> sys = ct.tf([[[87.8], [-86.4]], [[108.2], [-109.6]]], [[den, den], [den, den]])
+    >>> omega = np.logspace(-4, 1, 1000)
+    >>> sigma, omega = singular_values_plot(sys, plot=True)
+    >>> singular_values_plot(sys, 0.0, plot=False)
+    (array([[197.20868123],
+        [  1.39141948]]),
+     array([0.]))
```
bnavigator, line 1083:
```suggestion
    >>> import numpy as np
    >>> from control import tf, singular_values_plot
    >>> den = [75, 1]
    >>> sys = tf([[[87.8], [-86.4]], [[108.2], [-109.6]]],
    ...          [[den, den], [den, den]])
    >>> sigma, omega = singular_values_plot(sys, plot=True)
    >>> singular_values_plot(sys, 0.0, plot=False)
    (array([[197.20868123],
        [  1.39141948]]),
     array([0.]))
```

control/freqplot.py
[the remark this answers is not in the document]
forgi86: OK I see. The new version should be doctest compliant. Now

pytest --doctest-modules -k singular_values_plot control/freqplot.py 

does not return any errors

control/freqplot.py
forgi86: Fixed

control/freqplot.py
```diff
@@ -1039,14 +1025,238 @@ def gangof4_plot(P, C, omega=None, **kwargs):
...
+
+    Other Parameters
+    ----------------
+    grid : bool
+        If True, plot grid lines on gain and phase plots.  Default is set by
+        `config.defaults['freqplot.grid']`.
+
+    Examples
+    --------
+    >>> import numpy
```
bnavigator, line 1076:
```suggestion
    >>> import numpy as np
```

Please, either import under the name as it is used later, or leave the import out (together with `singular_values_plot` and change the `tf` call to `TransferFunction`), because the namespace of the module itself is available during a `pytest --doctest-modules` call.

forgi86: done, I kept the numpy as np import and used TransferFunction instead of tf

bnavigator:
Thank you @forgi86 for the hard work and also for the patience with all the picky reviews.

I will submit a PR shortly which addresses the backwards compatibility with the old name for the bode defaults.

forgi86:
Thank you for the review and for your kind message, @bnavigator. Was my first PR to python-control, I hope I will be able to contribute again in the future!

PR #603, opened 2021-04-02
remove statement that slycot only on linux

sawyerbfuller:
removed incorrect statement in intro.rst that slycot is only available on linux. It is now available on windows and Mac, too. And removed mention of fortran compiler; if somebody wants to compile slycot they can go the slycot project page to learn how.


ilayn:
On windows it is indeed next to impossible to compile it properly. The instructions are welcome but that is certainly not a trivial thing. And I should know since I am compiling SciPy everyday which involves Fortran, C++ and C. But so far no avail.

Issue #278, opened 2019-02-25
Create latex output for Jupyter notebooks


murrayrm:
PR #276 adds latex output for transfer functions.  It would be nice to add this for other LTI types as well (particularly StateSpace objects).  Having the output match the text approximation in `__repr__` seems like a good first cut at format to use.

repagh:
I got this representation in a Spyder console
![image](https://user-images.githubusercontent.com/2794509/59916764-59ebca00-9420-11e9-9e63-8805c58aad57.png)
I think we need to get the numbers printed in exponent form a bit nicer still.

roryyorke:
Here's a proof of concept for SS, presenting the system in "partitioned matrix" form.

The format is loosely based on Sympy's LaTeX output for matrices with floating point entries, which centre-aligns and uses a `\cdot` to separate the `10^x` part.

Some points:

  - MathJax doesn't support `@` in `\array` preamble [1], so one can't align columns on the decimal point.  One could left-align, and use `\llap{-}` for the minus signs; then the leading digits are aligned in each column.  MathJax limitations only apply for notebook output.
  - If one uses a "tabular" output like this, perhaps `_repr_html_` would be better than `_repr_latex_`; I haven't investigated that.  That might be better if one, e.g., labelled rows with state and output names, and columns with state and input names.
  - this doesn't match the current `repr` output (`A = [[ ... ]] B = [[ ... ]]`, etc.)
  - a complete display doesn't make sense for big (nx, ny, nu > 10?) systems; we could have something like numpy's or panda's display limits.  I think something similar applies to TFs, i.e., we should omit intermediate polynomial terms for high-degree denominators and numerators.  One trick would be still showing the partitions, so this could get pretty untidy-looking.

[1] https://github.com/mathjax/MathJax-docs/issues/205

```python
import numpy as np
from IPython.display import Latex

def _f2s(f):
    s = f'{f:.3g}'
    if 'e' in s:
        n, e = s.split('e')
        e = int(e)
        s = f'{n}\\cdot10^{{{e}}}'
    return s


def format_ss(A, B, C, D):
    nx = A.shape[0]
    nu = B.shape[1]
    ny = C.shape[0]

    lines = [
        r'\[',
        r'\left(',
        r'\begin{array}{' + 'c' * nx + '|' + 'c' * nu + '}'
        ]

    for Ai, Bi in zip(A, B):
        lines.append('&'.join([_f2s(Aij) for Aij in Ai]
[...]

PR #607, opened 2021-04-11
IPython LaTeX output only generated for small systems

roryyorke:
StateSpace._repr_latex_ now returns None for systems whose size is
greater than new config variable statesp.latex_maxsize.  System size
is the largest dimension of the partitioned system matrix.
statesp.latex_maxsize is 10.

See conversation staring at https://github.com/python-control/python-control/issues/278#issuecomment-811603495


bnavigator (approved):
LGTM

PR #616, opened 2021-04-28
Updated rlocus.py to remove warning by sisotool() with rlocus_grid=True

nirjhar-das:
At line 239 in rlocus.py parameter passed to _sgrid_func() has been corrected. It was giving error when calling sisotool() with parameter rlocus_grid=True.


bnavigator:
Thank you @nirjhar-das!

PR #643, opened 2021-07-31
Various docstring edits + fixed plot legends on cruise control example

billtubbs:


billtubbs:
Oh.  I just noticed there's a [jupyter notebook version](examples/cruise.ipynb) of the cruise control example.  I'll have to update that too.

billtubbs:
OKay, there's something I don't understand about editing a jupyter notebook.  After I make the edits and run the notebook locally, the setup.py test is failing (see above):

```
File "/usr/share/miniconda/lib/python3.9/site-packages/jupyter_client/kernelspec.py", line 230, in get_kernel_spec
    raise NoSuchKernel(kernel_name)
jupyter_client.kernelspec.NoSuchKernel: No such kernel named conda-env-control-dev-py
```

examples/cruise.ipynb
```diff
@@ -884,9 +893,9 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "Python [conda env:control-dev]",
    "language": "python",
-   "name": "python3"
+   "name": "conda-env-control-dev-py"
```
murrayrm, line 898: This won't work in the test environment.

murrayrm:
The problem with Jupyter notebooks is that they inherit information from your local machine.  So the problem is probably something like `conda-env-control-dev-py` is a local environment on your machine but it doesn't exist in the test environment.  I marked this line in the code so that you can see where it occurs.

One fix is to edit the Jupyter notebook directly (in a text editor) and change the code you want to fix.  Then commit that change and test it locally, throwing out the new version of the notebook that will be created when you run it (but still making sure it works).  Push that up to GitHub and it should be OK.  One caveat is that this will leave the notebook with the old figures, which may not be what you want.

Alternatively, run the notebook on your local machine and get things working, then go in and manually change the line that said `conda-env-control-dev-py` to say `python3`.

PR #649, opened 2021-08-26
Time response data class

murrayrm:
This PR adds a `TimeResponseData` class, following the discussion in #645, which provides a more object-oriented approach to the representation of time response data.  All time responses functions now return an instance of `TimeResponseData` containing the results of simulations.  The class implements an ``__iter__`` method that allows backward compatibility with current call signatures.

Old approach (still works):
```
t, y = ct.step_response(siso_sys)
plt.plot(t, y)

t, y = ct.step_response(mimo_sys)
plt.plot(t, y[0, 1], label="input 1 to output 0")
plt.plot(t, y[1, 1], label="input 1 to output 1")
```

New approach:
```
response = ct.step_response(siso_sys)
plt.plot(response.time, response.outputs)

response = ct.step_response(mimo_sys)
plt.plot(response.time, response.outputs[0, 1], label="input 1 to output 0")
plt.plot(response.time, response.outputs[1, 1], label="input 1 to output 1")
```

The attributes `time`, `outputs`, `states` and `inputs` are actually class properties that carry out "squeeze processing" allowing use of 1D arrays for SISO systems.  Time response data can also be accessed via the attributes `t`, `y`, `x`, and `u`, which are always in "MIMO" format (indexed by output, input (trace), and time:
```
response = ct.step_response(siso_sys)
plt.plot(response.t, response.y[0, 0])
```

Other notes:

* I adopted the terminology of a "trace" to handle the time response data returned by `step_response` and `impulse_response`, where a MIMO system generates a full matrix of input/output responses.  The output, state, and input responses for those functions are represented by a 3D array indexed by output/state/input, trace, and time.

* The terminology for `time`, `outputs`, `states` and `inputs` was chosen to be consistent with the `OptimalControlResult` class.  This is implemented via a set of properties that perform the squeeze (and transpose) processing to provide compatibility with our existing functionality.

* The `InputOutputData` class is currently only used to store the output of simulations, but it is set up to serve as a pure data representation class (similar to `FrequencyResponseData`) and could later be useful for things like input to system identification routines.

* No changes to existing unit tests were needed => should be fully backwards compatible.

At a future date, we may want to consider adding some additional functionality:

* Plot methods that plot input/output simulations in a uniform way
* Methods to convert data into `DataFrame` format for use of pandas plotting capabilities

I'm still not completely sure if this is a useful PR, but I was motivated by the way a similar approach taken in [JuliaControl](https://github.com/JuliaControl/ControlSystems.jl/pull/529) and thought I would give it a try.  Suggestions and comments welcome (will leave in draft form for a bit to allow for discussion).


murrayrm:
> Presumably when the latter [plotting functionality] comes we can do things like add label text, which can come in handy.

I've added the ability to store labels in the latest commit and set things up so that the signals names for I/O systems, if present, are copied over as the default labels for `input_output_response`.

sawyerbfuller:
@billtubbs Would this PR help with what you were working on for #645 ? ie., should we merge this PR?

murrayrm:
I've been using the branch with this PR as my default branch for a while and everything seems OK.  Ready to be merged, I think.

@billtubbs You mentioned in a separate note that you had made progress on #645.  Does this conflict with that?  If not, we can probably go ahead and merge.

bnavigator (approved):

billtubbs:
> @billtubbs You mentioned in a separate note that you had made progress on #645. Does this conflict with that? If not, we can probably go ahead and merge.

I don't think so. I haven't looked at this in detail but I was only planning to fix the specific axes and figure problems with some of the plot functions `rlocus`, `pzmap`, `bode_plot`, `nyquist_plot`, `gangof4_plot`, `nichols_plot`.  My work is probably already behind the master so will need revising anyway if there are any impacts.

Issue #650, opened 2021-09-04
Discrete system bode_plot nyquist line does not respect Hz parameter


dc5048:
https://github.com/python-control/python-control/blob/18976c1eabe56a6f9afe0aadea6679024b84b335/control/freqplot.py#L332

This line should be:
`omega_nyq_line = np.array((np.nan, nyquistfrq_plot, nyquistfrq_plot))`

... so that the nyquist limit will be plotted in Hz when requested.  Currently the nyquist limit always plots as rad/s, even if the Hz argument is True.

bnavigator:
Closed by #651

PR #651, opened 2021-09-06
FIX: plot Nyquist frequency correctly in Bode plot in Hz

murrayrm:
Quick fix for issue #650, using solution from @dc5048.


bnavigator:
For reference:

```python
import control
tf1 = control.tf(control.drss(3,1,1))
tf1.dt=0.1
control.bode_plot(tf1, Hz=True)
```

| Before | After |
| --- | ---|
| pr-651-pre | pr-651-post |

(Two different random discrete transfer functions)

PR #652, opened 2021-09-11
Check for unused subsystem signals in InterconnectedSystem

roryyorke:
Add capability to check for unused signals in InterconnectedSystem;
this check is invoked by default by `interconnect`.

This would have saved me a few hours when I made three typos in signal names, and I couldn't figure out why my time and frequency responses were all wrong.

Some questions:
  - is check_unused=True a reasonable default?  I think it is; it won't take long to scan the unused signals, and, if necessary, add them to the `ignore_*` lists
  -  are the `ignore_*` names reasonable?  I initially called them `dropped_inputs` and `dropped_outputs`, but I decided against that.  `known_unused_inputs` is more explicit, but a bit long.
  - should this be in the `InterconnectedSystem` constructor, rather than in `interconnect` ?

### DIFF
--- a/control/iosys.py
+++ b/control/iosys.py
@@ -2306,20 +2306,10 @@ def interconnect(syslist, connections=None, inplist=[], outlist=[],
 
     check_unused : bool
         If True, check for unused sub-system signals.  This check is
-        not done if connections is False, and not input and output
+        not done if connections is False, and neither input nor output
         mappings are specified.
 
     ignore_inputs : list of input-spec
-
-        A list of sub-system known not to be connected.  This is
-        *only* used in checking for unused signals, and does not
-        disable use of the input.
-
-        Besides the usual input-spec forms (see `connections`), an
-        input-spec can be a string give just the signal name, as for inpu
-
-    ignore_inputs : list of input-spec
-
         A list of sub-system inputs known not to be connected.  This is
         *only* used in checking for unused signals, and does not
         disable use of the input.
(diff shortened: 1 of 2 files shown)


murrayrm:
Quick responses on a few things, based on a quick read:

* I really like this functionality and completely agree we should include it!

* I think it would be OK to put this in the `interconnect()` function rather than the `InterconnectedSystem` constructor.  The intent of `interconnect()` was to provide some additional processing and a more user-friendly interface.

* It looks like there are some glitches in the documentation (`ignore_inputs` appears twice).

roryyorke:
Thanks for the feedback.  I've fixed the docstring.

>  think it would be OK to put this in the interconnect()

OK, good.  If `InterconnectedSystem` is used directly, `check_unused_signals` is available to be called after mappings have been set up.  I also made `unused_signals` public, though I don't know what applications it could have.

While writing this, it occurred to me it wouldn't be all *that* hard to output a [Graphviz](https://graphviz.org/) dot file, or, possibly, a data structure that could be fed into [NetworkX](https://networkx.org/), to allow respectively visualization and graph-theoretic analysis of the signal flow, but I don't have time for any of that at the moment.

sawyerbfuller:
Connecting @petercorke into this discussion because he is working on a nice-looking graphical renderer for this kind of interconnection. Not sure it is released yet though.

petercorke:
It’s kind of released, just not publicised at the moment.  If you pip install bdsim, then it is available as bdedit.py

PR #662, opened 2021-11-02
new Pid design function built on sisotool

sawyerbfuller:
Provides a manual means of tuning a PID controller but does not attempt an automatic solution, thus its name `pid_designer`. Edit: renamed function to `rootlocus_pid_designer` to distinguish it from loopshaping techniques. Partially addresses #657. 

Uses `Sisotool` to investigate the effect of adding or subtracting an
amount `deltaK` to the proportional, integral, or derivative (PID) gains of
a controller. One of the PID gains, `Kp`, `Ki`, or `Kd`, respectively, can
be modified at a time. `Sisotool` plots the step response, frequency
response, and root locus.

When first run, `deltaK` is set to 0; click on a branch of the root locus
plot to try a different value. Each click updates plots and prints
the corresponding `deltaK`. To tune all three PID gains, repeatedly call
`pid_designer`, and select a different `gain` each time (`'P'`, `'I'`,
or `'D'`). Make sure to add the resulting `deltaK` to your chosen initial
gain on the next iteration.

### DIFF
--- a/control/tests/sisotool_test.py
+++ b/control/tests/sisotool_test.py
@@ -6,11 +6,11 @@ import numpy as np
 from numpy.testing import assert_array_almost_equal
 import pytest
 
-from control.sisotool import sisotool
+from control.sisotool import sisotool, pid_designer
 from control.rlocus import _RLClickDispatcher
 from control.xferfcn import TransferFunction
 from control.statesp import StateSpace
-
+from control import c2d
 
 @pytest.mark.usefixtures("mplcleanup")
 class TestSisotool:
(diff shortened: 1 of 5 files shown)


control/tests/sisotool_test.py
```diff
@@ -140,3 +140,25 @@ def test_sisotool_mimo(self,  sys222, sys221):
         # but 2 input, 1 output should
         with pytest.raises(ControlMIMONotImplemented):
             sisotool(sys221)
+
+@pytest.mark.usefixtures("mplcleanup")
+class TestPidDesigner:
+    syscont = TransferFunction(1,[1, 3, 0])
+    sysdisc1 = c2d(TransferFunction(1,[1, 3, 0]), .1)
+    syscont221 = StateSpace([[-.3, 0],[1,0]],[[-1,],[.1,]], [0, -.3], 0)
```
bnavigator, line 148: I think the current CI failure is due to this code.

You have to hide the system creation in a fixture. Calling those constructors in the class definition means constructing the systems during test collection, where the matrix/array switch is not set up yet.

sawyerbfuller: I can't seem to get it to work. When I do eg. this: 

``` python
class TestPidDesigner:
   @pytest.fixture
    def syscont(self):
        return TransferFunction(1,[1, 3, 0])

    @pytest.fixture
    def sysdisc1(self):
        return c2d(TransferFunction(1,[1, 3, 0]), .1)

    # cont or discrete, vary P I or D
    @pytest.mark.parametrize('plant', (syscont, sysdisc1))
    @pytest.mark.parametrize('gain', ('P', 'I', 'D'))
    @pytest.mark.parametrize("kwargs", [{'Kp0':0.01},])
    def test_pid_designer_1(self, plant, gain, kwargs):
        pid_designer(plant, gain, **kwargs)```
The `pid_designer` function does not get the actual transfer function, it gets the fixture instead. 
This also raises an error because 'plant' is not found: `    @pytest.mark.parametrize('plant', ('syscont', 'syscont221'), indirect=True)`
Any idea how I give it the actual system?

bnavigator: Like `tsystem` in `timeresp_test.py`:

```python
class TestPidDesigner:
    @pytest.fixture
    def plant(self, request):
        plants = {
            'syscont': TransferFunction(1,[1, 3, 0]),
            'sysdisc1': c2d(TransferFunction(1,[1, 3, 0]), .1)
        }
        return plants[request.param]

    # cont or discrete, vary P I or D
    @pytest.mark.parametrize('plant', ('syscont', 'sysdisc1'), indirect=True)
    @pytest.mark.parametrize('gain', ('P', 'I', 'D'))
    @pytest.mark.parametrize("kwargs", [{'Kp0':0.01},])
    def test_pid_designer_1(self, plant, gain, kwargs):
        pid_designer(plant, gain, **kwargs)

sawyerbfuller: That did it, thanks.

bnavigator:
Same CI failure as in #659. #663 might work around it.

murrayrm:
FYI: this is generating a bunch of warning messages.  We should clean up when we get a chance.

PR #668, opened 2021-11-05
modify discrete-time contour for nyquist plots to indent around poles

sawyerbfuller:
Completing a “todo”  that was in the code, to make nyquist plots work better foe discrete-time systems (Continuous-time indentation was already in place )


bnavigator (approved):
LGTM

PR #669, opened 2021-11-05
fixed prewarp not working in c2d and sample_system, margin docstring improvements

sawyerbfuller:
also, fixed some typos and changed from `NotImplementedError` to `ControlMIMONotImplemented` error where needed.

### DIFF
--- a/control/dtime.py
+++ b/control/dtime.py
@@ -89,7 +89,8 @@ def sample_system(sysc, Ts, method='zoh', alpha=None, prewarp_frequency=None):
     if not isctime(sysc):
         raise ValueError("First argument must be continuous time system")
 
-    return sysc.sample(Ts, method, alpha, prewarp_frequency)
+    return sysc.sample(Ts,
+        method=method, alpha=alpha, prewarp_frequency=prewarp_frequency)
 
 
 def c2d(sysc, Ts, method='zoh', prewarp_frequency=None):
@@ -126,6 +127,7 @@ def c2d(sysc, Ts, method='zoh', prewarp_frequency=None):
     """
 
     #  Call the sample_system() function to do the work
-    sysd = sample_system(sysc, Ts, method, prewarp_frequency)
+    sysd = sample_system(sysc, Ts,
+        method=method, prewarp_frequency=prewarp_frequency)
 
     return sysd
--- a/control/dtime.py
+++ b/control/dtime.py
@@ -5,6 +5,7 @@ Functions for manipulating discrete time systems.
 Routines in this module:
 
 sample_system()
+c2d()
 """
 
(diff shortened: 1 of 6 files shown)


control/margins.py
```diff
@@ -522,10 +522,10 @@ def margin(*args):
         Gain margin
     pm : float
         Phase margin (in degrees)
-    wpc : float or array_like
-        Phase crossover frequency (where phase crosses -180 degrees)
     wgc : float or array_like
         Gain crossover frequency (where gain crosses 1)
+    wpc : float or array_like
+        Phase crossover frequency (where phase crosses -180 degrees)
```
bnavigator, line 528: Are you sure?

If you compare with the [MATLAB function](https://www.mathworks.com/help/control/ref/lti.margin.html):  

```matlab
[Gm,Pm,Wcg,Wcp] = margin(sys)
```
`Wcg`, the  frequency corresponding to the gain margin, is the frequency, where the phase crosses 180°, and `Wcp`, the frequency for the phase margin, is the gain crossover frequency. We just have differently named variables but use the same order here and in `stability_margins`.

sawyerbfuller: I think you're right. I'll revert that part of this PR and adjust those names to correspond to MATLAB convention

control/dtime.py
```diff
@@ -89,7 +89,8 @@ def sample_system(sysc, Ts, method='zoh', alpha=None, prewarp_frequency=None):
     if not isctime(sysc):
         raise ValueError("First argument must be continuous time system")
 
-    return sysc.sample(Ts, method, alpha, prewarp_frequency)
+    return sysc.sample(Ts,
+        method=method, alpha=alpha, prewarp_frequency=prewarp_frequency)
```
bnavigator, line 93: `alpha` is missing in the docstring.

control/dtime.py
```diff
@@ -126,6 +127,7 @@ def c2d(sysc, Ts, method='zoh', prewarp_frequency=None):
     """
 
     #  Call the sample_system() function to do the work
-    sysd = sample_system(sysc, Ts, method, prewarp_frequency)
+    sysd = sample_system(sysc, Ts,
+        method=method, prewarp_frequency=prewarp_frequency)
```
bnavigator, line 131: So this has not been caught by the unit tests, right? Should we add a test which calls `c2d` with nonstandard kwargs? Maybe parametrize `test_sample_system_prewarp` with `[plant.sample, c2d, sample_system]`

sawyerbfuller: good idea; done.

control/dtime.py
sawyerbfuller: done

PR #671, opened 2021-11-21
in nyquist plots, draw contour at specified magnitude if threshold is exceeded

sawyerbfuller:
This pr was motivated by a feedback system in which the default indent radius resulted in a very small, confusing contour. In this PR, the maximum radius is specified explicitly. This also helps visualize -1 point because the nyquist plot has a standard size relative to that point. To do so, the default indent radius is reduced and an extra contour is drawn when the regular contour leaves the plot window.  

Before: 
image
image
image
image (dt system)
After:
image
image
image
image(dt system)

### DIFF
--- a/control/freqplot.py
+++ b/control/freqplot.py
@@ -522,8 +522,8 @@ _nyquist_defaults = {
     'nyquist.mirror_style': '--',
     'nyquist.arrows': 2,
     'nyquist.arrow_size': 8,
-    'nyquist.indent_radius': 1e-6,
-    'nyquist.infinity_radius': 10,
+    'nyquist.indent_radius': 1e-3,
+    'nyquist.maximum_magnitude': 5,
     'nyquist.indent_direction': 'right',
 }
 
@@ -594,8 +594,9 @@ def nyquist_plot(syslist, omega=None, plot=True, omega_limits=None,
     arrow_style : matplotlib.patches.ArrowStyle
         Define style used for Nyquist curve arrows (overrides `arrow_size`).
 
-    infinity_radius : float
-
+    maximum_magnitude : float
+        Magnitude/radius at which to draw contours whose magnitude exceeds 
+        this value. 
 
     indent_radius : float
         Amount to indent the Nyquist contour around poles that are at or near
@@ -684,8 +685,8 @@ def nyquist_plot(syslist, omega=None, plot=True, omega_limits=None,
         'nyquist', 'indent_radius', kwargs, _nyquist_defaults, pop=True)
     indent_direction = config._get_param(
         'nyquist', 'indent_direction', kwargs, _nyquist_defaults, pop=True)
-    infinity_radius = config._get_param(
-        'nyquist', 'infinity_radius', kwargs, _nyquist_defaults, pop=True)
+    maximum_magnitude = config._get_param(
+        'nyquist', 'maximum_magnitude', kwargs, _nyquist_defaults, pop=True)
 
     # If argument was a singleton, turn it into a list
     if not hasattr(syslist, '__iter__'):
@@ -771,13 +772,13 @@ def nyquist_plot(syslist, omega=None, plot=True, omega_limits=None,
         # Compute the primary curve
         resp = sys(contour)
 
-        # compute infinity radius contour
+        # compute large radius contour where it exceeds 
         # fill with nans that don't plot
-        infinity_contour = np.full_like(contour, np.nan + 1j * np.nan)
+        large_mag_contour = np.full_like(contour, np.nan + 1j * np.nan)
         # where too big, use angle of resp but specifified mag
-        too_big = abs(resp) > infinity_radius
-        infinity_contour[too_big] = \
-            infinity_radius *(1+isys/40) * np.exp(1j * np.angle(resp[too_big]))
+        too_big = abs(resp) > maximum_magnitude
+        large_mag_contour[too_big] = \
+            maximum_magnitude *(1+isys/20) * np.exp(1j * np.angle(resp[too_big]))
         
         # Compute CW encirclements of -1 by integrating the (unwrapped) angle
         phase = -unwrap(np.angle(resp + 1))
@@ -806,7 +807,7 @@ def nyquist_plot(syslist, omega=None, plot=True, omega_limits=None,
 
             # Save the components of the response
             x, y = resp.real, resp.imag
-            x_inf, y_inf = infinity_contour.real, infinity_contour.imag
+            x_lg, y_lg = large_mag_contour.real, large_mag_contour.imag
 
             # Plot the primary curve
             p = plt.plot(x, y, '-', color=color, *args, **kwargs)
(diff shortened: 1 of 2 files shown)


murrayrm:
Nice!  Will try to look through the code in the coming days and leave any comments.

control/freqplot.py
```diff
@@ -838,6 +861,8 @@ def nyquist_plot(syslist, omega=None, plot=True, omega_limits=None,
         ax.set_xlabel("Real axis")
         ax.set_ylabel("Imaginary axis")
         ax.grid(color="lightgray")
+        ax.set_xlim(-maximum_magnitude*1.1, maximum_magnitude*1.1)
+        ax.set_ylim(-maximum_magnitude*1.1, maximum_magnitude*1.1)
```
bnavigator, line 865: The plot limits are really tied to the `maximum_magnitude` now. Even for systems like this:

```python
from control import TransferFunction, nyquist_plot
G = TransferFunction([1],[500, 500, 0])
nyquist_plot(G)
```

| Before | After |
| --------- | ------ |
| Figure_1_before | Figure_1 |



Is that desired?

bnavigator (commented):
Nice idea!

control/freqplot.py
sawyerbfuller: I think (though I am not positive) that @bnavigator's system might be a pathological case because you are mostly concerned with systems cascaded with their controllers, in which case you won't have a very low loop gain as this system does. 

That said, I rewrote the indentation code to add a consistent number of points at every indent, and further reduced the indent radius, which make it so everything but the most extremely low-gain systems have large contours that go outside of the plot window. This also helps with other pathological cases. 

also: 
* removed an indent test that confirms that no indentation happens when first omega is too large  
* made the red lines have the same line style as their associated contour. 
Updated after plots: 
image
image
And @bnavigator's example: 
image

System with poles not at origin: 
Before: 
image

After: 
image

sawyerbfuller:
Further updates and now ready to merge of ok'd:  
* rename `maximum_magnitude` to `plot_maximum_magitude`. Option to turn off this plot by setting its value to 0. 
* docstring description of behavior
* if and only if `omega` is specified, the old approach to indentation is used, otherwise indentation points are added.

control/freqplot.py
```diff
@@ -522,21 +522,23 @@ def gen_zero_centered_series(val_min, val_max, period):
...
+                 return_contour=False, warn_nyquist=True, 
+                 *args, **kwargs):
     """Nyquist plot for a system
 
     Plots a Nyquist plot for the system over a (optional) frequency range.
     The curve is computed by evaluating the Nyqist segment along the positive
     imaginary axis, with a mirror image generated to reflect the negative
     imaginary axis.  Poles on or near the imaginary axis are avoided using a
-    small indentation.  The portion of the Nyquist contour at infinity is not
+    small indentation. If portions of the Nyquist plot  The portion of the Nyquist contour at infinity is not
```
bnavigator, line 541: Some unintended text insertion here, I guess.

control/freqplot.py
```diff
@@ -550,7 +552,9 @@ def nyquist_plot(syslist, omega=None, plot=True, omega_limits=None,
         If True, plot magnitude
 
     omega : array_like
-        Set of frequencies to be evaluated, in rad/sec.
+        Set of frequencies to be evaluated, in rad/sec. Remark: if omega is
+        specified, you may need to specify specify a larger `indent_radius`
```
bnavigator, line 556: double specify

control/freqplot.py
bnavigator: Okay, bad example.

Let's consider a system without undamped resonance:

```python
In [1]: from control import TransferFunction, nyquist_plot
   ...: G = TransferFunction([0.5],[1, 1, 1])
   ...: nyquist_plot(G)
```

| before | after |
| --- | ---|
| nyquist_before | nyquist_after |

sawyerbfuller: Do you think this would be resolved by adding a test for whether there was any `maximum_magintude` contour drawn, and if there wasn't, then to let it zoom the plot by the default amount?

bnavigator: Sounds reasonable.

murrayrm:
I've been playing around with an example that did not work quite right in the original code because of the way that the points were generated along the frequency axis (basically points weren't dense enough, so you could miss encirclements around unstable poles at non-zero frequencies).  It looks like the new version of the code has some similar problems as well as introducing some new issues that I am not sure I understand yet.

Here is an example (derived from something more complicated) that works with the previous version but does not work with this version:
```
import numpy as np
import control as ct
import control.matlab as matlab

sys = ct.ss(ct.tf(*matlab.zpk2tf(
    [-3e-5 + 3j, -3e-5 - 3j], 
    [0, 1e-4 + 2j, 1e-4 - 2j, -1e-2], 
    1)))
K, S, E = ct.lqr(sys, np.eye(sys.nstates), 1)
L, P, E = ct.lqe(sys.A, sys.B, sys.C, 1, 1)
ctrl = ct.ss(sys.A - L @ sys.C - sys.B @ K, L, K, 0)
ct.nyquist_plot(sys * ctrl)
```
This code creates a process `sys` that has a pole at the origin and two unstable poles at `1e-4 +/- 2j` and then creates a stabilizing controller using LQR with an estimator.  Since the original system is unstable but the closed loop is stable (easy to check), the loop transfer function should have two CCW encirclements of the -1 point, so `nyquist_plot()` should return -2 (clockwise encirclements).

Running the original code yields the following plot and a return value of -2:
image

Running the new code yields the following plot and a return value of 0 (which is not correct):
image

In addition to the wrong contour and the wrong count of encirclements, there is something going on with the arrows that needs to be fixed, I think.

I have a more complicated example where you have to be very careful with the `indent_radius` in order to get the right answer.  I want to test that example as well, but we should probably find fix the errors above first.  (We should also add some unit tests that capture the incorrect encirclement count bug.)

sawyerbfuller:
@murrayrm what you suggest sounds like a reasonable explanation to me. 

So this is a second, unrelated reason why it might be nice to have automatic frequency vector generation code that adds more points near poles (the other being that then the shape of the bode plot can be better visualized). 

I have no idea what is going on with those arrows but it may be related to having discontinuous points

murrayrm:
A few additional comments after playing around with this a bit:

* Is there a way to have multiple overlapping extra contours offset a bit so that you can see that there are multiple encirclements?  The "system with poles not at origin" example above is one case showing why this could be useful (in the original diagram you can see two distinct curves, but in the modified version these curves overlap in a way that is non-obvious).

* Rather than setting `plot_maximum_magnitude` to zero to turn off the functionality, perhaps set it to `None`?

* I feel like the default value for `plot_maximum_magnitude` should be `None` so that you get a "regular" Nyquist plot by default, since otherwise people might be confused by this non-standard representation.

Also, I've put some preliminary code that does a better job at adding points near poles [here](https://github.com/murrayrm/python-control/tree/enh-nyquist).  I'm thinking of pulling that code into the `_determine_omega_vector` function in `freqplot.py`, in which case it would also provide more resolution in Bode plots.

sawyerbfuller:
@murrayrm re mulitple contours: makes sense, good idea. 

re turning off `plot_maximum_magintude` plot: I tried using `None` first but got stuck because I think`config._get_param`  interprets a `None` keyword as meaning that the default should be used. 

re default value for `plot_maximum_magnitude` being off: I agree it may be non-standard, but isn't the previous way also? The difference is the new way allows you to set a reasonable radius to show in response to indentations, rather than have it depend on indent size _and_ the magnitude of the system at that indent. 

One alternative to consider: maybe we could stick with the old way, but size each indent according to the size of the residue (or maybe there is some better formula), so that encirclements are always roughly the same size. You would probably need to further adjust the indent for each pole to avoid overlap with other indentations. 

This is a tricky interface issue.

PR #672, opened 2021-11-21
Additional I/O system type conversions

murrayrm:
This PR adds some type conversations for algebraic operations involving I/O systems, motivated by an example I was putting together where I wanted to generate a Nyquist plot for a linearized nonlinear I/O system (type `LinearIOSystem`) and a simple lead compensator (type `TransferFunction`).  Using `P * C` generated a type conversation error but `C * P` was OK as was `ct.tf(P) * C`.  It seemed annoying to have to do the explicit typecasting, so I decided to implement some of the operations that were needed to make `P * C` work (generates a `LinearICSystem`, which is a subclass of `StateSpace`).

The following functionality has been added:

* Added `__sub__`, so that you can write G1 - G2 for I/O systems
* Added conversions for scalars, arrays, state space, and transfer function systems so that they can be combined with I/O systems on either the left or right.
* Set the Numpy `__array_priority__` value so that operations involving ndarrays are handled as I/O system operations and I/O system operations take precedence over LTI operations involving ndarrays.

I also did some PEP8 cleanup on `iosys.py` while I was at it, and put in unit tests for the code.


murrayrm:
Rebased on top of latest master...

control/iosys.py
```diff
@@ -711,12 +790,16 @@ def __init__(self, linsys, inputs=None, outputs=None, states=None,
         states.  The new system can be a continuous or discrete time system.
 
         """
-        if not isinstance(linsys, StateSpace):
+        if isinstance(linsys, TransferFunction):
+            # Convert system to StateSpace
+            linsys = _convert_to_statespace(linsys)
+
+        elif not isinstance(linsys, StateSpace):
             raise TypeError("Linear I/O system must be a state space object")
```
bnavigator, line 798: ... or transfer function.

control/iosys.py
```diff
@@ -1509,27 +1586,31 @@ def check_unused_signals(self, ignore_inputs=None, ignore_outputs=None):
...
 
         # (isys, isig) -> signal-spec
         ignore_output_map = {}
         for ignore_output in ignore_outputs:
             if isinstance(ignore_output, str) and '.' not in ignore_output:
                 ignore_found = self._find_outputs_by_basename(ignore_output)
                 if not ignore_found:
-                    raise ValueError(f"Couldn't find ignored output {ignore_output} in subsystems")
+                    raise ValueError(f"Couldn't find ignored output "
+                                     "{ignore_output} in subsystems")
```
bnavigator, line 1603: second line needs the f-string prefix

bnavigator (commented):
LGTM

PR #673, opened 2021-11-26
Improved lqe calling functionality 

murrayrm:
This PR updates the `lqe()` function to allow it to be called with a `StateSpace` system as the first argument rather than the matrices that define the system.  This matches the functionality of the `lqr` command.  Docstrings have been updated to reflect the new functionality and to match the notation used for `lqr` a bit better.  Unit tests added to cover the new code.

This PR has some overlap with #670, so depending on which gets merged first it may be necessary to rebase.

### DIFF
--- a/control/statefbk.py
+++ b/control/statefbk.py
@@ -46,7 +46,8 @@ from . import statesp
 from .mateqn import care
 from .statesp import _ssmatrix, _convert_to_statespace
 from .lti import LTI
-from .exception import ControlSlycot, ControlArgument, ControlDimension
+from .exception import ControlSlycot, ControlArgument, ControlDimension, \
+    ControlNotImplemented
 
 # Make sure we have access to the right slycot routines
 try:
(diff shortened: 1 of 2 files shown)


control/statefbk.py
```diff
@@ -326,11 +341,61 @@ def lqe(A, G, C, QN, RN, NN=None):
...
+    if (A.shape[0] != nstates or A.shape[1] != nstates or
+        G.shape[0] != nstates or C.shape[1] != nstates):
+        raise ControlDimension("inconsistent system dimensions")
+
+    elif (Q.shape[0] != ninputs or Q.shape[1] != ninputs or
+          R.shape[0] != noutputs or R.shape[1] != noutputs or
+          N.shape[0] != ninputs or N.shape[1] != noutputs):
+        raise ControlDimension("incorrect covariance matrix dimensions")
+
+    # LT, P, E = lqr(A.T, C.T, G @ Q @ G.T, R)
```
bnavigator, line 396: Is this correct, or should it rather be
```suggestion
    # L, P, E = lqe(...)
```

?

sawyerbfuller: This was a comment (mostly left for myself) that it is also possible to get the LQE gains by using the lqr function as shown in this comment. LT means it's the transpose of the L you want.  Probably best just to remove the comment to avoid confusion.

PR #682, opened 2021-12-12
extrapolation in ufun throwed errors

miroslavfikar:
I came across situation when interpolation of u in ufun was requested beyond the actual time interval. Attached MWE file throws error, proposed modification resolves the issue. There were two possibilities to resolve the issue: either to take the highest value of time known (interpolation) or to prolonge the last interval to the right (extrapolation). I chose the first one to be consistent with already treated case to the left.

[testh2.txt](https://github.com/python-control/python-control/files/7698898/testh2.txt)


murrayrm:
@miroslavfikar It would be great to add a unit test the triggers the error (prior to your fix).

murrayrm:
I updated the fix to use extrapolation, since this is what is supposed to happen.  The example now runs without throwing and error.  (I couldn't figure out how to create a simple unit test since it seem to depend on the integrator asking to sample outside of the time range.)

@miroslavfikar It would be useful if you can confirm that the new version gives a correct result for your example.

miroslavfikar:
@murrayrm Yes, I confirm, I get the correct results with your commit, thanks.

PR #683, opened 2021-12-24
LQR using SciPy

murrayrm:
This PR updates the `mateqn` module to allow the use of SciPy linear algebra functions as a replacement for Slycot functions.  Among other things, this allows the use of the `lqr` command without having to install Slycot.

I implemented this by changing all of the `mateqn` functions (`lyap`, `care`, `dare`, etc) to accept at `method` keyword that can either be `scipy` or `slycot`.  The default value for `method` is `slycot` if it is installed, otherwise `scipy`, so there is no change in behavior if you have Slycot. 

Other changes:
* Updated `lqr` to call `care` rather than replicating the code and cleaned up a lot of the code in `mateqn` to be more Pythonic.
* Modified the `slycot_check` function so that it stores the result of checking whether `slycot` can be imported rather than trying to import every time the function is called.
* Implemented a function for checking size and properties of matrices passed to `mateqn` functions (`mateqn._check_shape`) that provides more consistent errors (this also required updating some of the unit tests).
* Modified `dare` to solve the generalized Sylvester equation using `slycot.sg02ad` rather than using `slycot.sg02md` to solve the simplified form of the discrete algebraic Riccati equation because there seems to be an error in the way that `stabilizing` is handled in `sg02md` (it returns the closed loop eigenvalues in the wrong order).  (I'm working on generating a concrete counterexample for posting to Slycot.)

### DIFF
--- a/control/mateqn.py
+++ b/control/mateqn.py
@@ -37,10 +37,12 @@
 
 import warnings
 import numpy as np
-from numpy import shape, size, asarray, copy, zeros, eye, \
-    finfo, inexact, atleast_2d
+from numpy import shape, size, copy, zeros, eye, finfo, inexact, atleast_2d
+
+import scipy as sp
 from scipy.linalg import eigvals, solve_discrete_are, solve
-from .exception import ControlSlycot, ControlArgument
+
+from .exception import ControlSlycot, ControlArgument, slycot_check
 from .statesp import _ssmatrix
 
 # Make sure we have access to the right slycot routines
@@ -84,7 +86,7 @@ __all__ = ['lyap', 'dlyap', 'dare', 'care']
 #
 
 
-def lyap(A, Q, C=None, E=None):
+def lyap(A, Q, C=None, E=None, method=None):
     """X = lyap(A, Q) solves the continuous-time Lyapunov equation
 
         :math:`A X + X A^T + Q = 0`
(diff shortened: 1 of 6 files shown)


control/mateqn.py
```diff
@@ -411,77 +392,59 @@ def care(A, B, Q, R=None, S=None, E=None, stabilizing=True):
...
-        if (size(B) > 1 and shape(B)[0] != n) or \
-           size(B) == 1 and n > 1:
-            raise ControlArgument("Incompatible dimensions of B matrix.")
-
-        if not _is_symmetric(Q):
-            raise ControlArgument("Q must be a symmetric matrix.")
+        # See if we should solve this using SciPy
+        if method == 'scipy':
+            if not stabilizing:
+                raise ValueError(
```
bnavigator, line 441: Is this a `ValueError` or a `ControlArgument`?

control/mateqn.py
```diff
@@ -623,191 +557,150 @@ def dare(A, B, Q, R, S=None, E=None, stabilizing=True):
...
+    if method == 'scipy' and not stabilizing:
+        raise ValueError(
+            "method='scipy' not valid when stabilizing is not True")
+
+    elif method == 'slycot':
+        return _dare_slycot(A, B, Q, R, S, E, stabilizing)
+
     else:
+        _check_shape("B", B, n, m)
+        _check_shape("Q", Q, n, n, square=True, symmetric=True)
```
bnavigator, line 590: Maybe `_dare_slycot()` and `_dare_scipy()` should be on the same level?

bnavigator (commented):
Looks good. A few minor comments:

control/mateqn.py
murrayrm: I agree the asymmetry is annoying.  I left it this way to avoid changing code that was working.  In `care` everything is in the main function, so at some point we should clean all of these up to be more uniform (either a single function or use `_method` uniformly).

PR #684, opened 2021-12-27
Fix warning messages in rootlocus_pid_designer unit tests

murrayrm:
This PR gets rid of warning messages introduced in PR #662:

* Since `rootlocus_pid_designer` leaves either `r` or `d` disconnected, the `interconnect` function was generating warnings (76 of them...).  This was fixed by using `check_unused=False`.
* Calling `rootlocus_pid_designer` with `plot=True` seems to generate a Bode plot that is -1 and this creates a warning:
```
control/ctrlutil.py:75: RuntimeWarning: invalid value encountered in remainder
  dangle_desired = (dangle + period/2.) % period - period/2.
```
For now this latter problem is fixed by skipping this test.

@sawyerbfuller I'm not quite sure how to fix this error.  If I try running the test by hand using
```
ct.rootlocus_pid_designer(syscont, input_signal='r', Kp0=0.01,
    derivative_in_feedback_path=True, gain='P')
```
then I get the following plot:

Sisotool_pid_designer_bug

### DIFF
--- a/control/sisotool.py
+++ b/control/sisotool.py
@@ -328,7 +328,7 @@ def rootlocus_pid_designer(plant, gain='P', sign=+1, input_signal='r',
     loop = interconnect((plant, Kpgain, Kigain, Kdgain, prop, integ, deriv,
                             C_ff, e_summer, u_summer),
                             inplist=['input', input_signal],
-                            outlist=['output', 'y'])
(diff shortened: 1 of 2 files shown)


murrayrm:
Merging this for 0.9.1, but will create an issue to track the possible bug above.

PR #693, opened 2022-01-19
added binder link

jonititan:


sawyerbfuller:
I thought this PR was pretty opaque but then I found this similar PR in slycot: https://github.com/python-control/Slycot/pull/168. Anybody else here have experience with this service? Seems on the surface to be an easy way to try things out.

sawyerbfuller:
Did a little bit of background research - Fernando Perez' (creator of iPython/Jupyter) has a link to the Binder service on a [page](https://jupytearth.org/jupyter-resources/ecosystem/binder.html) he runs for a grant from the NSF he got in part to fund development of Jupyter, so it seems legit to me. 

Proposed location for the link in the python-control docs seems reasonable as a way to 'try' the library.  

I did a brief test and it works: I was able to quickly plot a step response. It appears that slycot does not currently work. Given that there is still a lot of functionality in the library even without slycot, and that this feature seems useful enough, I've merged it. Thanks @jonititan !

Issue #692, opened 2022-01-19
Interconnect treats string as list


murrayrm:
If you pass the `interconnect` function a string instead of a list, it treats the string as a list (of characters).  For example:
```
P1 = ct.rss(2, 1, 1)
P1_iosys = ct.LinearIOSystem(P1, inputs='u1', outputs='y1')
P2 = ct.rss(2, 1, 1)
P2_iosys = ct.LinearIOSystem(P2, inputs='y1', outputs='y2')
P_series = ct.interconnect([P1_iosys, P2_iosys], inputs=['u1'], outputs=('y2'))
```
generates the following error:
```
ValueError: could not find signal y

```
The problem appears to be that `('y2')` is the same as `'y2'` in Python and is being interpreted as `['y', '2']`.  Changing this to `outputs=('y2',)` or `outputs=['y2']` fixes the problem, but it is still a bug.

roryyorke:
This is easy enough to fix (I hope!), see PR.

Two remarks: first, this description of the `inputs` arg mentions an option of an `integer count`, which I found difficult to understand, and which also doesn't seem to work:
```
    inputs : int, list of str or None, optional
        Description of the system inputs.  This can be given as an integer
        count or as a list of strings that name the individual signals.  If an
        integer count is specified, the names of the signal will be of the
        form `s[i]` (where `s` is one of `u`, `y`, or `x`).  If this parameter
        is not given or given as `None`, the relevant quantity will be
        determined when possible based on other information provided to
        functions using the system.
```

My test case:

```ipython
In [26]: P1 = ct.LinearIOSystem(ct.rss(2,1,1), inputs='u1', outputs='y1')
    ...: P2 = ct.LinearIOSystem(ct.rss(2,1,1), inputs='y1', outputs='y2')
    ...: ct.interconnect([P1, P2], inputs=1)
    ...: 
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-26-ca5ff6a9552e> in <module>
      1 P1 = ct.LinearIOSystem(ct.rss(2,1,1), inputs='u1', outputs='y1')
      2 P2 = ct.LinearIOSystem(ct.rss(2,1,1), inputs='y1', outputs='y2')
----> 3 ct.interconnect([P1, P2], inputs=1)

[...]

PR #698, opened 2022-01-29
Interpret str-type args to `interconnect` as non-sequence

roryyorke:
If the arguments `inputs` or `outputs` (or their aliases `input` or
`output`) are of type str, treat as a list containing that string.

Fixes gh-692.


bnavigator (approved):
LGTM

PR #699, opened 2022-01-29
Change InputOutputSystem repr to show type, name, inputs, and outputs

roryyorke:
I didn't know about `InputOutputSystem` 's `__str__` implementation, and I find the `__repr__` a bit bare-bones.  With this change:

```ipython
In [8]: import control as ct
   ...: P1 = ct.LinearIOSystem(ct.rss(2,1,1), inputs='u1', outputs='y1')
   ...: P2 = ct.LinearIOSystem(ct.rss(2,1,1), inputs='y1', outputs='y2')
   ...: 

In [9]: display(P1)
   ...: 
<LinearIOSystem:sys[6]:['u1']->['y1']>

In [10]: ct.interconnect([P1, P2], inputs=['u1'], outputs=['y2'])
    ...: 
Out[10]: <LinearICSystem:y2:['u1']->['y2']>

In [11]: display(ct.InputOutputSystem())
    ...: 
<InputOutputSystem:sys[9]:[]->[]>
```

This is a draft for comment - if it looks OK, I'll (eventually) improve it and add tests.  One improvement I can see already: make the input and output lists a little prettier by removing the quotes.

### DIFF
--- a/control/iosys.py
+++ b/control/iosys.py
@@ -181,7 +181,8 @@ class InputOutputSystem(object):
     nstates = 0
 
     def __repr__(self):
-        return self.name if self.name is not None else str(type(self))
+        return f'<{self.__class__.__name__}:{self.name}:{list(self.input_index)}->{list(self.output_index)}>'
+
 
     def __str__(self):
         """String representation of an input/output system"""


murrayrm:
I agree that a change would be useful here.  More generally, it would be nice to think through what we want to have displayed in various situations:

* `__repr__`: this is the default display for objects in iPython/Jupyter and is supposed to be something that could, in principle, be used to recreate the object.  For `StateSpace` systems, for example, this currently gives
```
StateSpace(array([[ 0.,  1.],
       [-1., -2.]]), array([[0.],
       [1.]]), array([[1., 0.]]), array([[0.]]))
```

* `__str__`: this is what you get when you type `print(sys)` and it gives a more verbose output.  The current output for a state space systems is:
```
A = [[ 0.  1.]
     [-1. -2.]]

B = [[0.]
     [1.]]

C = [[1. 0.]]

D = [[0.]]
```

* After PR #710 gets merged, creating a system with `ss` will produce a `LinearIOSystem`, which has a different `__str___` representation:
```
Object: sys[0]
Inputs (1): u[0], 
Outputs (1): y[0], 
States (2): x[0], x[1], 

A = [[ 0.  1.]
     [-1. -2.]]

B = [[0.]
     [1.]]

C = [[1. 0.]]

D = [[0.]]

```

* We also have a `_repr_latex_` method, which gets used in Jupyter notebooks.  This produces the following output:
Screen Shot 2022-03-18 at 1 11 05 PM

Finally, PR #710 has a conflicting change since I reset the __repr__ format there to include the type and system name.  So an `InputOutputSystem` object has output:
```
<class 'control.iosys.LinearIOSystem'>: sys[0]
```
I like the version in this draft PR better, but note that if you get something with lots of inputs and outputs, it could be a bit odd looking.

murrayrm:
PR #721 has a portion of this (`__repr__`) implemented, but leaves the other issues touched on above unchanged.  It would be nice to sort out a consistent set of representations that make sense in different environments (eg, Jupyter vs command line).

henklaak:
FYI, Spyder (a nifty little IDE) also uses the _repr_latex_() output

Not sure if it's the underlying IPython that does this magic, but it sure is handy to get your SS in a compact matrix in such editors. 

image

murrayrm:
This has been implemented since v0.10.0.

PR #710, opened 2022-03-16
I/O system enhancements

murrayrm:
This PR adds some new functionality for I/O systems that I used in a recent course on optimization-based control that I taught at Caltech:

* Modify the `ss()`, `rss()`, and `drss()` functions to return `LinearIOSystem` objects (instead of `StateSpace` objects).  This makes it easier to create LTI state space systems that can be combined with other I/O systems without having to add a conversation step.  Since `LinearIOSystem` objects are also `StateSpace` objects, no functionality is lost.  (This change is implemented through the introduction of a internal `NamedIOSystem` class, to avoid import cycles.)

* Added a new function `create_statefbk_iosystem()` that creates an I/O system for implementing a linear state feedback controller of the form u = ud - Kp(x - xd).  The function returns an I/O system that takes xd, ud, and x as inputs and generates u as an output.  The `integral_action` keyword can be used to define a set of outputs y = C x for which integral feedback is also included: u = ud - Kp(x - xd) - Ki(C x - C xd).

* The `lqr` and `dlqr` commands now accept an `integral_action` keyword that allows outputs to be specified for implementing integral action.  The resulting gain matrix has the form K = [Kp, Ki].  (This is useful for combining with the `integral_action` functionality in `create_statefbk_iosystem()`).

### DIFF
--- a/control/frdata.py
+++ b/control/frdata.py
@@ -204,7 +204,7 @@ class FrequencyResponseData(LTI):
                         w=1.0/(absolute(self.fresp[i, j, :]) + 0.001), s=0.0)
         else:
             self.ifunc = None
-        LTI.__init__(self, self.fresp.shape[1], self.fresp.shape[0])
+        super().__init__(self.fresp.shape[1], self.fresp.shape[0])
 
     def __str__(self):
         """String representation of the transfer function."""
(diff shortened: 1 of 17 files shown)


murrayrm:
I have some changes that build on this one, so I'm accepting the PR.  If you have comments on anything here, feel free to post and I can update with another PR to address.

sawyerbfuller:
Am a little late on this, but it looks good to me.

Issue #528, opened 2021-01-31
Add code to check for unrecognized keywords associated with default settings


murrayrm:
As pointed out in #526, it is possible to pass unrecognized keywords to `StateSpace` and not get an error message.  For example:
```
ct.StateSpace(A, B, C, D, remove_useless=True)
```
(the actual keyword should be `remove_useless_states`).

Similarly,
```
ct.set_defaults('statesp', remove_useless=True)
```
should generate an error.

murrayrm:
There is some code to help carry this out in PR #549 (`ct.config._get_param` has a new keyword `last`).

Issue #702, opened 2022-02-18
Set of constraints raises Error


ThomasGiavasopoulos:
I am using the python optimal control library to optimise the trajectory of a 2D rocket. Please find attached a minimum reproducible sample code of the optimisation process of the problem. Whenever I am trying to set constraints for only some of the states, I get the error described after the minimum reproducible example. Is there an appropriate way to define multiple separate constraints passed in the "constraints= [..., ..., ...]" argument of the solve.ocp() function? Also, instead of using range -np.inf to np.inf for state parameters that I do not want to bound, is there a value that makes the code ignore the specific state parameter constraints?

```
T_max_Single = 7607000/9

g = 9.81 
m = 114239.00
k = 18.74
I = 16071705.59

import numpy as np
import math
import control as ct
import control.optimal as opt
import matplotlib.pyplot as plt
import logging
import time
import os
plt.style.use("default")
from IPython.display import display
from colorama import *
import sympy as s
from scipy.integrate import odeint


#Definition of the states
def rocket_update(t, x, u, params):
        
    xpos = x[0] 
    U = x[1] 
    z = x[2] 
    W = x[3] 
    θ = x[4] 
    q = x[5] 
    
    T = u[0]
    a = u[1]

    # Return the derivative of the state
    dydt = np.array([U,
            T*np.cos(θ-a)/m, 
            W,
            T*np.sin(θ-a)/m-g,
            q,
            T*k*np.sin(a)/I])
    
    return(dydt)

def rocket_output(t, x, u, params):
    return x                            # return (full state)



Tf = 8



# Define the rocket movement dynamics as an input/output system
rocket = ct.NonlinearIOSystem(
    rocket_update, rocket_output, states=6, name='rocket',
    inputs=("T", "α"), outputs=("x", "U", "z", "W", "θ", "q")) 

# Define the time horizon (and spacing) for the optimization
horizon = np.linspace(0, Tf, Tf, endpoint=True)                                        #############IMPORTANT

# Provide an intial guess (will be extended to entire horizon)
bend_left = [T_max_Single, 0.01]          # slight left veer                     #############IMPORTANT


# Optimal Control Problem
Initial_x = 0 #m
Final_x =  5 #m

Initial_z =  0 #m
Final_z = 200 #m

Initial_θ = np.pi/2 #m
Final_θ = np.pi/3 #m

Final_Thrust = T_max_Single

[...]

PR #713, opened 2022-03-23
Keyword argument checking

murrayrm:
This PR implements checks for unrecognized keywords in all functions that accept variable keywords (addressing #528).  This is accomplished using a the `kwargs_test` unit test, which inspects the module, finds all functions with variable keyword arguments, and then makes sure there is a unit test that confirms that sending an unknown keyword generates an error.


murrayrm:
Rebased and changed from draft to ready to merge.  While improvements are possible, this has already caught several bugs (eg, commit 3656968, issue #702) and so I think it is OK to merge in.

PR #721, opened 2022-04-09
Updated system class functionality

murrayrm:
This PR contains a refactoring of the code that is used to implement the I/O and LTI systems classes, building on some of the changes started in PR #710 and adding new functionality (and consistency).  There are quite a few changes, but everything is either new functionality or backward compatible and so existing code does not need to change.

Summary of changes:
* Changed the `LTI` class to use `poles()` and `zeros()` for retrieving poles and zeros, with `pole()` and `zero()` generating a `PendingDeprecationWarning` (which is ignored by default in Python).  (The MATLAB compatibility module still uses `pole()` and `zero()`.)
* The `TimeResponseData` and `FrequencyResponseData` objects now implement a `to_pandas()` method that creates a simple pandas dataframe.
* The `FrequencyResponseData` class is now used as the output for frequency response produced by `freqresp()` and a new function `frequency_response` has been defined, to be consistent with the `input_output_response` function.  A `FrequencyResponseData` object can be assigned to a tuple to provide magnitude, phase, and frequency arrays, mirroring `TimeResponseData` functionality from PR #649.
* The `LTI` class is now a subclass of `NamedIOSystem` (formerly `_NamedIOSystem`), which allows state space and transfer functions systems to have system and signal names.  In addition, the separate `_NamedIOStateSystem` class was eliminated in favor of using `nstates` to distinguish a system with no state space functionality (`nstates == None`, used for `TransferFunction` and `FrequencyResponseData` systems) from a system defined in state space (`nstates >= 0`, used for `InputOutputSystem` and `StateSpace` systems).
* The `__repr__` method for NamedIOSystem has been updated to implement the changes in draft PR #699 and the `__str__` method has been updated for linear I/O system to combine the information for an I/O system and a state space system.  In particular, I/O systems are now displayed as `SystemClass:sysname:[input_labels]->[output_labels]>`.
* Common functionality for the LTI and InputOutputSystem classes is now pulled up into the `NamedIOSystem` class (and `namedio.py`) and a common parsing function `_process_namedio_keywords()` is used for consistent behavior across system creation functions.  
* The `drss`, `rss`, `ss2tf`, `tf2ss`, `tf2io`, and `ss2io` functions now all accept system and signal name arguments (via `_process_namedio_keywords()`.
* * The `rss` function can now accept a `dt` argument and can create either continuous-time or discrete-time random systems.  The `drss` function just calls `rss()` with `dt=True`.
* The `ss` function can now accept function names as arguments, in which case it creates a `NonlinearIOSystem` (I'm not sure how useful this is, but `ss` is a sort of wrapper function that calls the appropriate class constructor, so it was easy enough to implement.)
* The `interconnect` function now has a `warn_duplicate` flag that can be used to turn on/off warning messages about duplicate objects.  The default value is `None`, in which case warnings are suppressed for duplicated systems with `generic` names (of the form `sys[n]`. (The let's you do things like `H = G * G` without getting warning messages for state space systems without specified names.)
* The `SystemTrajectory` class for flat systems now has a `response()` method that generates a `TimeDataResponse` object.
* Added some missing type conversions between systems by using the `__array_priority__` class variable to allow the right operand to override the left operand when needed.  This mainly affect the `FrequencyResponseData` class.
* Updated unit tests, docstrings, user documentation, and examples to match new functionality.

The following diagram (from the user documentation) illustrates the relationship between the various classes:

classes


sawyerbfuller:
This is a nice re-architecting that gives  named signals for tfs and ss systems. 
* having `frequency _response` both a method and a function seems fine. 
* very helpful  inheritance figure
* I like the power of using frd as return type.  is `FRD.return_magphase` Documented?

PR #723, opened 2022-04-16
Improvements to Nichols chart plotting

roryyorke:
These are relatively minor updates to the Nichols chart (AKA grid) plotting.

It's far from ideal; further possible improvements:
   - more tightly fitting the chart to the view (or data) xlim and ylim (the "extent"); the pain here is figuring out which contours to plot, and where to place the labels
     - depending on the limits, some contours will end up with a two visible segments
   - recreating the axis on zoom and pan events
     - this might not be *too* difficult; Axes [support `'xlim_changed'` and `'ylim_changed'`](https://matplotlib.org/stable/api/axes_api.html#the-axes-class) events
   - a deluxe version of this would be too add (secondary?) axes, and show the closed-loop gain and phase in the "data value" text (the "x= <num>, y = <num>" on the Matplotlib toolbar)
   - I wonder if we could use a custom `ContourSet`s, and plug into however that labelling is done?

Here are some before-and-after pictures (left is before) of a few example systems.

The new code gives a better result in almost all respects.  One visible problem is that the closed-loop phase label for "0°" clashes with the "-40dB" (and similar) closed-loop gain label.  The gains become crowded when the gain span is large (final figure), but that is the same in both versions.

I haven't added tests; I'll see how much time I have to do that.  Easyish tests:
   - check for "tight" phase limits (see first example below)
   - check that the axes child text elements are have clip_on (I'd have to not check title  & tick labels, presume this is possible)
   - check for existence of phase labels, contingent on label_cl_phases flag
   - test both branches of the `if` in `_inner_extents`, and in fact test the behaviour in general by setting xlim & ylim before calling nichols_grid.
Please suggest other tests.

example1
example2
example3
example4

## Details (commit message)

Clip closed-loop contour labels.

Add labels for constant closed-loop phase contours.

Use smaller of data or view extents when deciding on how big, in
phase, a chart to create.

Use more uniformly spaced closed-loop phase contours, and use more
widely-spaced contours when phase extent is large.

Add optional `ax` argument, for axes to add grid to.

### DIFF
--- a/control/tests/nichols_test.py
+++ b/control/tests/nichols_test.py
@@ -3,9 +3,11 @@
 RMM, 31 Mar 2011
 """
 
+import matplotlib.pyplot as plt
+
 import pytest
 
-from control import StateSpace, nichols_plot, nichols
+from control import StateSpace, nichols_plot, nichols, nichols_grid, pade, tf
 
 
 @pytest.fixture()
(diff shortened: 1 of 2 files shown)


murrayrm:
Changes look good.  It might be worth adding some unit tests to cover the new features (eg, use of `ax` keyword).

roryyorke:
OK, I'll look at some tests, including testing the `ax` argument.

I'll probably wrap up the linfnorm PR first, and I'm not sure when I'll get this PR done.

roryyorke:
I changed nichols_grid to return the artists it adds; this made testing much easier, but also makes it possible to customize the grid appearance and labels.   Below is a not-especially-compelling example; more realistically one might change the intensity or alpha of the grid and labels.

```python
plt.clf()
mc,nc,ml,nl = ct.nichols_grid()
for i,(c,l) in enumerate(zip(nc,nl)):
     nc.set_color(f'C{i}')
     nl.set_color(f'C{i}')
for i,(c,l) in enumerate(zip(mc,ml)):
     c.set_color(f'C{i}')
     l.set_color(f'C{i}')
```

color-nichols

control/tests/nichols_test.py
```diff
@@ -26,3 +28,73 @@ def test_nichols(tsys, mplcleanup):
 def test_nichols_alias(tsys, mplcleanup):
     """Test the control.nichols alias and the grid=False parameter"""
     nichols(tsys, grid=False)
+
+
+class TestNicholsGrid:
```
bnavigator, line 33: Could you please add the mplcleanup fixture here or for the whole module?

See e.g.
https://github.com/python-control/python-control/blob/46b3c53f752ccc7359083e3347da82b6020c80e0/control/tests/phaseplot_test.py#L21-L22

or

https://github.com/python-control/python-control/blob/46b3c53f752ccc7359083e3347da82b6020c80e0/control/tests/freqresp_test.py#L22

bnavigator: This should also get rid of the necessity for `plt.clf()`

roryyorke: Thanks, that is tidier.

murrayrm:
This looks good to go to me.  @roryyorke: any additional changes that you plan to make?

roryyorke:
No, all done.

PR #729, opened 2022-04-24
ENH: add `linform` to compute linear system L-infinity norm

roryyorke:
linfnorm requires Slycot routine ab13dd, which does the actual work.

Added tests for SISO and MIMO continuous- and discrete-time.  MIMO
reference values are generated by `linfnorm` itself, but results have
been reviewed graphically for correctness.

The MIMO tests rely, unfortunately, on circular logic: the very function being tested has been used to generate test data.  I could scour the various textbooks I have for independent examples, but I'm more-or-less assuming that SLICOT's `AB13DD` is correct, and the tests are to make sure no gross errors have occurred in wrapping.  I've tried to cover the various possible cases of gpeak finite (but non-zero) and infinite, and fpeak 0, finite and non-zero, and infinite.  I just realized I haven't checked for gpeak = 0; I'll add that (system would have to be non-minimal, or be a static gain of 0; I'll try the former first).

Below are the MIMO test cases for review; everything looks OK to me.  The left-hand figures are continuous-time, the right-hand discrete time.

While doing this rather hacky test-generation, I found that `StateSpace`'s `repr` is not `eval`-able; this is due, in turn, to `numpy.ndarray`'s `repr` not being `eval`-able.  I don't know why that is, but I monkey-patched `StateSpace.__repr__` to address this.  This new `repr` *is* `eval`-able, and I think looks a little better too (look in `linfnorm_mimo_testcases.py` for examples).

gdamped
ghigh
gint
gstatic
gun
gunder

### DIFF
--- a/control/statesp.py
+++ b/control/statesp.py
@@ -55,6 +55,7 @@ from numpy.random import rand, randn
 from numpy.linalg import solve, eigvals, matrix_rank
 from numpy.linalg.linalg import LinAlgError
 import scipy as sp
+import scipy.linalg
 from scipy.signal import cont2discrete
 from scipy.signal import StateSpace as signalStateSpace
 from warnings import warn
@@ -63,7 +64,12 @@ from .namedio import _NamedIOStateSystem, _process_signal_list
 from . import config
 from copy import deepcopy
 
-__all__ = ['StateSpace', 'tf2ss', 'ssdata']
+try:
+    from slycot import ab13dd
+except ImportError:
+    ab13dd = None
+
+__all__ = ['StateSpace', 'tf2ss', 'ssdata', 'linfnorm']
 
 # Define module default parameter values
 _statesp_defaults = {
@@ -1888,3 +1894,61 @@ def ssdata(sys):
     """
     ss = _convert_to_statespace(sys)
     return ss.A, ss.B, ss.C, ss.D
+
+
+def linfnorm(sys, tol=1e-10):
+    """L-infinity norm of a linear system
+
+    Parameters
+    ----------
+    sys : LTI (StateSpace or TransferFunction)
+      system to evalute L-infinity norm of
+    tol : real scalar
+      tolerance on norm estimate
+
+    Returns
+    -------
+    gpeak : non-negative scalar
+      L-infinity norm
+    fpeak : non-negative scalar
+      Frequency, in rad/s, at which gpeak occurs
+
+    For stable systems, the L-infinity and H-infinity norms are equal;
+    for unstable systems, the H-infinity norm is infinite, while the
+    L-infinity norm is finite if the system has no poles on the
+    imaginary axis.
+
+    See also
+    --------
+    slycot.ab13dd : the Slycot routine linfnorm that does the calculation
+    """
+
+    if ab13dd is None:
+        raise ControlSlycot("Can't find slycot module 'ab13dd'")
+
+    a, b, c, d = ssdata(_convert_to_statespace(sys))
+    e = np.eye(a.shape[0])
+
+    n = a.shape[0]
+    m = b.shape[1]
+    p = c.shape[0]
+
+    if n == 0:
+        # ab13dd doesn't accept empty A, B, C, D;
+        # static gain case is easy enough to compute
+        gpeak = scipy.linalg.svdvals(d)[0]
+        # max svd is constant with freq; arbitrarily choose 0 as peak
+        fpeak = 0
+        return gpeak, fpeak
+
+    dico = 'C' if sys.isctime() else 'D'
+    jobe = 'I'
+    equil = 'S'
+    jobd = 'Z' if all(0 == d.flat) else 'D'
+
+    gpeak, fpeak = ab13dd(dico, jobe, equil, jobd, n, m, p, a, e, b, c, d, tol)
+
+    if dico=='D':
+        fpeak /= sys.dt
+
+    return gpeak, fpeak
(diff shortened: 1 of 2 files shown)


roryyorke:
I don't know how to handle those np.matrix "PendingDeprecation" warnings-as-errors.  I see `matrixfilter` and `matrixerrorfilter` marks in conftest.py, but I'm not sure when they're supposed to be used.

roryyorke:
I am still a bit lost with the PendingDeprecation error from Numpy, which I can't reproduce locally.  I developed this PR in a conda environment with Python 3.9.2, pytest 6.2.3 (which I see is a bit old), numpy 1.20.2, scipy 1.6.2, and Slycot 0.4.0.0.  python-control tests pass in this environment.

I just created a brand-new conda environment with Python 3.7.12, Pytest 7.1.2, Numpy 1.21.6, Scipy 1.7.3, and Slycot 0.4.0.0.  python-control tests all pass in this environment.

I'm running on a 64-bit Intel i5 Ubuntu 20.04 system.

bnavigator:
Your environment should be perfectly fine. I guess you can reproduce locally when you set `export PYTHON_CONTROL_ARRAY_AND_MATRIX=1`.

The critical part is not to create StateSpace or TransferFunction objects outside of the actual test functions. They are currently created at import time of the module, which is during the test collection.The filter fixtures don't apply yet. You have to move the systems into reference fixtures.

roryyorke:
> You have to move the systems into reference fixtures.

Thanks, I'll investigate that.

Not being able to reproduce the error locally makes this tricky.

roryyorke:
> export `PYTHON_CONTROL_ARRAY_AND_MATRIX=1`

Oops, missed that.  At least it's passing now.

control/statesp.py
```diff
@@ -1888,3 +1894,62 @@ def ssdata(sys):
...
+    sys : LTI (StateSpace or TransferFunction)
+      system to evalute L-infinity norm of
+    tol : real scalar
+      tolerance on norm estimate
+
+    Returns
+    -------
+    gpeak : non-negative scalar
+      L-infinity norm
+    fpeak : non-negatie scalar
```
murrayrm, line 1913: negatie → negative

control/statesp.py
```diff
@@ -1888,3 +1894,62 @@ def ssdata(sys):
...
+      Frequency, in rad/s, at which gpeak occurs
+
+    For stable systems, the L-infinity and H-infinity norms are equal;
+    for unstable systems, the H-infinity norm is infinite, while the
+    L-infinity norm is finite if the system has no poles on the
+    imaginary axis.
+
+    For systems with an undefined time base (i.e., sys.dt is None), a
+    ValueError is raised.
```
murrayrm, line 1922: Normally we allow `dt=None` to be treat as either continuous or discrete time, with default being continuous.  Should we do that here?  This is not a major issue since `dt=0` by default, but there are situations where a constant system can be converted to a state space or transfer function with `dt=None` (since it works as either discrete or continuous).

control/statesp.py
```diff
@@ -1888,3 +1894,62 @@ def ssdata(sys):
...
+        # ab13dd doesn't accept empty A, B, C, D;
+        # static gain case is easy enough to compute
+        gpeak = scipy.linalg.svdvals(d)[0]
+        # max svd is constant with freq; arbitrarily choose 0 as peak
+        fpeak = 0
+        return gpeak, fpeak
+
+    if sys.dt is None:
+        raise ValueError('unspecified time base')
+    dico = 'C' if sys.dt==0 else 'D'
```
murrayrm, line 1945: Alternatively:
```
dico = 'C' if sys.isctime() else 'D'
```

control/statesp.py
```diff
@@ -1888,3 +1894,62 @@ def ssdata(sys):
...
+    m = b.shape[1]
+    p = c.shape[0]
+
+    if n == 0:
+        # ab13dd doesn't accept empty A, B, C, D;
+        # static gain case is easy enough to compute
+        gpeak = scipy.linalg.svdvals(d)[0]
+        # max svd is constant with freq; arbitrarily choose 0 as peak
+        fpeak = 0
+        return gpeak, fpeak
```
murrayrm, line 1941: Note that this will return a value even if `sys.dt == None`, contrary to the documentation.

murrayrm (commented):
Some small comments.

The method of generating test cases is a bit nonstandard.  Do we need to check correctness of the algorithm itself (versus the processing of arguments, etc)?  Seems like we can assume slycot is working and just do a few simple unit tests??

control/statesp.py
roryyorke: I'll drop the ValueError, and use `.isctime()` as suggested below; this will treat these as continuous-time.

One case where `dt=None` by default is static gains (this is from observation, not from reading the code); and, in this case, it doesn't matter what the time base is.

roryyorke:
> The method of generating test cases is a bit nonstandard. Do we need to check correctness of the algorithm itself (versus the processing of arguments, etc)? Seems like we can assume slycot is working and just do a few simple unit tests??

Fair enough, I did get a bit carried away.

We do need to test at least for:
  - static gains (`ab13dd` doesn't handle these)
  - calling `ab13dd` with correct `DICO` value for continuous or discrete-time
  - scaling of frequency for discrete-time systems

What do you think about keeping one continuous-time SISO test (a slightly underdamped system?), an equivalent SISO discrete-time test, and one continuous-time MIMO test (probably duplicating the SISO test as a 2x2 diagonal system) ?  Also one static-gain test.

roryyorke:
All points addressed.  If the tests pass, I'll squash the PR.

bnavigator (approved):
LGTM

PR #739, opened 2022-05-31
Add passivity module, ispassive function, and passivity_test. Introduces optional dependency cvxopt. 

Mark-Yeatman:
I made a discussion topic a couple of weeks ago about making a passivity module, @murrayrm suggested making a draft PR. Here it is!


Mark-Yeatman:
I need to cut the formation of the LMI via symbolic variables for performance. Did a timing test on a size 20 state, input, output system, 99% of the time is spent constructing the LMI and converting to cvxopt form.

Mark-Yeatman:
> Would it be possible to implement the matrix handling and solving with NumPy and SciPy functions only?

I'm about 60% sure cvxopt is necessary to solve the linear matrix inequality. Sympy and lmi_sdp can definitely be dropped.

Cvxopt only requires BLAS/LAPACK, which it seems that slycot also requires. So this passivity package could be setup to be optional?

control/exception.py
```diff
@@ -84,3 +84,15 @@ def pandas_check():
...
+cvxopt_installed = None
+def cvxopt_check():
+    global cvxopt_installed
+    if cvxopt_installed is None:
+        try:
+            import cvxopt
+            cvxopt_installed = True
+        except:
+            cvxopt_installed = False
+    return cvxopt_installed
```
bnavigator, line 98:
```suggestion
    return cvxopt_installed

```

.github/workflows/python-package-conda.yml
```diff
@@ -13,6 +13,7 @@ jobs:
         python-version: [3.7, 3.9]
         slycot: ["", "conda"]
         pandas: [""]
+        cvxopt: ["conda"]
```
bnavigator, line 16: Test available and not available cvxopt

```suggestion
        cvxopt: ["", "conda"]
```

control/passivity.py
```diff
@@ -0,0 +1,61 @@
+'''
+Author: Mark Yeatman  
+Date: May 15, 2022
+'''
+
+from . import statesp as ss
```
bnavigator, line 6: Please import the official API names, not the internal module file name as a name which is also a API function.

```suggestion
from .statesp import StateSpace
```

(You actually don't use anything directly)

control/tests/passivity_test.py
```diff
@@ -0,0 +1,25 @@
+'''
+Author: Mark Yeatman  
+Date: May 30, 2022
+'''
+
+import numpy
+from control import ss, passivity
+from sympy import Matrix
```
bnavigator, line 8:
```suggestion
```

bnavigator (commented):
Thanks! Some initial review comments.

control/passivity.py
```diff
@@ -0,0 +1,61 @@
...
+import numpy as np
+import cvxopt as cvx
```
bnavigator, line 8: Make this import fail gracefully

control/passivity.py
```diff
@@ -0,0 +1,65 @@
...
+    Indicates if a linear time invarient system is passive
+
+    Constructs a linear matrix inequality and a feasibility optimization
+    such that is a solution exists, the system is passive.
+
+    The source for the algorithm is: 
+    McCourt, Michael J., and Panos J. Antsaklis. "Demonstrating passivity and dissipativity using computational methods." ISIS 8 (2013).
+    '''
+    if cvx is None:
+        raise ModuleNotFoundError
```
murrayrm, line 24: You might want to add a message to this exception so that the user knows what the need to do (eg, "cvxopt required for passivity module").

murrayrm (commented):
A few comments below.  Also, it would be nice to add a unit test to make sure that if cvxopt is not installed that the proper exception is raised.

control/passivity.py
```diff
@@ -0,0 +1,65 @@
...
+
+def is_passive(sys):
+    '''
+    Indicates if a linear time invarient system is passive
+
+    Constructs a linear matrix inequality and a feasibility optimization
+    such that is a solution exists, the system is passive.
+
+    The source for the algorithm is: 
+    McCourt, Michael J., and Panos J. Antsaklis. "Demonstrating passivity and dissipativity using computational methods." ISIS 8 (2013).
```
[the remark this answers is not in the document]
Mark-Yeatman: I have wrapped the line, but when I run "autopep8 --in-place passivity.py" it doesn't get corrected. Do you guys have a more aggressive formatter?

control/passivity.py
Mark-Yeatman: Change taken.

control/passivity.py
```diff
@@ -0,0 +1,65 @@
...
+        )
+        return V
+
+    P = np.zeros_like(A)
+    matrix_list = []
+    state_space_size = sys.nstates
+    for i in range(0, state_space_size):
+        for j in range(0, state_space_size):
+            if j <= i:
+                P = P*0.0
```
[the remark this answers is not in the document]
Mark-Yeatman: I'm really not sure. We can change it.

.github/workflows/python-package-conda.yml
```diff
@@ -13,6 +13,7 @@ jobs:
         python-version: [3.7, 3.9]
         slycot: ["", "conda"]
         pandas: [""]
+        cvxopt: ["", "conda"]
```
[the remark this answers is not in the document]
Mark-Yeatman: I think I did this correctly in my last push? Would you mind giving a code suggestion for it if I didn't?

Mark-Yeatman:
@bnavigator I believe I've resolved all the outstanding review comments?

control/passivity.py
bnavigator: The output needs to be tied together. Print going to `stdout` might end up somewhere else than the backtrace (e.g. the exception and output captures by pytest):

```suggestion
        raise ModuleNotFoundError("cvxopt required for passivity module")
```

```
[ben@skylab:~]% python3 -c 'raise  ModuleNotFoundError("cvxopt required for passivity module")'                                            [0]
  File "<string>", line 1, in <module>
ModuleNotFoundError: cvxopt required for passivity module
[ben@skylab:~]% python3 -c 'print("cvxopt required for passivity module"); raise ModuleNotFoundError'                                      [1]
cvxopt required for passivity module
  File "<string>", line 1, in <module>
ModuleNotFoundError
```

bnavigator (commented):
One more review comment

control/passivity.py
bnavigator: GitHub wouldn't let me suggest the removal of the print line in the same comment thread. Please remove the duplicate.

Mark-Yeatman: Will do

control/passivity.py
```diff
@@ -0,0 +1,66 @@
...
+
+try:
+    import cvxopt as cvx
+except ImportError as e:
+    cvx = None
+
+
+def is_passive(sys):
+    '''
+    Indicates if a linear time invarient system is passive
```
JBannwarth, line 16: `invarient` -> `invariant`

I'm just lurking around so I'm not too sure about PR etiquette. Hopefully this is okay feedback.

Mark-Yeatman: I appreciate it!

ilayn:
Instead of writing things in the standard SDP notation, you might want to use [CVXPY](https://www.cvxpy.org/examples/basic/sdp.html) notation (which acts like YALMIP) so you don't have to use the tiresome basis matrices. 

Also note that strictly proper systems cannot be checked via positive real lemma via strict inequalities due zero D + D.T block forcing the optimization problem to infeasibilty. You either have to use extended version or nudge D block with `epsilon*np.eye(m)`.

Mark-Yeatman:
I would definitely be up for using cvxpy, but it looks like it has more dependencies than just using cvxopt (https://www.cvxpy.org/install/#install-from-source vs https://cvxopt.org/install/#building-and-installing-from-source)? That seems like an issue for a project maintainer. 

You're right about the D = 0 case, I will nudge it as you suggested.

bnavigator:
CVXPY: I don't think it is necessary to employ a full domain specific language with all its dependencies for one hardcoded optimization problem.

control/passivity.py
```diff
@@ -0,0 +1,72 @@
...
+import numpy as np
+
+try:
+    import cvxopt as cvx
+except ImportError as e:
+    cvx = None
+
+lmi_epsilon = 1e-12
+
+def is_passive(sys):
```
bnavigator, line 15: We have `(sys.)isdtime()`, `(sys.)isctime()`, and `(sys.)issiso()`. If we want to maintain a sense of consistency, this should be changed to `ispassive(sys)` and wrapped by a `sys.ispassive()` method for `StateSpace` systems, possibly even a wrapper for other LTI systems which realize them into `StateSpace` before calling the function in the passivity module.

`ispassive` would also be consistent with the naming scheme from [NumPy](https://numpy.org/doc/stable/reference/routines.logic.html) and [stdlib Python](https://docs.python.org/3/library/math.html#math.isclose)

control/tests/passivity_test.py
```diff
@@ -15,10 +15,22 @@ def test_is_passive():
...
+    # happy path is passive
     assert(passivity.is_passive(sys))
 
+    # happy path not passive
     D = -D
     sys = ss(A, B, C, D)
 
     assert(not passivity.is_passive(sys))
 
+    #edge cases of D=0 boundary condition
```
bnavigator, line 27: Please put into separate tests.

control/passivity.py
```diff
@@ -10,6 +10,7 @@
 except ImportError as e:
     cvx = None
 
+lmi_epsilon = 1e-12
```
bnavigator, line 13: unused

control/passivity.py
```diff
@@ -0,0 +1,72 @@
...
+try:
+    import cvxopt as cvx
+except ImportError as e:
+    cvx = None
+
+lmi_epsilon = 1e-12
+
+def is_passive(sys):
+    '''
+    Indicates if a linear time invariant system is passive
```
bnavigator, line 17: Please use [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) to indicate the input parameter and return type.

control/passivity.py
Mark-Yeatman: Renaming the function for consistency sounds good.

I am planning on trying to add some sort of passivity related tools for systems beyond ones that are LTI (an LMI to check passivity was the lowest hanging fruit). In [Demonstrating Passivity and Dissipativity using Computational Methods](https://www3.nd.edu/~pantsakl/Publications/497-isis08%202013.pdf), there's methods for using sum-of-squares to compute passivity for polynomial nonlinear systems and I believe there's literature elsewhere for systems described by combinations of polynomials and trig functions. 

I think it is a good suggestion, but I'm not immediately sure what parent class to try to add a wrapper for ispassive() to.

control/tests/passivity_test.py
```diff
@@ -0,0 +1,63 @@
...
+    # strictly proper
+    sys = ss(A, B, C, D)
+    assert(passivity.ispassive(sys))
+
+    # ill conditioned
+    A = A*1e12
+    sys = ss(A, B, C, D)
+    assert(passivity.ispassive(sys))
+
+    # different combinations of zero A,B,C,D are 0
```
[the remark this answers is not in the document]
Mark-Yeatman: Got it! Thanks for working with me through this, sorry for some of the lazy commits.

bnavigator:
I took the liberty of fixing the missing skips. I hope you don't mind.

control/lti.py
```diff
@@ -201,6 +202,11 @@ def _dcgain(self, warn_infinite):
         else:
             return zeroresp
 
+    def ispassive(self):
+        # importing here prevents circular dependancy
+        from control.passivity import ispassive
+        ispassive(self)
```
bnavigator, line 208:
```suggestion
        return(ispassive(self))
```

Mark-Yeatman: Good catch!

PR #750, opened 2022-07-18
Passivity indices and support for discrete time systems. 

Mark-Yeatman:
This adds support for input and output passivity indices. The intention is to mimic the functionality of [ispassive](https://www.mathworks.com/help/control/ref/lti.ispassive.html) and [getPassiveIndex](https://www.mathworks.com/help/control/ref/lti.getpassiveindex.html) from Matlab (this implementation ignores frequency bands, relative passivity index, combined I/O passivity, and directional passivity index.)

This also adds support for computing indices and "is passive or not" for discrete time systems.


Mark-Yeatman:
Seems like the pipeline is having some issues unrelated to my changes?

bnavigator:
Yes it's unrelated. I'll have a look shortly.

control/passivity.py
```diff
@@ -1,41 +1,68 @@
...
-    -------
-    bool: 
-        The input system passive.
-    '''
+def __ispassive__(sys, rho=None, nu=None):
     if cvx is None:
         raise ModuleNotFoundError("cvxopt required for passivity module")
 
+    if sys.ninputs != sys.noutputs:
+        raise Exception(
```
bnavigator, line 60: Please use a more non-generic `Exception` class like `ControlDimension`

control/passivity.py
```diff
@@ -1,41 +1,68 @@
...
+            "The number of system inputs must be the same as the number of system outputs.")
+
+    if rho is None and nu is None:
+        raise Exception("rho or nu must be given a float value.")
```
bnavigator, line 64: `control.exceptions.ControlArgument`

control/tests/passivity_test.py
```diff
@@ -2,16 +2,17 @@
 Author: Mark Yeatman  
 Date: May 30, 2022
 '''
+from tokenize import Double
```
bnavigator, line 5: ?

control/tests/rlocus_test.py
```diff
@@ -112,7 +112,7 @@ def test_root_locus_zoom(self):
         assert_array_almost_equal(zoom_x, zoom_x_valid)
         assert_array_almost_equal(zoom_y, zoom_y_valid)
 
-    @pytest.mark.timeout(2)
+    # @pytest.mark.timeout(2)
```
bnavigator, line 115: ?

bnavigator:
> Seems like the pipeline is having some issues unrelated to my changes?

Fixed. I merged current main into your PR. You have to pull your remote branch before continuing to work or rebase your commits onto the updated main branch yourself.

control/passivity.py
```diff
@@ -1,41 +1,68 @@
 '''
 Author: Mark Yeatman  
-Date: May 15, 2022
+Date: July 17, 2022
 '''
 
+from msilib.schema import Error
```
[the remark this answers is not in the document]
Mark-Yeatman: My vscode extensions really like adding random imports when I mistype things.

control/tests/rlocus_test.py
bnavigator: Install pytest-timeout

control/tests/passivity_test.py
Mark-Yeatman: I also blame vscode here 🤷‍♂️

control/passivity.py
Mark-Yeatman: Change taken.

control/passivity.py
Mark-Yeatman: Change taken.

control/tests/rlocus_test.py
Mark-Yeatman: That's the ticket!

control/tests/passivity_test.py
```diff
@@ -54,6 +114,17 @@ def test_ispassive_edge_cases(test_input, expected):
...
 
 
+def test_ispassive_all_zeros():
+    A = numpy.array([[0]])
+    B = numpy.array([[0]])
+    C = numpy.array([[0]])
+    D = numpy.array([[0]])
+    sys = ss(A, B, C, D)
+
+    with pytest.raises(Exception):
```
bnavigator, line 124: Check for the more specific exception

control/passivity.py
```diff
@@ -45,43 +71,142 @@ def ispassive(sys):
...
+    sys: An LTI system
+        System to be checked.
+    nu: float
+        Concrete value for input passivity index. 
+    rho: float
+        Concrete value for output passivity index. 
+
+    Returns
+    -------
+    bool or float:
```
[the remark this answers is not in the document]
Mark-Yeatman: Ah, good catch, the "None" case is for when you try to get a passivity index but the problem is unfeasible.

control/tests/passivity_test.py
Mark-Yeatman: Change taken.

control/passivity.py
```diff
@@ -45,43 +71,143 @@ def ispassive(sys):
...
+    else:
+        sol_rho = __ispassive__(sys, nu=eps)
+        sol_nu = __ispassive__(sys, rho=eps)
+        if sol_rho is not None and sol_nu is not None:
+            return sol_rho[-1] > 0 or sol_nu[-1] > 0
+        else:
+            return False
+
+
+def getPassiveIndex(sys, index_type=None):
```
murrayrm, line 191: The convention in python-control is to use intercaps for class names, but not functions.  This should be `get_passive_index` (or `get_passivity_index`?).

Mark-Yeatman:
@bnavigator I'm still confused about why the checks are failing?

Mark-Yeatman:
Turns out I misread the documentation for the feedback function and got the function arguments in the wrong order. Everything should work now.

murrayrm:
There's some uncovered code that it would be useful to catch in the unit tests.  You can see the lines that are missing [here](https://coveralls.io/builds/51075148/source?filename=control%2Fpassivity.py).

bnavigator:
I have to admit that it is a bit confusing to me what nu, rho and the various instances of None mean. Especially that the meaning of the ispassive return value depends on them. So unless this is some standard convention adopted from the paper algorithms, I would rather like to see these cases implemented in separate API functions and methods. Same for `get_passivity_index()`.

Mark-Yeatman:
> I have to admit that it is a bit confusing to me what nu, rho and the various instances of None mean. Especially that the meaning of the ispassive return value depends on them. So unless this is some standard convention adopted from the paper algorithms, I would rather like to see these cases implemented in separate API functions and methods. Same for `get_passivity_index()`.

I'm trying to copy matlab's API here https://www.mathworks.com/help/control/ref/lti.getpassiveindex.html and https://www.mathworks.com/help/control/ref/lti.ispassive.html

bnavigator:
I see. The MATLAB function `isPassive` always returns the matlab equivalent of `bool` for the first return value `pf` and `R` as the passivity index if applicable. We usually mimic something like this with a `return_index` kwarg in the function.

Something like:

```
return (pf, R) if return_index else pf
```

Exact mimicry of the MATLAB function, if desired at all, should only be necessary in `control.matlab.isPassive()`

Note that the MATLAB functions claim to return a meaningful return value with both `rho` and `nu` not given which is equivalen to both `None` in your implementation.

Mark-Yeatman:
> if desired at all, should only be necessary in `control.matlab.isPassive'

Gotcha, I will change it around so that there's different function calls for each index type.

Mark-Yeatman:
@bnavigator I've split get_passivity_index() into get_output_fb_index() and get_input_ff_index().  I've also removed all the weird conditionals in ispassive(), it just has one code path now. I kept all the nu's and rho's in _solve_passivity_LMI() because it's easier to match the syntax from the source paper that way IMO.

murrayrm:
It looks like the passivity module is not (yet) loaded in __init__.m.  You should probably do that, so people don't have to import that passivity module explicitly (which allows `ct.ispassive(sys)`).  Also, we should update the Sphinx documentation to include `ispassive` in the list of functions that are implemented (in `doc/control.rst`).

Other than that, looks ready to go?

control/passivity.py
```diff
@@ -45,43 +84,127 @@ def ispassive(sys):
...
+    Indicates if a linear time invariant (LTI) system is passive
+
+    Parameters
+    ----------
+    sys: An LTI system
+        System to be checked.
+
+    Returns
+    -------
+    bool, float, or None:
```
bnavigator, line 204: Is only bool now

control/passivity.py
```diff
@@ -45,43 +84,127 @@ def ispassive(sys):
...
+        return -np.inf
+    else:
+        return sol[-1]
+
+
+def ispassive(sys, rho=None, nu=None):
+    '''
+    Indicates if a linear time invariant (LTI) system is passive
+
+    Parameters
```
bnavigator, line 197: Lacks the description for `rho` and `nu`

control/passivity.py
```diff
@@ -45,43 +84,127 @@ def ispassive(sys):
...
+        The OFP index 
+    '''
+    sol = _solve_passivity_LMI(sys, nu=eps)
+    if sol is None:
+        return -np.inf
+    else:
+        return sol[-1]
+
+
+def get_input_ff_index(sys, index_type=None):
```
bnavigator, line 171: `index_type` unused

control/passivity.py
Mark-Yeatman: meant to remove them

control/passivity.py
Mark-Yeatman: almost meant to remove it

control/passivity.py
Mark-Yeatman: yep, good catch

control/passivity.py
```diff
@@ -45,43 +84,126 @@ def ispassive(sys):
...
+        return -np.inf
+    else:
+        return sol[-1]
+
+
+def get_input_ff_index(sys):
+    '''
+    Returns the input feedforward passivity (IFP) index for the input system. This is the largest gain that can be 
+    placed in negative parallel interconnection with a system such that the new interconnected system is passive.
+    Parameters
```
murrayrm, line 175: As above.

control/passivity.py
```diff
@@ -45,43 +84,126 @@ def ispassive(sys):
...
+    sol = cvx.solvers.sdp(c, Gs=Gs, hs=hs)
+    return sol["x"]
+
 
-    return (sol["x"] is not None)
+def get_output_fb_index(sys):
+    '''
+    Returns the output feedback passivity index for the input system. This is largest gain that can be placed in
+    positive feedback with a system such that the new interconnected system is passive.
+    Parameters
```
murrayrm, line 154: This first line of the docstring needs to be a short phrase that describes the function, so that the table entriex in the sphinx documentation show up correctly.  See https://numpydoc.readthedocs.io/en/latest/format.html for information (in particular, "short summary" versus "extended summary").  See [`iosys.py`](https://github.com/python-control/python-control/blob/main/control/iosys.py) for lots of examples.

Also, I think you need a blank line before Parameters for Sphinx to properly parse things.

murrayrm (commented):
More small things to fix (sorry I didn't notice these on my previous pass..).

control/passivity.py
```diff
@@ -4,7 +4,7 @@
 '''
 
 import numpy as np
-from control import statesp as ss
+from control import statesp
```
bnavigator, line 7: The name collision was a friendly reminder, that you should only export your own functions when `from .passivity import *` is used: Define `__all__`.

control/passivity.py
```diff
@@ -45,43 +84,126 @@ def ispassive(sys):
...
+    '''
+
+    sol = _solve_passivity_LMI(sys, rho=eps)
+    if sol is None:
+        return -np.inf
+    else:
+        return sol[-1]
+
+
+def ispassive(sys):
```
bnavigator, line 193: But what about the cases, when you want to determine (boolean)
- input passivity (`nu != 0, rho = 0`), 
- output passivity (`nu = 0, rho != 0`)

now? One would have to use the `get_*_index()` functions. There is also no function to determine the relative passivity index.

Sorry if I created confusion earlier. I meant to make the function return values intuitive, not to remove functionality.

control/passivity.py
```diff
@@ -45,43 +84,126 @@ def ispassive(sys):
...
+
+    Parameters
+    ----------
+    sys: An LTI system
+        System to be checked.
+
+    Returns
+    -------
+    bool: 
+        The input system is passive.
```
bnavigator, line 205: "input"?

control/tests/passivity_test.py
```diff
@@ -54,6 +114,26 @@ def test_ispassive_edge_cases(test_input, expected):
...
+    A = numpy.array([[0]])
+    B = numpy.array([[0]])
+    C = numpy.array([[0]])
+    D = numpy.array([[0]])
+    sys = ss(A, B, C, D)
+
+    with pytest.raises(ValueError):
+        passivity.ispassive(sys)
+
+def test_rho_and_nu_are_none():
```
bnavigator, line 127: Is this really a forbidden case? I see it is in the current implementation, but is it from the definition? Also, what's the difference between 0 and `None` exactly?

control/passivity.py
Mark-Yeatman: You could determine input passivity with  `is_input_passive = get_input_ff_index(sys) > 0` (same for output passivity). I could add two more functions that do just that. 

Right now, I'm intentionally not fiddling with computing the relative passivity index, I need to go read the source that the Matlab documentation is citing "Sector Bounds in Stability Analysis and Control Design". If you think it should be included in this pull request, I'm totally willing to add it.

bnavigator: > I could add two more functions that do just that.

Despite earlier discussions, I would vote for including that functionality into `ispassive(sys, nu, rho)`

bnavigator: > Right now, I'm intentionally not fiddling with computing the relative passivity index, I need to go read the source that the Matlab documentation is citing "Sector Bounds in Stability Analysis and Control Design". If you think it should be included in this pull request, I'm totally willing to add it.

No that is fine. This can be defered to later PRs. Just raise a `NotImplemented` Exception instead of stating "This input is invalid". Same for I/O passivity index, I guess.

control/tests/passivity_test.py
Mark-Yeatman: I'm trying to use `None` as a marker to treat the index as a variable to be maximized. Trying to maximize both the input and output passivity indices is **bilinear** matrix inequality, versus a **linear** matrix inequality. The former is **not** convex.

control/passivity.py
Mark-Yeatman: > But what about the cases, when you want to determine (boolean)
> 
> * input passivity (`nu != 0, rho = 0`),
> * output passivity (`nu = 0, rho != 0`)
> 
> now? One would have to use the `get_*_index()` functions. 
> 
> Sorry if I created confusion earlier. I meant to make the function return values intuitive, not to remove functionality.

I believe I've captured this use case now.

Mark-Yeatman: > No that is fine. This can be defered to later PRs. Just raise a NotImplemented Exception instead of stating "This input is invalid". **Same for I/O passivity index, I guess.**

I'm confused about the bolded part?

bnavigator: Sorry, forget the bolded part. Sloppy editing by me. When there is no `get_passive_index()` catchall function you don't need  that.

But *(optional)* nice to have as a future reminder to implement it:

```python
def get_relative_index(sys):
    """Returns the relative passivity index for the system.

    (not implemented yet)
    """
    raise NotImplemented("Relative passivity index not implemented")


def get_combined_io_index(sys):
    """Returns the combined I/O passivity index for the system.

    (not implemented yet)
    """
    raise NotImplemented("Combined I/O passivity index not implemented")


def get_directional_index(sys):
    """Returns the directional passivity index for the system.

    (not implemented yet)
    """
    raise NotImplemented("Directional passivity index not implemented")

```

control/tests/passivity_test.py
bnavigator: Alright, you make a distinction between the parameters for `ispassive()` and `solve_passivity_LMI()`. Got it.

control/passivity.py
```diff
@@ -45,43 +69,180 @@ def ispassive(sys):
...
+    ofp_index: float
+        Output feedback passivity index.
+    ifp_index: float
+        Input feedforward passivity index.
+
+    Returns
+    -------
+    bool: 
+        The system is passive.
+    '''
```
bnavigator, line 247: "equiavlent"

The math doesn't render well:

ispassive_spyder

This one does not throw any linter errors at me:

```suggestion
    r"""Indicate if a linear time invariant (LTI) system is passive.

    Checks if system is passive with the given output feedback (OFP) and input
    feedforward (IFP) passivity indices.

    Parameters
    ----------
    sys: An LTI system
        System to be checked.
    ofp_index: float
        Output feedback passivity index.
    ifp_index: float
        Input feedforward passivity index.

    Returns
    -------
    bool: 
        The system is passive.

    Notes
    -----
    Querying if the system is passive in the sense of
    
    .. math:: V(x) >= 0 \land \dot{V}(x) <= y^T u

    is equivalent to the default case of `ofp_index = 0`, `ifp_index = 0`.

    Note that computing the ofp_index and ifp_index for a system, then using
    both values simultaneously to as inputs to this function is not guaranteed
    to have an output of `True` (the system might not be passive with both
    indices at the same time).
    
    For more details, see [1].

    References
    ----------
    .. [1] McCourt, Michael J., and Panos J. Antsaklis
          "Demonstrating passivity and dissipativity using computational methods." 
    """   
```


ispassive_spyder_new


(Yours shows me D300 D301 D401 in spyder)
https://numpydoc.readthedocs.io/en/latest/format.html
http://www.pydocstyle.org/en/6.1.1/error_codes.html

Mark-Yeatman: That is much better, thanks for doing that!

control/passivity.py
Mark-Yeatman: Awesome, I'll take that change.

Mark-Yeatman:
> You actually need the `r"""` here or have to use double backslashes below.

Is there a name for what this is?

bnavigator:
> Is there a name for what this is?

["longstring stringliteral with raw stringprefix" or "raw triple-quoted strings"](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals)

Mark-Yeatman:
> > Is there a name for what this is?
> 
> ["longstring stringliteral with raw stringprefix" or "raw triple-quoted strings"](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals)
> 
> [PEP 257](https://peps.python.org/pep-0257/):
> 
> > For consistency, always use `"""triple double quotes"""` around docstrings. Use `r"""raw triple double quotes"""` if you use any backslashes in your docstrings. For Unicode docstrings, use `u"""Unicode triple-quoted strings"""`.

Noice

murrayrm:
@Mark-Yeatman @bnavigator This looks like it is ready to merge?  Anything else planned?

bnavigator:
Thanks @Mark-Yeatman! Looking forward to future pull request with further enhancements of this module.

PR #755, opened 2022-08-04
Fix timebase bug in InterconnectedSystem (issue #754)

murrayrm:
This PR fixes the timebase bug identified in issue #754.  The fix was to modify the processing of the timebase in `InterconnectedSystem` so that it looks at the systems for the default timebase.


murrayrm:
Conda-based builds are failing.  I checked and this is also happening on the current main branch.  This may be something associated with Slycot 0.5.0??

control/tests/timebase_test.py
```diff
@@ -0,0 +1,37 @@
...
+        (0, 0.1, ValueError),
+        (0, None, 0),
+        (0.1, 0, ValueError),
+        (0.1, 0.1, 0.1),
+        (0.1, None, 0.1),
+        (None, 0, 0),
+        (None, 0.1, 0.1),
+        (None, None, None),
+        (0.2, None, 0.2),
+        (0.2, 0.1, ValueError),
```
sawyerbfuller, line 18: Great PR. Would be good to add tests for `dt=True`, which are compatible with all other discrete-time timebases. https://python-control.readthedocs.io/en/0.9.2/conventions.html

murrayrm: Great suggestion. Done.

Issue #764, opened 2022-08-24
control.matlab.lsim does not work for systems of order > 1


jonschlipf:
When calling the control.matlab.lsim with a higher-order system (at least 2x2 matrices), there is a prompt:
```
ValueError: Parameter ``T`` must have same elements as the number of columns in input array ``U`` 
```
Inverting U or T does not help. Instead, it prompts a different error with matrix multiplication.
It works properly when calling control.timeresp.forced_response directly, with the "transpose" flag set to false. 
The error is likely due to the transpose flag, and its implementation throughout the declaration of control.timeresp.forced_response

```
import numpy as np
import control
import control.matlab

time = np.linspace(0.0, 511.0e-6, 512)
DAC = np.sin(time)
ADC = np.cos(time)

input_Kalman = np.transpose(np.concatenate(([[DAC]], [[ADC]]), axis=1)[0])
Af = [[0.45768416, -0.42025511], [-0.43354791, 0.51961178]]
Bf = [[2.84368641, 52.05922305], [-1.47286557, -19.94861943]]
Cf = [[1.0, 0.0], [0.0, 1.0]]
Df = [[0.0, 0.0], [0.0, 0.0]]
ss_Kalman = control.ss(Af, Bf, Cf, Df, 1.0e-6)
x_est, _, _ = control.matlab.lsim(ss_Kalman, input_Kalman, time)
```

bnavigator:
Confirmed

```
In [5]: input_Kalman.shape
Out[5]: (512, 2)

In [6]: x_est, _, _ = control.matlab.lsim(ss_Kalman, input_Kalman, time)
Traceback (most recent call last):

  Input In [6] in <cell line: 1>
    x_est, _, _ = control.matlab.lsim(ss_Kalman, input_Kalman, time)

  File ~/src/python-control/control/matlab/timeresp.py:297 in lsim
    out = forced_response(sys, T, U, X0, return_x=True, transpose=True)

  File ~/src/python-control/control/timeresp.py:978 in forced_response
    raise ValueError('Parameter ``T`` must have same elements as'

ValueError: Parameter ``T`` must have same elements as the number of columns in input array ``U``


In [7]: x_est, _, _ = control.matlab.lsim(ss_Kalman, input_Kalman.T, time)
Traceback (most recent call last):

  Input In [7] in <cell line: 1>
    x_est, _, _ = control.matlab.lsim(ss_Kalman, input_Kalman.T, time)

[...]

PR #765, opened 2022-08-25
fix control.matlab.lsim bug for discrete time system

murrayrm:
This PR fixes issue #764.  There was a dimension check in `forced_response` for discrete time systems that occurred before processing of the `transpose` parameter.  Removing the dimension check (which gets done later, after `transpose` processing) fixes the bug.  Add a unit test to catch the error and also turned on some tests that were `@slycotonly` but that didn't rely on slycot.


bnavigator:
LGTM, thanks!

PR #767, opened 2022-08-31
Dc motor branch

DaveNganga:
Created a new tutorial file that uses a Collimator DC model to model optimal values for a DC motor Shaft.

### DIFF
--- a/examples/dc-motor.py
+++ b/examples/dc-motor.py
@@ -1,0 +1,63 @@
+######################################################################################
+# Uses Collimtaor to calculate                                                       #
+# the optimal values for a DC motor shaft position                                   #
+# See here: https://www.collimator.ai/tutorials/dc-motor-position-controller-design  #
+# for full tutorial                                                                  #
+######################################################################################
+
+import pycollimator as collimator
+import matplotlib.pyplot as plt
+
+token = "cbc28a3a-50c4-4903-8396-c186b45b460d"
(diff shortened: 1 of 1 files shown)


murrayrm:
Can you provide some more context for this tutorial and why it should be included here?

It looks like this uses collimator.ai, which is a commercial tool (I think).  Since python-control is open source, I'm not sure it would be compatible to put an example that relies on a non-open source library.

bnavigator:
- The same file is added twice with two different names
- All of the python-control related code seems to be encapsulated in `pycollimator`. This differs from the referenced  https://www.collimator.ai/tutorials/dc-motor-position-controller-design .
- Never hardcode tokens and project-uuids into a publicily available example

DaveNganga:
Yeah so Collimator is an online modeling and simulation platform much like Matlab/Simulink, and while Collimator is not open source it does have a free version for individuals with small projects. The file was an implementation from the motor controller design linked in the python file: https://www.collimator.ai/tutorials/dc-motor-position-controller-design 
as per the duplicate names, it was an error on my end: I didn't see it added the first time I put the file in hence the reason for the duplicate names. 
As per the hard-coded tokens: They're public uuid's and tokens. 
PyCollimator is still in its developmental stages and is the package that would need to be called to implement Collimator through API. 
I'm sorry for the confusion

PR #773, opened 2022-10-01
Add figsize and dpi parameters to control.sisotool

IanViotti:
The figure for sisotool plot is currently too small and hard to read.

I added  figsize and dpi as optional parameters to the function and passed it in the plt.subplots() function.


bnavigator:
Unless we give specific values in order to improve the default behavior of `sisotools`, I think that this is better controlled by the user in [Matplotlib's rcParams defaults](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.figure.html) without any interference by python-control.

IanViotti:
> @IanViotti thanks for your PR. Do you think what @bnavigator suggests is a sufficient solution?

Absolutely. Sorry I didn't close the pull request.

PR #774, opened 2022-10-03
Move sys._update_params(params) before TimeResponseData return when nstate == 0

hyumo:
It was found that for system without states, `params` not are updated at the right timing in the `input_output_response` function. 

Need to move `sys._update_params(params)` before `TimeResponseData` returns if `nstate == 0`


bnavigator:
Thank you

PR #780, opened 2022-10-14
Improve compatibility of state space representation using LaTeX

gonmolina:
It is not work with KaTex renderer

Maths in notebooks are rendered using KaTex in vscode. This result in:
Screenshot_20221014_124728

With these changes:
Screenshot_20221014_135354

It is algo working in Jupyter Lab.


bnavigator:
Hi @gonmolina, thank you for your contribution!

The change looks reasonable to me. Could you please adjust the tests accordingly?

For example here:
https://github.com/python-control/python-control/blob/9c0ab033805e52d8c30a4aa7eb52859f11c642db/control/tests/statesp_test.py#L1014-L1031

I see the TransferFunction LaTeX representation already has `$$`.

bnavigator:
Excellent! Thank you!

PR #788, opened 2022-11-13
Fix interconnect type conversion bug for StateSpace systems

murrayrm:
This PR fixes a problem that was identified in PR #785, where interconnecting a `LinearIOSystem` with a `StateSpace` system via the `interconnect` function did not work correctly.  In particular, if you created a mixed system of this type you would get back an `InterconnectedSystem` that would generate an error is you tried to simulate it or evaluate the dynamics.  This was fixed by adding a few lines of code to `interconnect()` that convert `StateSpace` and `TransferFunction` objects to `LinearIOSystems`, mimicking what is done with operator overloading.

In addition, there was a bug where the `param` keyword was not allowed in the `dynamics` and `output` functions.  This is now fixed and tested with a unit test.

### DIFF
--- a/control/iosys.py
+++ b/control/iosys.py
@@ -376,16 +376,17 @@ class InputOutputSystem(NamedIOSystem):
         right hand side of the dynamical system. If the system is continuous,
         returns the time derivative
 
-            dx/dt = f(t, x, u)
+            dx/dt = f(t, x, u, params)
 
         where `f` is the system's (possibly nonlinear) dynamics function.
         If the system is discrete-time, returns the next value of `x`:
 
-            x[t+dt] = f(t, x[t], u[t])
+            x[t+dt] = f(t, x[t], u[t], params)
 
-        Where `t` is a scalar.
+        where `t` is a scalar.
 
-        The inputs `x` and `u` must be of the correct length.
+        The inputs `x` and `u` must be of the correct length.  The `params`
+        argument is an optional dictionary of parameter values.
 
         Parameters
         ----------
@@ -395,6 +396,8 @@ class InputOutputSystem(NamedIOSystem):
             current state
         u : array_like
             input
+        params : dict (optional)
+            system parameter values
 
         Returns
         -------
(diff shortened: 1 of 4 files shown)


control/iosys.py
```diff
@@ -369,23 +369,24 @@ def _rhs(self, t, x, u, params={}):
         NotImplemented("Evaluation not implemented for system of type ",
                        type(self))
 
-    def dynamics(self, t, x, u):
+    def dynamics(self, t, x, u, params={}):
```
roryyorke, line 372: Is there a reason not to use the `params=None`, `if params is None: params={}` pattern here?  Using mutable objects for default parameters is usually not a good idea.

Also, why not **kwargs?

roryyorke (commented):
LGTM, except for the one comment.

control/iosys.py
```diff
@@ -369,23 +369,24 @@ def _rhs(self, t, x, u, params={}):
...
         """Compute the dynamics of a differential or difference equation.
 
         Given time `t`, input `u` and state `x`, returns the value of the
         right hand side of the dynamical system. If the system is continuous,
         returns the time derivative
 
-            dx/dt = f(t, x, u)
+            dx/dt = f(t, x, u, params)
```
sawyerbfuller, line 379:
```suggestion
            dx/dt = f(t, x, u[, params])
```

control/iosys.py
```diff
@@ -369,23 +369,24 @@ def _rhs(self, t, x, u, params={}):
...
 
         where `f` is the system's (possibly nonlinear) dynamics function.
         If the system is discrete-time, returns the next value of `x`:
 
-            x[t+dt] = f(t, x[t], u[t])
+            x[t+dt] = f(t, x[t], u[t], params)
```
sawyerbfuller, line 384:
```suggestion
            x[t+dt] = f(t, x[t], u[t][, params])
```

sawyerbfuller (commented):
One other thought: currently, `StateSpace.dynamics` does not have `params` as a kwarg, but it probably should (maybe with a warning that it will be ignored).

roryyorke:
Like the `for x in sequence or []:` pattern.  LGTM.

PR #790, opened 2022-11-13
fix _isstatic() to use nstates==0

murrayrm:
This PR fixes the issues identified in PR #785 regarding the way that `StateSpace._isstatic` was defining a static system.  New version requires `nstates == 0`.

### DIFF
--- a/control/tests/iosys_test.py
+++ b/control/tests/iosys_test.py
@@ -41,6 +41,9 @@ class TestIOSys:
             [[-1, 1], [0, -2]], [[0, 1], [1, 0]],
             [[1, 0], [0, 1]], np.zeros((2, 2)))
 
+        # Create a static gain linear system
+        T.staticgain = ct.StateSpace([], [], [], 1)
+
         # Create simulation parameters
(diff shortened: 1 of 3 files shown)


control/tests/iosys_test.py
```diff
@@ -51,7 +54,7 @@ class TSys:
     def test_linear_iosys(self, tsys):
         # Create an input/output system from the linear system
         linsys = tsys.siso_linsys
-        iosys = ios.LinearIOSystem(linsys)
+        iosys = ios.LinearIOSystem(linsys).copy()
```
roryyorke, line 57: is it obvious why this `.copy` is necessary?

roryyorke (commented):
LGTM

control/tests/iosys_test.py
murrayrm: Definitely not obvious, but this gets rid of a warning later on since the `interconnect` function generates a warning if a system appears more than once in an interconnected system, since then the name of system is not unique.  (It's also possible to turn the warning off with `warn_dupicate=False`).

PR #793, opened 2022-11-17
fix error when an IOSystem is combined with a TransferFunction system

sawyerbfuller:
Fixes a bug in which system and system name were not being handled correctly when a `TransferFunction` system is combined with other linear systems using `interconnect`.


bnavigator (approved):
LGTM

PR #794, opened 2022-11-19
check for and fix mutable keyword defaults

murrayrm:
This PR adds a check for keyword arguments in all functions that are initialized to mutables and fixes a few errors that were found.  Setting keyword defaults to mutables is still allowed by listing the function in a table.  This should help introduction of errors if a mutable is used as a default keyword in future updates.


roryyorke:
Looks good to me.  FWIW [pylint](https://pylint.pycqa.org/en/latest/) has a warning for mutable args.

Issue #786, opened 2022-11-07
Modest proposal for discussion: eliminate `LinearICSystem` class


sawyerbfuller:
It is not clear to me what purpose there is for a `LinearICSystem`. One possibility is as a biquad cascade*, but this can already be implemented by forcing the interconnected systems to be `NonlinearIOSystems`, and is not currently supported by `LinearICSystem`s anyway because the constituent systems are combined into a single system with A, B, C, D matrices and signal names. 

Is there any other use case for a `LinearICSystem`? I can't figure out what benefit there is to preserving other characteristics of an `InterconnectedSystem`, if such a system is always used in an equivalent way to a `LinearIOSystem`.  

*There are cases when you don't want a combined interconnected series of systems to be tunred into a single state-space or transfer function system. For example, if you want to numerically interconnect a series of linear systems to create a biquad cascade to avoid numerical issues that crop up when combining large and small coefficients. This comes up for example when implementing a high-order pade approximation (see e.g. https://www.embeddedrelated.com/showarticle/927.php)

murrayrm:
The current system class hierarchy has been stable for a few years now and so I think this issue can be closed.

For posterity: the `LinearICSystem` class is used when you create a linear system via `interconnect` and it maintains both the interconnected system structure (subsystems, interconnection matrix, etc) as well as the state space structure (A, B, C, D matrices).  If you simulate a `LinearICSystem` using `forced_response` (or the other linear response functions) then it will use the linear system structure.  If you simulate it using `input_output_response`, it will use the subsystem structure (computing the right hand side from all of the individual subsystems and interconnections).  The latter is probably a bit slower.

One justification for keeping the `LinearICSystem` object is that if we implement functionality that utilizes the subsystem structure (eg, being able to extract signals via a hierarchical signal name like 'subsys.signal') then this would allow that to be done for linear interconnected systems.

Note also that using `series`, `parallel`, and `feedback` (or their equivalent operators) will not generate a `LinearICSystem`.

Issue #728, opened 2022-04-24
`InputOutputSystem.copy` makes shallow copies


roryyorke:
This is may be by design, but I found it unexpected (see below).  If it *is* by design, it would help to add "shallow" to the doctring.

```python
In [220]: g = ct.rss()

In [221]: g.D[0,0]
Out[221]: -1.8274167505512213

In [222]: g2 = g.copy()

In [223]: g2.D[0,0] = 1

In [224]: g.D[0,0]
Out[224]: 1.0
```

One way to deep copy is to use `control.ss`; continuing from above:

```
In [225]: g3 = ct.ss(g)

In [226]: g3.D[0,0] = 5

In [227]: g.D[0,0]
Out[227]: 1.0
```

murrayrm:
Good catch.

My sense is that if someone calls the `copy` method explicitly, we should do a deep copy.  But we should also think about compatibility with NumPy, SciPy, etc and I'm not sure what they do.

Some other situations we should also think about:

* If you call the `ss` or `tf` factory functions and pass them a state space system or transfer function, respectively, they just return that object.  Should they do a copy instead?  Shallow or deep?
* When you use a single system more than once in an interconnected system, we generate a copy (so that things like the state of one block are distinct from the state of the other).  Shallow or deep?

roryyorke:
> But we should also think about compatibility with NumPy, SciPy, etc and I'm not sure what they do.

The answer is "it's complicated"; for arrays of immutable types (e.g., numbers), `ndarray.copy` appears to be deep:

```python
n [5]: a = np.array([1,2,3])

In [6]: b = a.copy()

In [7]: b[0] = 100

In [8]: a[0]
Out[8]: 1
```

but for mutable array elements, with dtype=object, you can see the copies are actually shallow:

```python
n [11]: c = np.array([{'a':23},{'b':99}],dtype=object)

In [12]: c[0]
Out[12]: {'a': 23}

In [13]: d = c.copy()

In [14]: d[0]['b'] = 100

In [15]: c[0]
Out[15]: {'a': 23, 'b': 100}
```

> If you call the ss or tf factory functions and pass them a state space system or transfer function, respectively, they just return that object. Should they do a copy instead?

[...]

PR #797, opened 2022-11-23
Preserve signal names upon conversion to discrete-time

sawyerbfuller:
This PR updates `sys.sample` in both `StateSpace` and `TransferFunction` to return a system with the same input and output labels, which is convenient when constructing interconnected systems using `interconnect`. 

The changes in this PR only copy signal names, but the name of the system is not preserved. I haven't used the system name feature before but my impression was that it should be unique, but please let me know if you think it should be passed as well.

### DIFF
--- a/control/config.py
+++ b/control/config.py
@@ -288,7 +288,7 @@ def use_legacy_defaults(version):
         set_defaults('control', default_dt=None)
 
         # changed iosys naming conventions
-        set_defaults('iosys', state_name_delim='.',
+        set_defaults('namedio', state_name_delim='.',
                      duplicate_system_name_prefix='copy of ',
                      duplicate_system_name_suffix='',
                      linearized_system_name_prefix='',
--- a/control/config.py
+++ b/control/config.py
@@ -97,6 +97,9 @@ def reset_defaults():
     from .rlocus import _rlocus_defaults
     defaults.update(_rlocus_defaults)
 
+    from .namedio import _namedio_defaults
+    defaults.update(_namedio_defaults)
+
     from .xferfcn import _xferfcn_defaults
     defaults.update(_xferfcn_defaults)
 
--- a/control/dtime.py
+++ b/control/dtime.py
@@ -47,7 +47,8 @@ $Id: dtime.py 185 2012-08-30 05:44:32Z murrayrm $
 
 """
 
-from .namedio import isctime
+from .namedio import isctime, _process_namedio_keywords
+from .iosys import ss 
 from .statesp import StateSpace
 
 __all__ = ['sample_system', 'c2d']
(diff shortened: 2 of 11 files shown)


murrayrm:
I suggest following the conventions used in `linearize()`:
* Use `copy` keyword to control whether signal names are copied (default can be True, although for `linearize` it is False for some reason).
* Use `name` keyword to set the system name.  If `name` is not specified and if `copy` is `False`, a generic name <sys[id]> is generated with a unique integer id.  If `copy` is `True`, the new system name is determined by adding the prefix and suffix strings in `config.defaults['iosys.linearized_system_name_prefix']` and `config.defaults['iosys.linearized_system_name_suffix']`, with the default being to add the suffix '$linearized'.

Doing this for `sample()`, the default name of the new system could be the old system with `$sampled` appended (and overridable via something like `iosys.sample_system_name_prefix` and `iosys.sample_system_name_suffix`.

Also, we might want to override the `sample()` method in `LinearIOSystem` so that sampling a linear input/output system gives a linear input/output system (rather than just a `StateSpace` system).  This won't matter much since `StateSpace` is converted to `LinearIOSystem` if it is connected to anything, but would be a bit cleaner (and points to the fact that we should probably merge `StateSpace` and `LinearIOSystem` at some point, as pointed out already in issue #786).

sawyerbfuller:
@murrayrm thanks for the pointer. I refactored the code used in `iosys.linearize` to be a general-use method `_copy_names` that now lives in `namedio` that does the copying of various dicts correctly over to a new system.

Other changes: 
* moved naming defaults associated with linearization and sampling to `namedio` because they are related to naming. OK? 

Questions: 
1. I moved naming defaults associated with linearization and sampling to `namedio` because they are related to naming. OK? 
2. I changed the keyword to copy names to `copy_names` from `copy` in `linearize` and `sample` because I thought `copy` might somehow suggest copying the whole system somehow and this better represented what it should do. This does (slightly) break backward compatibility. OK? 
3. the new `sample` allows copying the names but also supplying new keyword arguments that additionally allow those signal labels to be overridden. This is suggested for how it works in the `linearize` docstrings but what actually happens is that `copy` ignored any keyword-supplied names. I changed this so that supplied names override copied names. OK? 
4. `sample` copies the names over by default because that is how they are likely to be used, but currently `linearize` has the old behavior of not copying over the names by default, but I am not sure if this is the best.  sound reasonable?

murrayrm:
I like the revisions and pretty much agree with them all, especially moving the functionality up to namedio.py (which is where it belongs).  Some additional thoughts:

* One way to preserve backward compatibility would be to use `kwargs` to allow `copy` to still be used, but with a deprecation warning.
* Rather than `$copy`, it seems like `$sampled` would be a better choice for `sample_system()`.
* We might want to create a `copy` method for `NamedIOSystem` and use the `$copy` suffix there.

sawyerbfuller:
Great. New commit:
* simplifies `_copy_names` to remove unnecessary functionality 
* sets default behavior to `copy_names=True` in `iosys.linearize`. happy to revert it if you think it's better not to. 
* fix docstring errors
* more unit tests

Following @murrayrm's comment above: 
* `linearize` now accepts `copy` keyword, with a deprecation warning 
* `namedio.copy` now does a deepcopy instead of a shallow copy, addressing #728. Before, all of its dicts and arrays were just references to objects that could be modified. The new system has `$copy` appended to its name, as before. 
* `dtime.sample_system` now gets all of the naming functionality that is now present in `sys.sample`
* `sample` and `sample_system` append `$sampled` to the system name by default

murrayrm:
This all looks good to me.  Will let it sit for another day or two just in case someone else wants to take a look, but I think it is ready to merge.

murrayrm (approved):

PR #798, opened 2022-11-24
Fix find_eqpt when y0 is None

adswid:
When `y0` is `None`, `dy` in the root function should not be calculated (since it tries to subtract `NoneType`).  In general, when `y0` is `None`, the evaluation of the output function is not necessary.


murrayrm:
LGTM.

Issue #782, opened 2022-10-18
About the "horizon" parameter in the solve_ocp() method


zhengzl18:
### Discussed in https://github.com/python-control/python-control/discussions/781

<div type='discussions-op-text'>

<sup>Originally posted by **zhengzl18** October 17, 2022</sup>
Hi,
I'm running the example code of the `optimal` module: https://python-control.readthedocs.io/en/0.9.2/optimal.html#example
When calling the `solve_ocp` method, a parameter called "horizon" is required:
```
horizon = np.linspace(0, Tf, 3, endpoint=True)
result = opt.solve_ocp(
    vehicle, horizon, x0, traj_cost, constraints,
    terminal_cost=term_cost, initial_guess=u0)
```
In this setting, it works just fine.

But when I tried to change the horizon into something like (changing nothing but the number of points in the linspace):
```
horizon = np.linspace(0, Tf, 4, endpoint=True)
```
the result seemed to be extremely strange, with the inputs u always remaining the same:
```
Summary statistics:
* Cost function calls: 9
* Constraint calls: 20
* System simulations: 27
* Final cost: 16000.0
    cost: 16000.0
     fun: 16000.0
  inputs: array([[10., 10., 10., 10.],
       [ 0.,  0.,  0.,  0.]])
     jac: array([ 0.00000000e+00,  2.44140625e-04,  1.22070312e-04,  0.00000000e+00,
       -3.62099047e+06, -6.84054738e+06, -2.12256850e+06, -7.49211857e+05])
 message: 'Optimization terminated successfully'
    nfev: 9
     nit: 5
    njev: 1
 problem: <control.optimal.OptimalControlProblem object at 0x7ff0cb281df0>
  states: array([[  0.        ,  33.33333333,  66.66666667, 100.        ],
       [ -2.        ,  -2.        ,  -2.        ,  -2.        ],
       [  0.        ,   0.        ,   0.        ,   0.        ]])
  status: 0
 success: True
    time: array([ 0.        ,  3.33333333,  6.66666667, 10.        ])
       x: array([10., 10., 10., 10.,  0.,  0.,  0.,  0.])

```

It's the case for all `num` except 3 in the np.linspace(0, Tf, num).

I'm confused by this results and wonder what's the actual use of 'horizon'? (I do notice from the comments that it's a "List of times at which the optimal input should be computed", but I still can't understand)

Hope for any help.

Thanks!</div>

PR #799, opened 2022-11-26
Add collocation method for optimal control problems

murrayrm:
This PR changes the default way that optimal control problems are solved from using a shooting method to using collocation.  This results in more robust (and faster) solutions, especially for trajectory generation for unstable or marginally stable systems (where shooting is very sensitive to small parameter variations).

The following new functionality is included:

* The `solve_ocp()` function and `OptimalControlProblem` class have new parameters `trajectory_method` that be set to either `shooting` (default for discrete time systems) or `collocation` (default for continuous time systems). 
* When collocation is used, the `initial_guess` parameter can either be an input trajectory (as before) or a tuple consisting of a state trajectory and an input trajectory.
* Constraints can now be specified in the form `LinearConstraint(A, lb, ub)` or `NonlinearConstraint(fun, lb, ub)` instead of the previous forms `(LinearConstraint, A, lb, ub)` and `(NonlinearConstraint, fun, lb, ub)`.  The legacy format is still supported.
* The documentation (docstrings and manual) have been updated to reflect the new functionality.
* New unit tests are included that tease out the functionality of the various possibilities.
* Benchmarks comparing shooting and collocation have been added.

This PR also addresses issue #782 (see also discussion in #781) which were showing some of the weakness of the shooting methodology.


sawyerbfuller:
LGTM

murrayrm:
Updated the initial post to reflect the additional commits, so that we don't lose track of the changes.

PR #804, opened 2022-11-30
docstring improvements

sawyerbfuller:
Improvements needed when looking at the html function reference
* add `NonlinearIOSystem` to system creation section, `interconnect` to interconnection section
* remove `evalfr` and `freqresp` (they still live in matlab module). Replace them with `__call__` methods and `frequency_response` respectively. The former ends up being a bit verbose, I'm open to other ideas for how to present this.  
* updates `lyap` and `care` descriptions to be consistent with other functions
* misc other formatting and consistency issues in various functions

### DIFF
--- a/control/dtime.py
+++ b/control/dtime.py
@@ -84,6 +84,14 @@ def sample_system(sysc, Ts, method='zoh', alpha=None, prewarp_frequency=None,
     copy_names : bool, Optional
         If True, copy the names of the input signals, output
         signals, and states to the sampled system.
+
+    Returns
+    -------
+    sysd : linsys
+        Discrete time system, with sampling rate Ts
+
+    Other Parameters
+    ----------------
     inputs : int, list of str or None, optional
         Description of the system inputs.  If not specified, the origional
         system inputs are used.  See :class:`InputOutputSystem` for more
@@ -93,11 +101,6 @@ def sample_system(sysc, Ts, method='zoh', alpha=None, prewarp_frequency=None,
     states : int, list of str, or None, optional
         Description of the system states.  Same format as `inputs`. Only
         available if the system is :class:`StateSpace`.
-
-    Returns
-    -------
-    sysd : linsys
-        Discrete time system, with sampling rate Ts
 
     Notes
     -----
--- a/control/dtime.py
+++ b/control/dtime.py
@@ -84,14 +84,6 @@ def sample_system(sysc, Ts, method='zoh', alpha=None, prewarp_frequency=None,
     copy_names : bool, Optional
         If True, copy the names of the input signals, output
         signals, and states to the sampled system.
-
-    Returns
-    -------
-    sysd : linsys
-        Discrete time system, with sampling rate Ts
-
-    Additional Parameters
-    ---------------------
     inputs : int, list of str or None, optional
         Description of the system inputs.  If not specified, the origional
         system inputs are used.  See :class:`NamedIOSystem` for more
@@ -101,6 +93,11 @@ def sample_system(sysc, Ts, method='zoh', alpha=None, prewarp_frequency=None,
     states : int, list of str, or None, optional
         Description of the system states.  Same format as `inputs`. Only
         available if the system is :class:`StateSpace`.
+
+    Returns
+    -------
+    sysd : linsys
+        Discrete time system, with sampling rate Ts
 
     Notes
     -----
--- a/control/dtime.py
+++ b/control/dtime.py
@@ -86,7 +86,7 @@ def sample_system(sysc, Ts, method='zoh', alpha=None, prewarp_frequency=None,
         signals, and states to the sampled system.
     inputs : int, list of str or None, optional
         Description of the system inputs.  If not specified, the origional
-        system inputs are used.  See :class:`NamedIOSystem` for more
+        system inputs are used.  See :class:`InputOutputSystem` for more
         information.
     outputs : int, list of str or None, optional
         Description of the system outputs.  Same format as `inputs`.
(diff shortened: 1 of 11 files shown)


sawyerbfuller:
old: 
image

image

new: 
image

image

control/iosys.py
```diff
@@ -2203,24 +2203,21 @@ def linearize(sys, xeq, ueq=None, t=0, params=None, **kw):
...
         states to the linearized system.
-
-    Returns
-    -------
-    ss_sys : LinearIOSystem
-        The linearization of the system, as a :class:`~control.LinearIOSystem`
-        object (which is also a :class:`~control.StateSpace` object.
-
-    Additional Parameters
-    ---------------------
```
murrayrm, line 2214: As above, this should be named "Other Parameters" and should go _after_ the "Returns" section.

control/statesp.py
```diff
@@ -1338,14 +1338,6 @@ def sample(self, Ts, method='zoh', alpha=None, prewarp_frequency=None,
...
             If True, copy the names of the input signals, output
             signals, and states to the sampled system.
-
-        Returns
-        -------
-        sysd : StateSpace
-            Discrete-time system, with sampling rate Ts
-
-        Additional Parameters
-        ---------------------
```
murrayrm, line 1348: As above, this should be named "Other Parameters" and should go _after_ the "Returns" section.

control/xferfcn.py
```diff
@@ -1130,21 +1130,18 @@ def sample(self, Ts, method='zoh', alpha=None, prewarp_frequency=None,
...
             If True, copy the names of the input signals, output
             signals, and states to the sampled system.
-
-        Returns
-        -------
-        sysd : TransferFunction system
-            Discrete-time system, with sample period Ts
-
-        Additional Parameters
-        ---------------------
```
murrayrm, line 1140: As above, this should be named "Other Parameters" and should go _after_ the "Returns" section.

doc/control.rst
```diff
@@ -78,8 +79,9 @@ Control system analysis
 
     dcgain
     describing_function
-    evalfr
-    freqresp
+    frequency_response
+    TransferFunction.__call__
+    StateSpace.__call__
```
murrayrm, line 84: I agree this format is a bit long and cumbersome.  Would it make sense to have this show up as something more like `sys(x[, squeeze])` or something like that.  This can be done by writing the desired signature as the first line of the docstring, as described [here](https://numpydoc.readthedocs.io/en/latest/format.html).

doc/control.rst
```diff
@@ -114,10 +116,10 @@ Control system synthesis
 
     acker
     create_statefbk_iosystem
-    dlqr
     h2syn
     hinfsyn
     lqr
+    dlqr
```
murrayrm, line 122: I understand why this might go here, but doesn't it make more sense to list alphabetically? 

Alternatively, we could list this as `lqe, dlqe`, since they go together?

Note that if you pass a `LTI` object to `lqr` (and `lqe`), the function will check the timebase to determine whether to use the continuous time or discrete time version of the function.

doc/control.rst
```diff
@@ -156,8 +157,8 @@ Stochastic system support
 
     correlation
     create_estimator_iosystem
-    dlqe
```
murrayrm, line 159: Alphabetical?

doc/matlab.rst
```diff
@@ -86,6 +86,9 @@ Compensator design
    sisotool
    place
    lqr
+   dlqr
+   lqe
+   dlqe
```
murrayrm, line 91: Should these be in alphabetical order?

murrayrm (commented):
Good catches on the documentation.  I think we should comply with the [numpydoc style guide](https://numpydoc.readthedocs.io/en/latest/format.html), which involves renaming "Additional Parameters" to "Other Parameters" and leaving them _after_ the "Returns" section.

control/dtime.py
```diff
@@ -84,24 +84,21 @@ def sample_system(sysc, Ts, method='zoh', alpha=None, prewarp_frequency=None,
...
         Description of the system outputs.  Same format as `inputs`.
     states : int, list of str, or None, optional
         Description of the system states.  Same format as `inputs`. Only
         available if the system is :class:`StateSpace`.
 
+    Returns
+    -------
+    sysd : linsys
+        Discrete time system, with sampling rate Ts
+
```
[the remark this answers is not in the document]
sawyerbfuller: Thanks. I had noticed that there were a few places where the docstrings in the library listed "additional Parameters" and sphinx was not handling it right. So I took it out. And since making this commit I discovered that sphinx was looking for "other parameters."  I'll look around for any more instances of "additional parameters" and change them back to "other parameters"

doc/control.rst
sawyerbfuller: Good idea. I tried to rename the call string for `TransferFunction.__call__` in the docstring as suggested in  numpydoc but no luck: 

image

Will see if I can find any examples in numpy doc itself I can use

sawyerbfuller: Ok looked around in numpy and scipy and found `scipy.interpolate.interp1d`, which is a class that has a `__call__` method. But it is not linked in a function reference like this, so it did not provide a way to change how the call signature is shown. Perhaps there is a way to tell sphinx to manually change its call signature, but I coudln't find it. 

Unless anybody else has other ideas about how to refer to this, I would propose to leave it as it is in this PR and plan to update later if somebody figures something out.

doc/control.rst
sawyerbfuller: re-alphebetized them in recent commit.

murrayrm (commented):
LGTM

control/matlab/__init__.py
```diff
@@ -84,6 +84,7 @@
 from ..rlocus import rlocus
 from ..dtime import c2d
 from ..sisotool import sisotool
+from ..stochsys import *
```
bnavigator, line 87: Can we make these explicit imports please?

murrayrm: I think the two functions that we want (since they are present in MATLAB) are lqe and dlqe.  The other functions are unique to python-control.

PR #809, opened 2022-12-07
fix gain handling in rlocus and sisotool

sawyerbfuller:
current problem is that if you want to pass an initial gain to `sisotool` to plot as starting point gain in the various plots, an error is raised if it is not an array. so you have to do `sisotool(sys, kvect=(1.1, ))` which is awkward. With this fix, <s>you can do `sisotool(sys, kvect=1.1)`</s> update: changed function call to look like `sisotool(sys, initial_gain=1.1)`

### DIFF
--- a/control/config.py
+++ b/control/config.py
@@ -97,6 +97,9 @@ def reset_defaults():
     from .rlocus import _rlocus_defaults
     defaults.update(_rlocus_defaults)
 
+    from .sisotool import _sisotool_defaults
+    defaults.update(_sisotool_defaults)
+
     from .namedio import _namedio_defaults
     defaults.update(_namedio_defaults)
 
--- a/control/matlab/wrappers.py
+++ b/control/matlab/wrappers.py
@@ -117,7 +117,7 @@ def nyquist(*args, **kwargs):
 def _parse_freqplot_args(*args):
     """Parse arguments to frequency plot routines (bode, nyquist)"""
     syslist, plotstyle, omega, other = [], [], None, {}
-    i = 0;
+    i = 0
     while i < len(args):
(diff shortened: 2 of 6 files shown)


sawyerbfuller:
Added some more fixes to the `kvect` kwarg in rlocus and sisotool. `kvect` was intended to let the user specify their own array of gains, but it wasn't being used that way. Two primary changes aim to make it do what I believe was its intended purpose: 

1. I deprecated `kvect` in sisotool, because you probably don't want to specify your own set of gains to plot in the rlocus plot in sisotool anyway. You probably want to use the ones that are computed automatically. Especially because sisotool/rlocus allows you to zoom in, and it re-computes gain points for the new zoomed root locus plot, to get a more precise click. Instead, there is now a new keyword, `initial_gain` that allows the user to specify the starting value of the closed-loop gain to plot in its four plots. `kvect` can be still given, and it reproduces the old behavior of sisotool using `kvect[0]` as the initial gain, and ignoring everything else in `kvect`. And there is now a deprecation warning. 

2.  For rlocus, you might actually want to specify your own `kvect` if for example you only want to see where the roots go for a certain set of gains. But this wasn't what was happening: the rlocus plot ignores the `kvect` argument altogether and essentially always uses its own pre-computed gains. This is because any time the axis range is changed, `kvect` is automatically recomputed, which is basically every time the plot is created.  I fixed it so that the new behavior is that if `kvect` is supplied to rlocus, it is no longer recomputed automatically if the plot is zoomed. 

3. Other improvements: 
* docstring fixes
* pep8 cleanups
* change `mymat` to `root_array` in rlocus to better describe its function

murrayrm (commented):
LGTM

PR #811, opened 2022-12-10
Enable scalar division of state-space objects

roryyorke:
I think this is a reasonable if minor improvement; it lets one write `gscaled = g / 2` instead of `gscaled = g * (1/2)`.  I'm less sure about division by arrays, but I see multiplication by arrays is allowed, and it seems natural to have symmetry between these operations.

I added a simple test for `StateSpace / scalar` and `StateSpace / array`, but not for `InputOutputSystem / scalar`. etc.

I also modified the type-conversion tests as required.

### DIFF
--- a/control/iosys.py
+++ b/control/iosys.py
@@ -347,7 +347,9 @@ class InputOutputSystem(NamedIOSystem):
         return newsys
 
     def __truediv__(sys2, sys1):
-        """Multiply two input/output systems (series interconnection)"""
+        """Division of input/output systems
+
+        Only division by scalars and arrays of scalars is supported"""
         # Note: order of arguments is flipped so that self = sys2,
         # corresponding to the ordering convention of sys2 * sys1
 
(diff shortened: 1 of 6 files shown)


murrayrm (commented):
LGTM.

control/iosys.py
```diff
@@ -346,6 +346,17 @@ def __neg__(sys):
         # Return the newly created system
         return newsys
 
+    def __truediv__(sys2, sys1):
+        """Multiply two input/output systems (series interconnection)"""
```
sawyerbfuller, line 350: should this docstring say "divide"?

sawyerbfuller (changes requested):
looks good to me - with one possible update in a docstrong

roryyorke:
Whoops, yes. Should be similar to StateSpace truediv docstring. Earliest I
can look at this is Friday.

PR #812, opened 2022-12-10
Update docstring for impulse for discrete sys

sawyerbfuller:
Update impulse_response doc string to indicate that the impulse size is unit area for discrete systems.

### DIFF
--- a/control/timeresp.py
+++ b/control/timeresp.py
@@ -1796,7 +1796,8 @@ def impulse_response(sys, T=None, X0=0., input=None, output=None, T_num=None,
     -----
     This function uses the `forced_response` function to compute the time
     response. For continuous time systems, the initial condition is altered to
-    account for the initial impulse.
+    account for the initial impulse. For discrete-time aystems, the impulse is 
+    sized so that it has unit area.
 
     Examples
     --------


murrayrm:
We definitely need to update the documentation, but I wonder whether we shouldn't also consider what the right definition of the impulse response is for a discrete time system with non-unit sampling time.  MATLAB apparently defines the impulse response for a discrete time system as being the response to an input that is 1 at the initial time and then 0 after that.  This matches the description in the [Wikipedia entry for impulse response][(https://en.wikipedia.org/wiki/Impulse_response#Mathematical_considerations).

sawyerbfuller:
I made this change a little while back to match what matlab does, which is a unit-area impulse for discrete-time systems: https://www.mathworks.com/help/control/ref/lti.impulse.html. This arrangement is probably what you want if you are viewing a discrete-time system as a discretized model of a continuous-time system. 

Probably unit area is favored by engineering, unit magnitude by math.

Probably a subjective choice. Anybody here have further insight into such a choice?

ilayn:
As the sampling frequency grows the response starts to get larger if impulse is kept at unity. Hence the same dynamical system sampled at different frequencies will give different results with the same unity impulse which is not correct. In a hand-waving argument we have to scale the energy of the impulse with frequency to have equivalent supply. Otherwise, we need to type math here which is not fun :)

You can test it yourself by generating a [1, 0, 0, 0, ...] array and supply to `lsim` for different sampled versions of the same system to see the effect I am attempting to describe above.

murrayrm:
@ilayn's comment makes it clear why scaling by the sampling time is the right approach (and I was mistaken about what MATLAB does).

cdboschen:
Personally (speaking from an engineer modelling discrete time systems) I don't like the additional scaling. I wrote #974 to share my perspective. Trying to bottom line with an example similar to @ilayn's comment: If I had an accumulator as the output is added to the next input $H(z) = \frac{z}{z-1}$ then having a different result for different sampling rates is exactly what I would expect to happen. If I don't want that, I set $dt=1$ and work in time units of samples instead of seconds. (Similar to why we choose to work in fractional frequency as cycles per sample instead of Hz or radians/sample instead of seconds). The system I would build would have the scaling to match the continuous time result we are trying to achieve. And when I model that I would include that scaling in my model to match my actual implementation. The idea of having the result scale by $dt$ or $1/dt$ makes sense when we are using the tool to map between continuous and discrete time domains such as the c2d function but doesn't yet make sense to me to do that when we are already discrete; to me the impulse response should match the definition of "unit sample response" for that system based on its description.

PR #817, opened 2022-12-18
Fix readthedocs to use pip-based install 

murrayrm:
This PR fixes the build process on readthedocs to use pip-based installed instead of the deprecated setup.py.  I also pinned docutils to v0.16 to fix bullets, as documented in [readthedocs/sphinx_rtd_theme issue #1115](https://github.com/readthedocs/sphinx_rtd_theme/issues/1115).

A successful build of the documentation can be found [here](https://murrayrm-python-control.readthedocs.io/en/readthedocs-17dec2022/).


.readthedocs.yaml
```diff
@@ -0,0 +1,21 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the version of Python and other tools you might need
+build:
+  os: ubuntu-20.04
```
bnavigator, line 10: Is there a reason not to use a more recent Ubuntu 22.04? Ubuntu 20.04 gets freely available updates only for two more years, and it's likely that readthetdocs will drop it then.

murrayrm: This was the default from the [readthedocs documentation](https://docs.readthedocs.io/en/stable/config-file/v2.html#build-os), but it looks like `ubuntu-22.04` is also supported, so I'll bump it up to that.

Issue #820, opened 2022-12-21
Flaky test failure in TestFlatSys with current slycot


bnavigator:
From the Slycot CI on a few runs of the test matrix, but not all:

```python
2022-12-21T09:54:30.8524100Z =================================== FAILURES ===================================
2022-12-21T09:54:30.8524781Z ______ TestFlatSys.test_kinematic_car_ocp[basis6-None-constraints6-None] _______
2022-12-21T09:54:30.8525337Z 
2022-12-21T09:54:30.8525622Z self = <control.tests.flatsys_test.TestFlatSys object at 0x7ffa1993a430>
2022-12-21T09:54:30.8526148Z vehicle_flat = <FlatSystem:sys[3472]:['v', 'delta']->['x', 'y', 'theta']>
2022-12-21T09:54:30.8526560Z basis = <BSplineFamily: nvars=None, degree=[8], smoothness=[7]>, guess = None
2022-12-21T09:54:30.8527014Z constraints = ([8, -0.1], [12, 0.1]), method = None
2022-12-21T09:54:30.8527252Z 
2022-12-21T09:54:30.8527434Z     @pytest.mark.parametrize(
2022-12-21T09:54:30.8527773Z         "basis, guess, constraints, method", [
2022-12-21T09:54:30.8528179Z         (fs.PolyFamily(8, T=10), 'prev', None, None),
2022-12-21T09:54:30.8528603Z         (fs.BezierFamily(8, T=10), 'linear', None, None),
2022-12-21T09:54:30.8528964Z         (fs.BSplineFamily([0, 10], 8), None, None, None),
2022-12-21T09:54:30.8529417Z         (fs.BSplineFamily([0, 10], 8), 'prev', None, 'trust-constr'),
2022-12-21T09:54:30.8529871Z         (fs.BSplineFamily([0, 10], [6, 8], vars=2), 'prev', None, None),
2022-12-21T09:54:30.8530326Z         (fs.BSplineFamily([0, 5, 10], 5), 'linear', None, 'slsqp'),
2022-12-21T09:54:30.8530807Z         (fs.BSplineFamily([0, 10], 8), None, ([8, -0.1], [12, 0.1]), None),
2022-12-21T09:54:30.8531174Z         (fs.BSplineFamily([0, 5, 10], 5, 3), None, None, None),
2022-12-21T09:54:30.8531474Z     ])
2022-12-21T09:54:30.8531750Z     def test_kinematic_car_ocp(
[...]

PR #821, opened 2022-12-24
Add test matrix against operating environments

murrayrm:
This PR adds a GitHub action to run unit tests against a (large) matrix of different environments, including operating system, BLAS variant, and installation/build method (pip, conda), copied over from slycot.  This is in response to issue #820, where an error was triggered under Ubuntu using the generic BLAS library for NumPy 1.24.0.

The matrix of tests is quite time consuming (about 45 minutes), and so the action is only trigged on pull requests (so if you are using GitHub actions in your own forks, they won't be affected).

In the initial PR, the errors from issue #820 are present (so the tests will fail).  In a subsequent push to this PR, I added a "fix" by doing an xfail on that specific test case (and demonstrating one way this set of test might be used).

Also: when a failure occurs in the flatsys test case, information is printed out about the configuration to help in identifying the cases that are failing.

### DIFF
--- a/.github/conda-env/build-env.yml
+++ b/.github/conda-env/build-env.yml
@@ -1,0 +1,4 @@
+name: build-env
+dependencies:
+  - boa
+  - numpy !=1.23.0
--- a/.github/conda-env/test-env.yml
+++ b/.github/conda-env/test-env.yml
@@ -1,5 +1,6 @@
 name: test-env
 dependencies:
+  - conda-build  # for conda index
   - pip
   - coverage
   - coveralls
--- a/.github/scripts/set-conda-test-matrix.py
+++ b/.github/scripts/set-conda-test-matrix.py
@@ -1,0 +1,34 @@
+""" set-conda-test-matrix.py
+
+Create test matrix for conda packages
+"""
+import json, re
+from pathlib import Path
+
+osmap = {'linux': 'ubuntu',
+         'osx': 'macos',
+         'win': 'windows',
+         }
+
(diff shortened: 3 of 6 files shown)


murrayrm:
Added checks + xfail for Ubuntu with generic BLAs with numpy 1.24.0, in response to issue #820. Near as I can tell, this is some sort of edge case that gets triggered under this specific set of circumstances.  Not sure this is the right way to handle it, but given that all of the other environments work out, I don't think it is a bug in `solve_flat_ocp`.

bnavigator:
python-control is pure Python and *should* not depend on the environment. Problems in the environment usually show up when the consumed libraries change, not during python-control PRs.

We already go through that test matrix in the Slycot CI and run the python-control test suite there. Not sure if we really should duplicate the resource consumption here.

murrayrm:
I agree this _shouldn't_ be needed @bnavigator, but the failure pointed out in issue #820 is an example of where python-control does depend on the operating environment, because there is some sort of (non-slycot) interaction with something in the build environment (and only under Ubuntu, NumPy 1.24, generic BLAS). 

I've changed the trigger to workflow_dispatch, which allows this to be manually triggered on the main branch, plus an PRs that affect files specifically associated with this workflow.  That will get rid of the resource consumption except in the rather exceptional cases where we need it.  (I also imagine that we might customize the test matrix here over time, so it could diverge from slycot.)

My current plan is to add a step in the [release Instructions](https://github.com/python-control/python-control/wiki/Release-instructions) suggesting running this workflow before a release, just to flag anything where python-control depends on the details of the OS, BLAS version, or installation method.

PR #833, opened 2022-12-31
remove __commit__ from version test

murrayrm:
When I did the release for v0.9.3, I got `control.__version__` as 'dev'.  It looks like this occurred because we try to import `__commit__` from `control/_version.py` but that information is not actually included there by setuptools_scm.

I'll re-release with this change in a bit.

### DIFF
--- a/control/__init__.py
+++ b/control/__init__.py
@@ -79,7 +79,7 @@ from .exception import *
 
 # Version information
 try:
-    from ._version import __version__, __commit__
+    from ._version import __version__
 except ImportError:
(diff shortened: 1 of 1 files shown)


murrayrm:
Update: it is not possible to re-issues a release, so this will have to become release 0.9.4.

bnavigator:
You can release 0.9.3.post1

bnavigator:
@murrayrm, IMHO the release title on GitHub should include the version number:
image

PR #849, opened 2023-01-14
Small fixes based on Caltech CDS 112 course

murrayrm:
This PR contains some small fixes that I discovered while teaching CDS 112 at Caltech in Winter 2023:

* Fixed bug in `create_statefbk_iosystem` that didn't proper handle 1D gain schedules.
* `flatsys.point_to_point` generates an error if you specify a cost or constraint with less than 3 time points.
* Sampling a `LinearIOSystem` returns a `LinearIOSystem` (before it returned a `StateSpace` system).
* Added an `add_unused` keyword parameter to `interconnect` that allows unused inputs or outputs to be added as inputs or outputs of the interconnected system (useful for doing a "partial" interconnection).
* `create_mpc_iosystem` now allows system and signal names to be specified via appropriate keywords.
* Changed `type` keyword for `create_statefbk_iosystem` to `controller_type` (linear or nonlinear).  Using `type` generates a deprecation warning. 
* Added `control_indices` and `state_indices` to `create_statefbk_iosystem` to allow partial interconnection (eg, for inner/outer loop construction).
* Updated unit tests, docstrings, and comments.


control/statefbk.py
```diff
@@ -684,7 +685,7 @@ def create_statefbk_iosystem(
         hull of the scheduling points, the gain at the nearest point is
         used.
 
-    type : 'linear' or 'nonlinear', optional
+    controller_type : 'linear' or 'nonlinear', optional
```
bnavigator, line 688: That's an API change. The milestone should be 0.10.0 (or 1.0.0) then,

murrayrm: I put in code to handle legacy calls, so `type` will still work.  See line 732.

bnavigator:
Rebased onto current main, in order to resolve conflicts due to #857

Issue #862, opened 2023-02-17
bode_plot phase wrapping incorrect for multiple systems.


henklaak:
Minimum example:
```
# two RHP's
G = zpk([],[1,1], gain=1)  
bode_plot(G)
bode_plot((G,G))
```

The second call shows a figure with different phases for each (identical) system.

![bode1](https://user-images.githubusercontent.com/3962727/219655726-fae608b9-6cbc-4d16-a1f3-69d846910276.png)
![bode2](https://user-images.githubusercontent.com/3962727/219655742-e6b12471-61fd-4fed-92a3-e9fef5ee25a4.png)

henklaak:
I'm working a PR for it.

murrayrm:
Fixed in #863

PR #863, opened 2023-02-17
Solve #862 and #864: bode_plot phase wrapping incorrect for multiple systems

henklaak:
Function argument clobbered by local assignment. Use local variable instead.


henklaak:
Need an extra pair of eyes here. Can't relate the errors to the PR work.

henklaak:
matplotlib 3.6.3 => 3.7.* seems to break our tests. Will investigate further.

henklaak:
Solves #862 and #864. If you prefer, I can make separate PRs.

murrayrm:
Thanks for the fixes!

PR #869, opened 2023-02-21
Feature print zpk

### LINKED ISSUE #64: Feature request pretty printing zpk form
It would be nice to be able to print TransferFunction instances in zero-pole-gain form. I haven't found this functionality if it exists. Maybe a simple `zpk_print()` method would suffice.

henklaak:
Fixes #64

```
G = tf([2, 6, 4],[1,-2,1], display_format='poly')
print(G)

G = tf([2, 6, 4],[1,-2,1], display_format='zpk')
print(G)

G = tf([2, 6, 4],[1,-2,1], dt=0.1, display_format='poly')
print(G)

G = tf([2, 6, 4],[1,-2,1], dt=0.1, display_format='zpk')
print(G)

```
Produces:
```

2 s^2 + 6 s + 4
---------------
 s^2 - 2 s + 1


2 (s + 1) (s + 2)
-----------------
 (s - 1) (s - 1)


2 z^2 + 6 z + 4
---------------
 z^2 - 2 z + 1

dt = 0.1


2 (z + 1) (z + 2)
-----------------
 (z - 1) (z - 1)

dt = 0.1
```

### DIFF
--- a/control/xferfcn.py
+++ b/control/xferfcn.py
@@ -505,9 +505,13 @@ class TransferFunction(LTI):
                 if abs(polygain) == 0 and abs(dcgain) == 0:
                     k = 1
                 else:
+                    if abs(polygain) == 0:
+                        raise ValueError(
+                            f"Transfer function has infinite gain. "
+                            "Please check polynomials.")
                     k = dcgain/polygain
                 if not np.isreal(k):
-                    raise ValueError("Transfer function has complex valued gain. "
+                    raise ValueError(f"Transfer function has complex valued gain (k = {k}). "
                                      "Please check polynomials for non-complimentary poles.")
 
                 k = np.abs(k)
--- a/control/xferfcn.py
+++ b/control/xferfcn.py
@@ -78,6 +78,7 @@ def _float2str(value):
     formatter = "{:" + config.defaults.get('xferfcn.floating_point_format', ':.4g') + "}"
     return formatter.format(value)
 
+
 class TransferFunction(LTI):
     """TransferFunction(num, den[, dt])
 
--- a/control/xferfcn.py
+++ b/control/xferfcn.py
@@ -75,8 +75,8 @@ _xferfcn_defaults = {
 }
 
 def _float2str(value):
-    formatter = "{:" + config.defaults.get('xferfcn.floating_point_format', ':.4g') + "}"
-    return formatter.format(value)
+    _num_format = config.defaults.get('xferfcn.floating_point_format', ':.4g')
+    return f"{value:{_num_format}}"
 
 
 class TransferFunction(LTI):
--- a/control/xferfcn.py
+++ b/control/xferfcn.py
@@ -71,8 +71,12 @@ __all__ = ['TransferFunction', 'tf', 'zpk', 'ss2tf', 'tfdata']
 # Define module default parameter values
 _xferfcn_defaults = {
     'xferfcn.display_format': 'poly',
+    'xferfcn.floating_point_format': '.4g'
 }
 
+def _float2str(value):
+    formatter = "{:" + config.defaults.get('xferfcn.floating_point_format', ':.4g') + "}"
+    return formatter.format(value)
 
 class TransferFunction(LTI):
     """TransferFunction(num, den[, dt])
@@ -1313,7 +1317,7 @@ def _tf_polynomial_to_string(coeffs, var='s'):
     N = len(coeffs) - 1
 
     for k in range(len(coeffs)):
-        coefstr = '%.4g' % abs(coeffs[k])
+        coefstr = _float2str(abs(coeffs[k]))
         power = (N - k)
         if power == 0:
             if coefstr != '0':
@@ -1355,7 +1359,7 @@ def _tf_factorized_polynomial_to_string(roots, gain=1, var='s'):
     """Convert a factorized polynomial to a string"""
 
     if roots.size == 0:
-        return f"{gain:.4g}"
+        return _float2str(gain)
 
     factors = []
     for root in sorted(roots, reverse=True):
@@ -1364,29 +1368,29 @@ def _tf_factorized_polynomial_to_string(roots, gain=1, var='s'):
                 factor = f"{var}"
                 factors.append(factor)
             elif root > 0:
-                factor = f"{var} - {np.abs(root):.4g}"
+                factor = f"{var} - {_float2str(np.abs(root))}"
                 factors.append(factor)
             else:
-                factor = f"{var} + {np.abs(root):.4g}"
+                factor = f"{var} + {_float2str(np.abs(root))}"
                 factors.append(factor)
         elif np.isreal(root * 1j):
             if root.imag > 0:
-                factor = f"{var} - {np.abs(root):.4g}j"
+                factor = f"{var} - {_float2str(np.abs(root))}j"
                 factors.append(factor)
             else:
-                factor = f"{var} + {np.abs(root):.4g}j"
+                factor = f"{var} + {_float2str(np.abs(root))}j"
                 factors.append(factor)
         else:
             if root.real > 0:
-                factor = f"{var} - ({root:.4g})"
+                factor = f"{var} - ({_float2str(root)})"
                 factors.append(factor)
             else:
-                factor = f"{var} + ({-root:.4g})"
+                factor = f"{var} + ({_float2str(-root)})"
                 factors.append(factor)
 
     multiplier = ''
     if round(gain, 4) != 1.0:
-        multiplier = f"{gain:.4g} "
+        multiplier = _float2str(gain) + " "
 
     if len(factors) > 1 or multiplier:
         factors = [f"({factor})" for factor in factors]
--- a/control/xferfcn.py
+++ b/control/xferfcn.py
@@ -486,7 +486,7 @@ class TransferFunction(LTI):
                 outstr += "\n" + numstr + "\n" + dashes + "\n" + denstr + "\n"
 
         # If this is a strict discrete time system, print the sampling time
-        if self.isdtime(strict=True):
+        if type(self.dt) != bool and self.isdtime(strict=True):
             outstr += "\ndt = " + str(self.dt) + "\n"
 
         return outstr
--- a/control/xferfcn.py
+++ b/control/xferfcn.py
@@ -92,6 +92,9 @@ class TransferFunction(LTI):
         time, positive number is discrete time with specified
         sampling time, None indicates unspecified timebase (either
         continuous or discrete time).
+    display_format: None, 'poly' or 'zpk'
+        Set the display format used in printing the TransferFunction object.
+		Default behavior is polynomial display.
 
     Attributes
     ----------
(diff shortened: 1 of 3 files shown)


control/xferfcn.py
```diff
@@ -468,6 +468,82 @@ def __str__(self, var=None):
...
+               1
+        ---------------
+        (s - 1) (s - 1)
+
+        """
+
+        mimo = self.ninputs > 1 or self.noutputs > 1
+        if var is None:
+            # TODO: replace with standard calls to lti functions
+            var = 's' if self.dt is None or self.dt == 0 else 'z'
```
sawyerbfuller, line 489:
```suggestion
            var = 's' if self.isctime() else 'z'
```

control/xferfcn.py
```diff
@@ -468,6 +468,82 @@ def __str__(self, var=None):
...
+        >>> from control import tf
+        >>> G = tf([1],[1, -2, 1])
+        >>> G.to_zpk()
+               1
+        ---------------
+        (s - 1) (s - 1)
+
+        """
+
+        mimo = self.ninputs > 1 or self.noutputs > 1
```
sawyerbfuller, line 486: Not sure if it would work as well with the way you have written your logic but there is a `self.issiso()` method available.

control/xferfcn.py
```diff
@@ -468,6 +468,82 @@ def __str__(self, var=None):
...
+                    k = 1
+                else:
+                    if abs(polygain) == 0:
+                        raise ValueError(
+                            f"Transfer function has infinite gain. "
+                            "Please check polynomials.")
+                    k = dcgain/polygain
+                if not np.isreal(k):
+                    raise ValueError(f"Transfer function has complex valued gain (k = {k}). "
+                                     "Please check polynomials for non-complimentary poles.")
```
sawyerbfuller, line 520:
```suggestion
                                     "Please check polynomials for non-complementary poles.")
```

sawyerbfuller (commented):
This would be a very nice feature. I am not sure there are any plans to have a new object class that stores things in zpk form directly, so having convenient methods like this to access that data seems to me like the way to go. 

 A few suggestions/thoughts: 
* might be simpler and more compact on lines 497ish to 523ish to use scipy’s [tf2zpk](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.tf2zpk.html#scipy.signal.tf2zpk). And maybe more robust? Unless your code does it better than how scipy does it. 
* consider refactoring to have this use an intermediate method called something like `zpkdata` that serves as a direct way to access zpk numerical values. 
* In the nice-to-have category would be automatic latex representation when you are in a Jupyter notebook, like how tf and ss systems do it.

control/xferfcn.py
```diff
@@ -468,6 +468,82 @@ def __str__(self, var=None):
...
+
+        mimo = self.ninputs > 1 or self.noutputs > 1
+        if var is None:
+            # TODO: replace with standard calls to lti functions
+            var = 's' if self.dt is None or self.dt == 0 else 'z'
+        outstr = ""
+
+        for i in range(self.ninputs):
+            for j in range(self.noutputs):
+                if mimo:
```
sawyerbfuller, line 494: I think where in the library we use zero-based indexing so we should probably do that here.

sawyerbfuller: (To clarify: this refers to the next line where printing occurs)

control/xferfcn.py
henklaak: Ah, better. Will fix.

control/xferfcn.py
henklaak: US/UK. Will fix.

control/xferfcn.py
henklaak: I copied the behaviour from how TransferFunction prints mimo systems. 
See https://github.com/python-control/python-control/blob/8972e8292aa6705907a43e00fb5f9c9fc9c259e4/control/xferfcn.py#L443

Better be consistent. Opinion?

control/xferfcn.py
henklaak: Will fix.

henklaak:
@sawyerbfuller  Thanks for the review.

The nice-to-have latex would be difficult, since jupyter calls `_repr_latex_` which is currently used for polynomial printing.
These is no obvious way (except introducing a new default boolean)  to switch  between full polynomial / factorized printing. Personally I am not a big fan of the defaults mechanism, as they are hidden from the public API.

Maybe a specific subclass for ZPK, but that would be an invasive change. Let me mull it over.

I'll have a look to see what's possible/convenient for the other suggestions/thoughts.

control/xferfcn.py
sawyerbfuller: I see, good point. Math uses 1-based indexing and python uses 0-based. Luckily there is not much ambiguity here because the first one printed out is index 1. 

I guess my take would be to leave it as it is in that case, like it is elsewhere.

sawyerbfuller:
@henklaak one possibility that might not break anything might be to add a field to TransferFunction, maybe ‘representation’?, that could be set to either ‘poly’ or ‘zpk’ in init. Might be automatically set to ‘zpk’ if instantiated with `zpk`. And output accordingly in response to that flag in `_repr_latex`?

henklaak:
Thanks for the invaluable comments. The PR has become so much better now. 
I had completely overlooked the tf2zpk functionality in SciPy :-(

control/xferfcn.py
```diff
@@ -92,6 +92,9 @@ class TransferFunction(LTI):
         time, positive number is discrete time with specified
```
sawyerbfuller, line 92: I can't suggest changes above this line so I am suggsting them here: 
``` python
_xferfcn_defaults = {
    'xferfcn.display_format': 'poly',
    }
```

control/xferfcn.py
```diff
@@ -198,6 +201,14 @@ def __init__(self, *args, **kwargs):
         #
         # Process keyword arguments
         #
+        if display_format is None:
```
sawyerbfuller, line 204:
```suggestion
        self.display_format = kwargs.pop('display_format', 
                                                          config.defaults['xferfcn.display_format'])
```

control/xferfcn.py
```diff
@@ -198,6 +201,14 @@ def __init__(self, *args, **kwargs):
...
+            display_format = 'poly'
```
sawyerbfuller, line 205:
```suggestion
```

control/xferfcn.py
```diff
@@ -198,6 +201,14 @@ def __init__(self, *args, **kwargs):
...
+
+        if display_format not in ('poly', 'zpk'):
```
sawyerbfuller, line 207:
```suggestion
        if self.display_format not in ('poly', 'zpk'):
```

control/xferfcn.py
```diff
@@ -198,6 +201,14 @@ def __init__(self, *args, **kwargs):
...
+            raise ValueError("display_format must be 'poly' or 'zpk',"
+                             " got '%s'" % display_format)
```
sawyerbfuller, line 209:
```suggestion
                             " got '%s'" % self.display_format)
```

control/xferfcn.py
```diff
@@ -198,6 +201,14 @@ def __init__(self, *args, **kwargs):
...
+
+        self.display_format = display_format
```
sawyerbfuller, line 211:
```suggestion
```

control/xferfcn.py
```diff
@@ -149,7 +152,7 @@ class TransferFunction(LTI):
     # Give TransferFunction._rmul_() priority for ndarray * TransferFunction
     __array_priority__ = 11     # override ndarray and matrix types
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, display_format=None, **kwargs):
```
sawyerbfuller, line 155:
```suggestion
    def __init__(self, *args, **kwargs):
```

control/xferfcn.py
```diff
@@ -92,6 +92,9 @@ class TransferFunction(LTI):
         time, positive number is discrete time with specified
         sampling time, None indicates unspecified timebase (either
         continuous or discrete time).
+    display_format: None, 'poly' or 'zpk'
+        Set the display format used in printing the TransferFunction object.
+        Default behavior is polynomial display.
```
sawyerbfuller, line 97:
```suggestion
        Default behavior is polynomial display and can be changed by 
        changing config.defaults['xferfcn.display_format'].
```

control/xferfcn.py
```diff
@@ -1486,6 +1552,9 @@ def tf(*args, **kwargs):
         Polynomial coefficients of the numerator
     den: array_like, or list of list of array_like
         Polynomial coefficients of the denominator
+    display_format: None, 'poly' or 'zpk'
+        Set the display format used in printing the TransferFunction object.
+        Default behavior is polynomial display.
```
sawyerbfuller, line 1557:
```suggestion
        Default behavior is polynomial display and can be changed by 
        changing config.defaults['xferfcn.display_format'].
```

sawyerbfuller (approved):

henklaak:
I think we have solved this quite nicely.

sawyerbfuller:
In statesp the number format option is     `’statesp.latex_num_format': '.3g'`  Maybe we should standardize on a common name?

sawyerbfuller:
`display_num_format`?

henklaak:
I considered that, but the screen space requirements that lead to the format specifiers are rather unrelated.
How about we work on this in a new issue and PR and discuss it there? Can be done quickly from my side.

control/xferfcn.py
```diff
@@ -69,7 +69,14 @@
 
 
 # Define module default parameter values
-_xferfcn_defaults = {}
+_xferfcn_defaults = {
+    'xferfcn.display_format': 'poly',
+    'xferfcn.floating_point_format': '.4g'
+}
+
+def _float2str(value):
```
ilayn, line 77: I would suggest against doing this. It's anti-pattern for Python and also there is a simpler way to grab the default at the code level and reuse. 

```python
>>> _num_format = config.defaults.get('xferfcn.floating_point_format', ':.4g')
>>> print(f"Bla {0.232654684684654654:{_num_format}}")
Bla 0.232655
```

So you don't need to inject f-string machinery to the string representation.

sawyerbfuller:
@henklaak re display format for text vs latex: makes sense I agree they have different requirements, if I understand what you’re saying. In that case we should have setttings for both latex and text. 

Should they then have otherwise consistent names with only  “latex” vs “print” identifiers that distinguish? That would suggest making this one `print_num_format`? 

if you think it should have a different name then hapoy to leave that discussion to a different  PR. 

Otherwise this looks good to me

henklaak:
Down with the flu. Will continue end of the week.

control/xferfcn.py
murrayrm: @henklaak If you can fix up this final issue, I think this PR is ready to post.

henklaak:
Thanks @murrayrm for taking this one.
Some serious family health issues, I've been a bit low on motivation/interest.
H.

PR #871, opened 2023-02-22
Fix root_locus() handling of ax parameter

### LINKED ISSUE #870: root_locus() does not properly handle ax parameter
Minimal example:

```
import control
import matplotlib.pyplot as pltimport control

tf1 = control.tf([1],[1,1])

fig, ax = plt.subplots(1, 2)
control.root_locus(tf1, ax=ax[0])
control.root_locus(tf1, ax=ax[1])
plt.show()
```

Produces:
bode1

Expected:
bode2

PR  incoming :-)

henklaak:
Fixes #870

### DIFF
--- a/control/rlocus.py
+++ b/control/rlocus.py
@@ -260,7 +260,7 @@ def root_locus(sys, kvect=None, xlim=None, ylim=None,
             if isdtime(sys, strict=True):
                 zgrid(ax=ax)
             else:
-                _sgrid_func(fig=fig if sisotool else None)
+                _sgrid_func(ax)
         else:
             ax.axhline(0., linestyle=':', color='k', linewidth=.75, zorder=-20)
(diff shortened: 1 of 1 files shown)


sawyerbfuller:
This seems like a more straightforward and correct way of implementing sgrid - can you confirm `sisotool` still works as you might expect on that transfer function, and that the gridlines turn off when `rlocus_grid=False`?

Also, we currently have an excess of sgrid functions. There is another : ) It is to be found in `grid.py` and it is called `sgrid`. I haven't compared the two but that one may be more modern/faster? (though it also needs to be upgraded to accept an ax keyword). If so, maybe the `root_locus` function should switch to using that one and delete the one in `rlocus`.

sawyerbfuller:
Looks good to me

PR #874, opened 2023-02-25
Fix editsdefaults fixture

### LINKED ISSUE #872: Fixture editsdefaults is broken
```
@pytest.fixture(scope="function")
def editsdefaults():
    restore = control.config.defaults.copy()
    yield
    control.config.defaults = restore.copy()
```

The fixture replaces the global `config.config.defaults` with a *copy* after the yield.
Other modules that have already imported `control.config.defaults` will still have their original reference.

A proper restore should replace all items in dictionary separately instead.

henklaak:
A trivial PR, fixes #872


bnavigator:
Ok, thanks.

PR #882, opened 2023-04-07
new examples for Simulink-like interconnection of digital and nonlinear systems

sawyerbfuller:
The goal of this PR is to provide examples of how to use `interconnect` to create sophisticated interconnections of systems of different types in a manner that is reminiscent of Simulink. In particular, `simulating_sampled_data_systems.ipynb` provides a few helper functions to accurately simulate interconnections of discrete-time controllers, a nonlinear plant, and time delays. 

Since Python does not have a great numerical integration system for delay differential equations, the approach that is used is to discretize all of the components with a short time interval, including the discrete-time controller. 

A future PR will incorporate some of these helper functions into the library, particularly `sampled_data_controller()`.


doc/examples.rst
```diff
@@ -44,10 +44,12 @@ using running examples in FBS2e.
 
    cruise
    describing_functions
+   interconnect_tutorial
```
murrayrm, line 47: We should probably update the introductory test at the top of this section since the Jupyter notebooks in this section are no longer using running examples in FBS2e (true before this change, but I noticed it now).  How about something like:

> These notebooks demonstrate the use of modeling, anaylsis, and design tools
using examples from textbooks ([FBS2e](https://fbswiki.org/wiki/index.php?title=FBS), [OBC](https://fbswiki.org/wiki/index.php?title=OBC)), courses, and other sources.

(links would have to be done differently, put using GitHub markup here to show where the links go).

sawyerbfuller: sounds good. INcorporated the suggested rewording. 

Also improved the name on one of the examples and switched to lists for specifying set of interconnected systems according to upcoming changes in `interconnect`.

murrayrm:
The new notebook name is _really_ long (56 characters!).  Perhaps worth a bit of discussion about file naming conditions in this folder.  I have been trying to use shorter names (eg, 'mhe-pvtol' and 'stochresp'), but there are a few longer ones (e.g. 'check-controllability-and-observability.py').

Do we care?

sawyerbfuller:
@murrayrm I think that filename was a bit long. Took a stab at shortening it. 

In any case, the title of the notebook is what appears in the readthedocs page, not its filename: 

image

Issue #686, opened 2021-12-31
rootlocus_pid_designer plot looks strange + needs documentation


murrayrm:
From PR #684:

If I use the following code (from the now skipped unit test in `control/tests/rootlocus_test.py`:
```
syscont = ct.TransferFunction(1,[1, 3, 0])
ct.rootlocus_pid_designer(syscont, input_signal='r', Kp0=0.01,
    derivative_in_feedback_path=True)
```
then I get the following plot:

![Sisotool_pid_designer_bug](https://user-images.githubusercontent.com/293362/147444112-acc09729-aa11-43e1-9ef0-aef7a8ab314c.png)

murrayrm:
Also, there is currently no entry for rootlocus_pid_designer in the online documentation (from the docs/ directory).

PR #883, opened 2023-04-14
fix blank bode plot in rootlocus_pid_designer

sawyerbfuller:
Starts with a small gain rather than zero to insure that the frequency response plots appear. Addresses #686

### DIFF
--- a/control/sisotool.py
+++ b/control/sisotool.py
@@ -22,8 +22,8 @@ def sisotool(sys, initial_gain=None, xlim_rlocus=None, ylim_rlocus=None,
              plotstr_rlocus='C0', rlocus_grid=False, omega=None, dB=None,
              Hz=None, deg=None, omega_limits=None, omega_num=None,
              margins_bode=True, tvect=None, kvect=None):
-    """
-    Sisotool style collection of plots inspired by MATLAB's sisotool.
+    """Sisotool style collection of plots inspired by MATLAB's sisotool.
+
     The left two plots contain the bode magnitude and phase diagrams.
     The top right plot is a clickable root locus plot, clicking on the
     root locus will change the gain of the system. The bottom left plot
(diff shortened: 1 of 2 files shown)


murrayrm:
I tried running this on the example I pointed out in #686 and when I click on the root locus, nothing changes.  I then tried the same system in `sisotool` and that also didn't work.  I'm assuming this works for others?  Any clues about what I am doing wrong?

Also, why is 0.001 the right initial_gain to use?  We have a default value for this (`sisotool.initial_gain`) that is set to 1.  Shouldn't we use that instead?

murrayrm:
Also, in looking at the code a bit, I am confused about the fact that the initial gain seems to be set using the `Kp0`, `Ki0`, and `Kd0` parameters, but then it is reset to 0.001 for the gain that you choose to vary.  Should you be able to override this through one of those parameters?

sawyerbfuller:
Regarding default gain: for rootlocus-pid-designer, in principle you are already starting with baseline nonzero kd0, kp0, and ki0, and the step responses show result of that. You are interested in any addition delta-k to one of those, so it makes sense to start that one with a small value. But a zero value erases Bose plot. Np.eps is another option but it would make a very low gain bode plot so I picked a kind of arbitrary small nonzero value.

sawyerbfuller:
Re non-interactive plot: doesn’t work in Jupiter, have to use magic command %matplotlib qt. Assuming you tried that, can you zoom? The way it works you have to click near one of its pre calculated points. It computes more if you zoom in.

sawyerbfuller:
@murrayrm On further thought based on your questions, I realized that a deltaK parameter would be useful needed for non-interactive use, so that the bode plot was usable. Added it in.

murrayrm:
Using %matplotlib fixed the issues I was having, so I added a bit of documentation.

PR #884, opened 2023-04-17
make _convert_to_statespace properly pass signal and system names

sawyerbfuller:
update: this PR has been converted to a bugfix in `_convert_to_statespace` to insure it copies signal and system names correctly. Found the bug by using the test suite. 

Old PR description: 
On some <s>systems</s> computers , the process of converting `tf` systems into `LinearIOSystems` in the `interconnect` function loses signal names -- if the transfer function is static. For example, if a system is defined as `sys=tf(1, 2, inputs='a', outputs='b')`, an error is raised when that system is interconnected with others because `sys` gets copied but loses its signal names, something along the lines of `Signal 'a' not found`.

But this doesn't happen on my system on Python 3.9, 3.10, or 3.11. Maybe something peculiar to windows? 

This PR adds unit tests in an attempt to see if our test suite catches this bug.

### DIFF
--- a/control/statesp.py
+++ b/control/statesp.py
@@ -170,9 +170,9 @@ class StateSpace(LTI):
 
     The StateSpace class is used to represent state-space realizations of
     linear time-invariant (LTI) systems:
-    
+
     .. math::
-    
+
           dx/dt &= A x + B u \\
               y &= C x + D u
 
@@ -1561,7 +1561,8 @@ def _convert_to_statespace(sys):
             return StateSpace(
                 ssout[1][:states, :states], ssout[2][:states, :sys.ninputs],
                 ssout[3][:sys.noutputs, :states], ssout[4], sys.dt,
-                inputs=sys.input_labels, outputs=sys.output_labels)
+                inputs=sys.input_labels, outputs=sys.output_labels,
+                name=sys.name)
         except ImportError:
             # No Slycot.  Scipy tf->ss can't handle MIMO, but static
             # MIMO is an easy special case we can check for here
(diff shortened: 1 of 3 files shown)


sawyerbfuller:
Found it! Losing signal names when slycot is not installed. 

Working on a fix.

sawyerbfuller:
Found the bug and squashed it. `_convert_to_statespace` was not passing signal names correctly in all cases, particularly with no slycot.

murrayrm:
@sawyerbfuller Should the system name be copied over directly or should it be modified by appending `$converted` or something along those lines?  One reason _not_ to change the system name is that if you use `interconnect` with transfer functions, it is more natural to use the same system name.  But this means that you can have two objects (the original TransferFunction and a StateSpace object) both having the same name...

sawyerbfuller:
@murrayrm my first thought is that the private method should keep the same system name because it is being used is internally. 

But maybe if the system was changed by the user using ‘ss’, then that default should be overrided and the system should get a new name. sys$converted?

murrayrm:
I like the idea of having "$converted" appended when you do an explicit conversion (eg, via `ss` applied to a `TransferFunction`).  That way you don't end up with two explicit objects floating around that have the same name but aren't the same object.

If we do that, we should probably generate a utility function/method that renames objects, probably in `namedio.py` so that we get some uniformity (including use of prefix and/or suffix, setting configuration defaults, etc).  That goes beyond this PR, so could be split off as a separate issue?

sawyerbfuller:
HI @murrayrm that sounds good. Yes there are a few issues surrounding when to add `$converted` that may be outside of scope for this PR.  E.g. if you call `LinearIOSystem` on a `StateSpace` system, does that get a suffix? if you call `ss` on a system, what method is responsible for adding `$converted`. 

For this PR, OK if we leave it as it is, with `_convert_to_statespace` not changing the name? This is what you want when you are interconnecting `TransferFunction` and `StateSpace` objects and the former are being converted internally.

PR #885, opened 2023-04-19
update nyquist_plot for DT transfer functions with poles at 0 and 1

sawyerbfuller:


sawyerbfuller:
Was raising an error in these systems.

Issue #294, opened 2019-04-17
Implement d2d in MATLAB compatibility


tomaszG247:
I am not sure if there is an obvious way around this but would be useful to have d2d to resample a discrete time system.

Great job on the tool otherwise

murrayrm:
Great! I've tagged you on this one.  Make sure to take a look at the [python-control developers wiki](https://github.com/python-control/python-control/wiki) for more information on how to contribute.

tobin:
I came here looking for `d2c`. Has there been any progress?

KemmerN:
I would also like to see a d2c function added to the python-control. I have written a version on my own if there is any intrest I would offer looking into adding it.

bnavigator:
Please look into #888 and feel free to submit a pull request of your own idea. We will review it and then can tell if it is interesting.

PR #888, opened 2023-04-27
Discrete to continuous

kletze:
Added a function to convert discrete time LTI systems to continuous time LTI systems.
The function uses the inverse of the zoh-method. A mathematical derivation can be found for example on this page: https://studywolf.wordpress.com/tag/zero-order-hold/

### DIFF
--- a/control/statesp.py
+++ b/control/statesp.py
@@ -164,7 +164,7 @@ def _f2s(f):
 
 
 class StateSpace(LTI):
-    """StateSpace(A, B, C, D[, dt])
+    r"""StateSpace(A, B, C, D[, dt])
 
     A class for representing state-space models.
 
--- a/control/statesp.py
+++ b/control/statesp.py
@@ -164,7 +164,7 @@ def _f2s(f):
 
 
 class StateSpace(LTI):
-    r"""StateSpace(A, B, C, D[, dt])
+    """StateSpace(A, B, C, D[, dt])
 
     A class for representing state-space models.
 
@@ -1988,3 +1988,41 @@ def linfnorm(sys, tol=1e-10):
         fpeak /= sys.dt
 
     return gpeak, fpeak
+
+def discrete2cont(sys, method='zoh'):
+    """ Transform a discrete state-space system to a continuous state-space system.
+    Parameters
+	------
+	sys : Discrete LTI (StateSpace or TransferFunction) system to transform
+	method: The method that was used for discretization
+
+	Returns
+	------
+	sysc : Continuous LIT (StateSpace or TransferFunction) system depending on the input form.
+    """
+    # check if system is of type state space and convert if not
+    #if sys.	
+    
+    if not(isinstance(sys, StateSpace)):
+	    sys = convert_to_statespace(sys)
+	    was_tf = true
+    if method == 'zoh':
+        Ac = scipy.linalg.logm(sys.A)/sys.dt
+        Bc = np.linalg.inv(np.linalg.inv(Ac)@(sys.A-np.eye(np.shape(sys.A)[0])))@sys.B
+		
+        # remove small  numbers
+        Ac = np.around(Ac/1e-6)*1e-6
+        Bc = np.around(Bc/1e-6)*1e-6
+        
+        # turn -0 to 0
+        Ac[np.abs(Ac)<1e-6] = 0
+        Bc[np.abs(Bc)<1e-6] = 0
+		
+        sysc = StateSpace(Ac,Bc,sys.C,sys.D,0)
+    else:
+        raise NotImplementedError("Method: " + method + " is not implemented in this function!")
+
+    if was_tf:
+        sysc = ss2tf(sysc)	
+		
+    return sysc


murrayrm:
I'm not sure this function is really needed.  What's the use case where one wants to convert from continuous time to a (zero-order hold) continuous time model?

If we do include this, it needs to use python-control [condition conventions](https://github.com/python-control/python-control/wiki/Coding-conventions) and include unit tests (in the `control/tests/` directory).

sawyerbfuller:
thanks for the contribution! 

I have not run into a need for this function but it has been suggested before in #294 , and it is available in Matlab https://www.mathworks.com/help/control/ref/dynamicsystem.d2c.html (but not, as far as I can tell, in scipy). 

In addition to adding unit tests, I suggest moving this to being a method of the LTI class in lti.py because it applies equally to both ss and tf systems.  and adding a function that calls that method. Please also see the example of lti.sample_system to see how to properly pass other system parameters like signal and system names.

sawyerbfuller:
I just found that Harold has a cont2discrete method — see `_undescretize` in https://github.com/ilayn/harold/blob/main/harold/_discrete_funcs.py

might have some good ideas.

ilayn:
Possibly my most useless spent week on a nerdy subject 😃 It is certainly possible but like Richard, I don't know what value it adds to anything. I just generalized [this paper](https://doi.org/10.1080/00207170802247728). Not that it makes it more useful but I was neck deep in IQCs so seemed simple enough. Didn't publish it but put it in `harold` instead. You can do arbitrary LFR connections between s and z in case needed.

tobin:
Hi folks, I actually ended up here precisely because I just started using the controls system toolbox and was looking for an equivalent of the MATLAB `d2c` function. I have experimental data in the z-domain, but I like to convert it to the s-domain to think about the continuous time poles and zeroes and to integrate with continuous time models. I'm surprised to see that this is not regarded as useful? Or perhaps I misunderstand the discussion.

murrayrm:
Seems like a valid use case (which is what I was asking about).  A simple implementation is in this PR, but it needs to follow coding conventions, include unit tests, and be documented (in the user manual).

@kletze Are you available to update?  If not, perhaps @tobin can take over this PR?

ilayn:
> I have experimental data in the z-domain, but I like to convert it to the s-domain to think about the continuous time poles and zeroes and to integrate with continuous time models.

Note that, there is no such unique mapping, so you have to make some decision on how to map it back from z to s domain, hence the discussion above. If you have experimental data it is better to get the rest of the models to z domain. 

But I guess fighting this point forever is not useful in particular for the academic use, so probably there is a need for it.

robertobucher:
I have an (old) implementation that I use, but also in this case it is important to modify the code  to follow your conventions.

```
def d2c(sys,method='zoh'):
    """
    Continous to discrete conversion with ZOH method

    Call:
    sysc=d2c(sys,method='log')

    Parameters
    ----------
    sys :   System in statespace or Tf form 
    method: 'zoh' or 'tustin'

    Returns
    -------
    sysc: continous system ss or tf
    

    """
    flag = 0
    if isinstance(sys, ct.TransferFunction):
        sys = ct.tf2ss(sys)
        flag = 1

    a = sys.A
    b = sys.B
    c = sys.C
    d = sys.D
    Ts = sys.dt
    n = np.shape(a)[0]
    nb = np.shape(b)[1]
    nc = np.shape(c)[0]
    tol=1e-12
    
    if method=='zoh':
        if n==1 and a[0,0]==1:
            A = 0
            B = b/sys.dt
            C = c
            D = d
        else:
            tmp1 = np.hstack((a,b))
            tmp2 = np.hstack((np.zeros((nb, n)), np.eye(nb)))
            tmp = np.vstack((tmp1, tmp2))
            s = la.logm(tmp)
            s = s/Ts
            if la.norm(np. imag(s), np.inf) > np.sqrt(sp.finfo(float).eps):
                print('Warning: accuracy may be poor')
            s = np.real(s)
            A = s[0:n,0:n]
            B = s[0:n,n:n+nb]
            C = c
            D = d
    elif method=='foh':
        a = np.asmatrix(a)
        b = np.asmatrix(b)
        c = np.asmatrix(c)
        d = np.asmatrix(d)
        Id = np.asmatrix(np.eye(n))
        A = la.logm(a)/Ts
        A = np.real(np.around(A,12))
        Amat = np.asmatrix(A)
        B = (a-Id)**(-2)*Amat**2*b*Ts
        B = np.real(np.around(B,12))
        Bmat = np.asmatrix(B)
        C = c
        D = d - C*(Amat**(-2)/Ts*(a-Id)-Amat**(-1))*Bmat
        D = np.real(np.around(D,12))
    elif method=='tustin':
        a = np.asmatrix(a)
        b = np.asmatrix(b)
        c = np.asmatrix(c)
        d = np.asmatrix(d)
        poles = la.eigvals(a)
        if any(abs(poles-1)<200*sp.finfo(float).eps):
            print('d2c: some poles very close to one. May get bad results.')
        
        I = np.asmatrix(np.eye(n,n))
        tk = 2 / np.sqrt (Ts)
        A = (2/Ts)*(a-I)*la.inv(a+I)
        iab = la.inv(I+a)*b
        B = tk*iab
        C = tk*(c*la.inv(I+a))
        D = d- (c*iab)
    else:
        print('Method not supported')
        return
    
    sysc=ct.StateSpace(A,B,C,D)
    if flag==1:
        sysc = ct.ss2tf(sysc)
    return sysc
```

robertobucher:
The function discrete_to_continuous is important because of the design of discrete controllers, but it is important to have other methods as only zoh (in particular in education).

Usually we proposed  to our students the following method in order to design a discrete controller:

1.  Convert the continuous plant to a discrete plant
2. Convert the discrete plant to continuous plant using "tustin"
3. Design the controller in continuous time domain with the usual methods (Bode, Nyquist etc.) with the new continuous transfer function
4. Convert the continuous controller to discrete using "tustin"

tobin:
I'm happy to take a crack at it next week.

kletze:
My original use case for this PR was to change the sampling time of models estimated in the discrete-time domain from measured values.
Since Matlab provides the d2d method, I figured it would be kind of handy, to implement this functionality here as well.
A look at the GNU Octave source code shows that within the d2d function, it simply first calls the d2c function and then the c2d function with the corresponding sampling time: https://github.com/gnu-octave/pkg-control/blob/main/inst/%40lti/d2d.m.
However, since the pull request for the “d2c” function didn’t really gain broad acceptance and @ilyan had already implemented the function in
https://github.com/ilayn/harold/blob/main/harold/_discrete_funcs.py, and I was deeply involved in other projects, the PR somewhat fell off my radar.
At the time, it was sufficient for my purposes to implement only the ZOH method.
If the PR is updated here, it definately makes sense to implement the other methods as well, as @robertobucher has shown in his code.
I won’t be available for the next two weeks. After that, I could take care of the matter, and if @tobin joins in before then, all the better.

KybernetikJo:
The bilinear transformation ('tustin') is implemented in slicot, it is part of slycot 0.6. 

- https://en.wikipedia.org/wiki/Bilinear_transform, compare formulas with slicot
- [ab04md, slicot](https://www.slicot.org/objects/software/shared/doc/AB04MD.html)
- [ab04md, slycot](https://github.com/python-control/Slycot/blob/master/slycot/analysis.py#L141)
- [ab04md, slycot, test](https://github.com/python-control/Slycot/blob/master/slycot/tests/test_ab04md.py#L7)

murrayrm:
This PR is more than 3 years old.  Unless there are plans to rebase and revise it, I suggest we close it out.

Issue #690, opened 2022-01-09
Create matlab's equivalent to bandwidth.


Kreijstal:
Is it possible to create matlab's equivalent to bandwidth in control?
https://www.mathworks.com/help/control/ref/lti.bandwidth.html
or are there any alternatives available?

murrayrm:
Definitely possible, although this can also be done with the output from `bode_plot` and the Numpy `argmin` function, I suspect.

If you (or someone) want to contribute code to do this, we can put it either in the main package or in the MATLAB compatibility module.

Kreijstal:
Here is _roughly_ how I solved it for me, I don't have time for providing a pull request,, but maybe someone writing the pull request can get inspired. 

```python
import scipy
import numpy as np
def bandwidth(sys,dbgain=-3):
  result=scipy.optimize.root(lambda s:control.matlab.freqresp(sys,s)[0][0]-control.matlab.dcgain(sys)*10**(dbgain/20),1)
  if result.success:
    return np.abs(result.x)[0]
  else:
    raise Exception(result.message)
```

roryyorke:
Looks like merged #889 closes this?

Kreijstal:
I have not used control in years, but I am happy it got solved

PR #889, opened 2023-05-01
bandwidth feature 

SCLiao47:
As discussed in #690, the bandwidth feature is added to the LTI class.

Implementation:
1. The range of possible frequency is extracted from `freqplot._default_frequency_range`. 
2. The index, $idx$, is identified, which corresponds to the first frequency dropped by a desired amount relative to the DC gain.
3. Bisection search (`scipy.optimize.root_scalar`) is used to solve for the bandwidth in the frequency bracket $[\omega[idx-1], \omega[idx]]$.

Notes:
1. The bandwidth of a system with infinite DC gain is set to be `np.nan`, which follows the convention of Matlab.
2. The above approach is chosen to solve the bandwidth instead of the `scipy.optimize.root` suggested in issue #690. The suggested approach requires an initial guess and might fail, e.g., a system with a resonant peak at a frequency higher than the bandwidth. 

Notes on git commits:
git commits are duplicated as I was having some issues to push. I pulled (suggested by git error hint), and it prompts `"merge made by the 'recursive' strategy.` Not sure whether this is a problem.

### DIFF
--- a/control/xferfcn.py
+++ b/control/xferfcn.py
@@ -74,6 +74,7 @@ _xferfcn_defaults = {
     'xferfcn.floating_point_format': '.4g'
 }
 
+
 def _float2str(value):
     _num_format = config.defaults.get('xferfcn.floating_point_format', ':.4g')
     return f"{value:{_num_format}}"
@@ -1246,7 +1247,7 @@ class TransferFunction(LTI):
 
         """
         return self._dcgain(warn_infinite)
-    
+
     def _isstatic(self):
         """returns True if and only if all of the numerator and denominator
         polynomials of the (possibly MIMO) transfer function are zeroth order,
@@ -1406,6 +1407,7 @@ def _tf_factorized_polynomial_to_string(roots, gain=1, var='s'):
         factors = [f"({factor})" for factor in factors]
 
     return multiplier + " ".join(factors)
+
 
 def _tf_string_to_latex(thestr, var='s'):
     """ make sure to superscript all digits in a polynomial string
--- a/control/xferfcn.py
+++ b/control/xferfcn.py
@@ -1246,6 +1246,10 @@ class TransferFunction(LTI):
 
         """
         return self._dcgain(warn_infinite)
+    
+    def bandwidth(self, dbdrop=-3):
+        """Return the bandwith"""
+        return self._bandwidth(dbdrop)
 
     def _isstatic(self):
         """returns True if and only if all of the numerator and denominator
--- a/control/xferfcn.py
+++ b/control/xferfcn.py
@@ -1247,10 +1247,6 @@ class TransferFunction(LTI):
         """
         return self._dcgain(warn_infinite)
     
-    def bandwidth(self, dbdrop=-3):
-        """Return the bandwith"""
-        return self._bandwidth(dbdrop)
-
     def _isstatic(self):
         """returns True if and only if all of the numerator and denominator
         polynomials of the (possibly MIMO) transfer function are zeroth order,
--- a/control/lti.py
+++ b/control/lti.py
@@ -203,11 +203,6 @@ class LTI(NamedIOSystem):
             return zeroresp
 
     def bandwidth(self, dbdrop=-3):
-        """Return the bandwidth"""
-        raise NotImplementedError("bandwidth not implemented for %s objects" %
-                                  str(self.__class__))
-
-    def _bandwidth(self, dbdrop=-3):
         # check if system is SISO and dbdrop is a negative scalar
         if not self.issiso():
             raise TypeError("system should be a SISO system")
--- a/control/lti.py
+++ b/control/lti.py
@@ -209,8 +209,11 @@ class LTI(NamedIOSystem):
 
     def _bandwidth(self, dbdrop=-3):
         # check if system is SISO and dbdrop is a negative scalar
-        if (not self.issiso()) and (dbdrop >= 0):
-            raise ValueError("#TODO ")
+        if not self.issiso():
+            raise TypeError("system should be a SISO system")
+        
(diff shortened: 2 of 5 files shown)


murrayrm:
Thanks for this contribution, @SCLiao47.

I think the `_bandwidth` method should be renamed to `bandwidth` and put in the `LTI` class in `lti.py`.  With that change, you should be able to get rid of the `bandwidth` methods in `TransferFunction` and `StateSpace`, since they will get the method from their parent class.  You will probably have to override `bandwidth` in `FrequencyResponseData` since the `dcgain` method is not implemented there.

I have some other suggestions, but if you agree with the change above then I would start there, since it will move things around.

SCLiao47:
@murrayrm thank you for the comments and suggestions.

The `_bandwidth` method was implemented as I was following the `dcgain` method in the `LTI `class in `lti.py`. But as you suggested renaming it to `bandwidth` can simplify the code structure. One other thing I wasn't so sure about is "what other classes should have the `bandwidth` method." The `FrequencyResponseData` class is one of them as you mentioned, and I wonder if there are others as well?

[Regarding development flow]
- I'm still learning the development flow on GitHub. The [contribute wiki page](https://github.com/python-control/python-control/wiki/How-to-contribute-with-a-pull-request) (step 11) mentioned that I could keep working on the branch under my own fork. However, does the new commit directly sync to this pull request? 
- One other thing I notice while following the contribute wiki page. It seems like GitHub has renamed the `master` branch to `main`. Maybe there needs some edition to the wiki page but I'm not so sure.

sawyerbfuller:
Hi @SCLiao47  

i think only the classes that derive from LTI should get a bandwidth method: tf, ss, and frd. 

As for your question about development: adding new commits to your branch is very seamless. Add them to your own branch (on your own computer) and then, when you do git push origin, they will appear right here on your pull request. 

Our main branch name is grandfathered in as master.

murrayrm:
I've updated the documentation to refer to the main branch (master is deprecated).

SCLiao47:
@sawyerbfuller @murrayrm thank you for the suggestions and comments. I've restructured the `bandwidth `method in the `LTI` class in `lti.py`.

# FRD
As for the `FrequencyResponseData` class, I wonder how you would suggest implementing the `bandwidth` method for it? At the moment, the class seems to not support interpolating, which is needed to infer DC gain and compute the bandwidth. However, this will probably fall outside of the scope of this feature. Is there a way to get around this?

# Development flow
I think I need to make sure I understand the flow and might need some clarification. In step 8 of the [contribute wiki page](https://github.com/python-control/python-control/wiki/How-to-contribute-with-a-pull-request), the intention is to update and sync the _local code_ (on my machine) to the _upstream version_ (python-control/python-control). 

What happened in the instructions:
1. After clicking [`sync fork` and `Update branch`](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork) on github.com, my _forked repo_ (SCLiao47/python-control) will be sync to the _upstream version_. Both repos are remote on github. 
2. The first chunk of command lines (on my machine) is to sync my _local main branch_ to the _main_ of the _forked repo_ on GitHub. 
3. The second chunk of command lines is to bring the commits in the _local feature branch_ to the _local main_. The `git rebase` allows resolving conflict before merging _local feature_ into _upstream main_ and maintaining a cleaner git graph. 

If the above understanding is correct, I wonder about the following: 

1. The _local main_ is not intended to be developed and made changed from my understanding. Why do we need to `git push` after `merge origin/main`?
2. Can we skip `sync fork` on github.com and replace `git merge origin/main` with `git merge upstream/main`? This will have the same result that the _local main_ will be synced to _upstream main_, right?

control/xferfcn.py
```diff
@@ -1246,7 +1246,7 @@ def dcgain(self, warn_infinite=False):
 
         """
         return self._dcgain(warn_infinite)
-
+
```
murrayrm, line 1249: These excess space should be removed.

SCLiao47: Updated.

bnavigator:
> What happened in the instructions:
> 
> 1. After clicking [`sync fork` and `Update branch`](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork) on github.com, my _forked repo_ (SCLiao47/python-control) will be sync to the _upstream version_. Both repos are remote on github.

Correct.

> 2. The first chunk of command lines (on my machine) is to sync my _local main branch_ to the _main_ of the _forked repo_ on GitHub.

Correct.

> 3. The second chunk of command lines is to bring the commits in the _local feature branch_ to the _local main_. The `git rebase` allows resolving conflict before merging _local feature_ into _upstream main_ and maintaining a cleaner git graph.

Yes, the commands make sure that your local feature branch also has all the recent changes of your recently updated local `main` which should now be at the same commit as `upstream/main`
 
> If the above understanding is correct, I wonder about the following:
> 
> 1. The _local main_ is not intended to be developed and made changed from my understanding. Why do we need to `git push` after `merge origin/main`?

If you didn't change anything in your local main before the merge, you don't need to push. The push is a noop then. In that case the merge will be a fast-forward and it is sufficient to alternatively just.

```shell
git switch main; git pull
````

But if you have changes in local main (which you should not), you will notice here.

> 2. Can we skip `sync fork` on github.com and replace `git merge origin/main` with `git merge upstream/main`? This will have the same result that the _local main_ will be synced to _upstream main_, right?

Yes, but then you need the push so that your `origin/main` does not stay behind. These sequences are equivalent, assuming `origin/main` is already the tracked remote branch of your local clone.

1. 
    ```
    git switch main; git fetch --all; git merge upstream/main; git push
    ```
2. on main press "sync fork", then:
    ```
   git switch main; git pull
   ```

PR #892, opened 2023-05-20
Add missing labels when returning TimeResponseData

### LINKED ISSUE #891: Missing labels from forced_response output
Hi,

I wrote a state-space system representation with labels, like:

``` Python
sys = ct.ss(
  A, B, C, D,
  name="motor",
  states=("Ia", "Wm"),
  inputs=("Va", "Tl"),
  outputs=("Ia", "Wm", "Va", "Tl"),
)
```

But after simulating it, the result (a `TimeResponseData`), lacks information about the labels of the simulated system:
``` Python
T = np.linspace(0, 0.360, 360)
Va = np.ones_like(T) * 36
Tl = np.linspace(0, 15, len(T))

res = ct.forced_response(motor, T, U=[Va, Tl], X0=[0, 0])
print(res.output_labels)  # this is None
print(res.input_labels)  # this is None
print(res.state_labels)  # this is None
```

I wanted to use `.to_pandas()`, and after assigning it manually, it works, so it's probably just a matter of passing it forward internally in the code:
``` Python
res.output_labels = motor.output_labels
res.input_labels = motor.input_labels
res.state_labels = motor.state_labels
df = res.to_pandas().set_index("time")
```

For completeness, this is the error if I try to use `.to_pandass()` without the abovementioned workaround:

``` Python
    648 # Create a dict for setting up the data frame
    649 data = {'time': self.time}
    650 data.update(
--> 651     {name: self.u[i] for i, name in enumerate(self.input_labels)})
    652 data.update(
    653     {name: self.y[i] for i, name in enumerate(self.output_labels)})
    654 data.update(
    655     {name: self.x[i] for i, name in enumerate(self.state_labels)})

TypeError: 'NoneType' object is not iterable
```

Thanks

joaoantoniocardoso:
Fixes #891

### DIFF
--- a/control/timeresp.py
+++ b/control/timeresp.py
@@ -694,7 +694,10 @@ def _process_labels(labels, signal, length):
             raise ValueError("Name dictionary for %s is incomplete" % signal)
 
     # Convert labels to a list
-    labels = list(labels)
+    if isinstance(labels, str):
+        labels = [labels]
+    else:
+        labels = list(labels)
 
     # Make sure the signal list is the right length and type
     if len(labels) != length:
--- a/control/timeresp.py
+++ b/control/timeresp.py
@@ -1111,6 +1111,8 @@ def forced_response(sys, T=None, U=0., X0=0., transpose=False,
 
     return TimeResponseData(
         tout, yout, xout, U, issiso=sys.issiso(),
+        output_labels=sys.output_labels, input_labels=sys.input_labels,
+        state_labels=sys.state_labels,
         transpose=transpose, return_x=return_x, squeeze=squeeze)
 
 
@@ -1374,8 +1376,16 @@ def step_response(sys, T=None, X0=0., input=None, output=None, T_num=None,
     # Figure out if the system is SISO or not
     issiso = sys.issiso() or (input is not None and output is not None)
 
+    # Select only the given input and output, if any
+    input_labels = sys.input_labels if input is None \
+        else sys.input_labels[input]
+    output_labels = sys.output_labels if output is None \
+        else sys.output_labels[output]
+
     return TimeResponseData(
         response.time, yout, xout, uout, issiso=issiso,
+        output_labels=output_labels, input_labels=input_labels,
+        state_labels=sys.state_labels,
         transpose=transpose, return_x=return_x, squeeze=squeeze)
 
 
@@ -1704,9 +1714,15 @@ def initial_response(sys, T=None, X0=0., input=0, output=None, T_num=None,
     # Figure out if the system is SISO or not
     issiso = sys.issiso() or (input is not None and output is not None)
 
+    # Select only the given output, if any
+    output_labels = sys.output_labels if output is None \
+        else sys.output_labels[0]
+
     # Store the response without an input
     return TimeResponseData(
         response.t, response.y, response.x, None, issiso=issiso,
+        output_labels=output_labels, input_labels=None,
+        state_labels=sys.state_labels,
         transpose=transpose, return_x=return_x, squeeze=squeeze)
 
 
@@ -1798,7 +1814,7 @@ def impulse_response(sys, T=None, X0=0., input=None, output=None, T_num=None,
     -----
     This function uses the `forced_response` function to compute the time
     response. For continuous time systems, the initial condition is altered to
-    account for the initial impulse. For discrete-time aystems, the impulse is 
+    account for the initial impulse. For discrete-time aystems, the impulse is
     sized so that it has unit area.
 
     Examples
@@ -1869,8 +1885,16 @@ def impulse_response(sys, T=None, X0=0., input=None, output=None, T_num=None,
     # Figure out if the system is SISO or not
     issiso = sys.issiso() or (input is not None and output is not None)
 
+    # Select only the given input and output, if any
+    input_labels = sys.input_labels if input is None \
+        else sys.input_labels[input]
+    output_labels = sys.output_labels if output is None \
+        else sys.output_labels[output]
+
     return TimeResponseData(
         response.time, yout, xout, uout, issiso=issiso,
+        output_labels=output_labels, input_labels=input_labels,
+        state_labels=sys.state_labels,
         transpose=transpose, return_x=return_x, squeeze=squeeze)
 
 
--- a/control/iosys.py
+++ b/control/iosys.py
@@ -1862,7 +1862,7 @@ def input_output_response(
 
         return TimeResponseData(
             t_eval, y, None, u, issiso=sys.issiso(),
-            output_labels=sys.output_index, input_labels=sys.input_index,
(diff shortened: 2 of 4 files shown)


control/timeresp.py
```diff
@@ -1111,6 +1111,8 @@ def forced_response(sys, T=None, U=0., X0=0., transpose=False,
 
     return TimeResponseData(
         tout, yout, xout, U, issiso=sys.issiso(),
+        output_labels=sys.output_index, input_labels=sys.input_index,
+        state_labels=sys.state_index,
```
murrayrm, line 1115: This is OK, but it might be more clear to use `sys.output_labels`, etc.

control/timeresp.py
```diff
@@ -1374,8 +1376,14 @@ def step_response(sys, T=None, X0=0., input=None, output=None, T_num=None,
     # Figure out if the system is SISO or not
     issiso = sys.issiso() or (input is not None and output is not None)
 
+    # Select only the given input and output, if any
+    input_index = sys.input_index if input is None else {k: 0 for k, v in sys.input_index.items() if v == input}
+    output_index = sys.output_index if output is None else {k: 0 for k, v in sys.output_index.items() if v == output}
+
```
murrayrm, line 1382: Consider changing this to 
```
    input_labels = sys.input_labels if input is None \
        else sys.input_labels[input]
    output_labels = sys.output_labels if output is None else \
        sys.output_labels[output]
```
which seems cleaner.

control/timeresp.py
```diff
@@ -1374,8 +1376,14 @@ def step_response(sys, T=None, X0=0., input=None, output=None, T_num=None,
...
     return TimeResponseData(
         response.time, yout, xout, uout, issiso=issiso,
+        output_labels=output_index, input_labels=input_index,
+        state_labels=sys.state_index,
```
murrayrm, line 1386:
```
        output_labels=output_labels, input_labels=input_labels,
        state_labels=sys.state_labels,
```

control/timeresp.py
```diff
@@ -1704,9 +1712,14 @@ def initial_response(sys, T=None, X0=0., input=0, output=None, T_num=None,
     # Figure out if the system is SISO or not
     issiso = sys.issiso() or (input is not None and output is not None)
 
+    # Select only the given output, if any
+    output_index = sys.output_index if output is None else {k: 0 for k, v in sys.output_index.items() if v == output}
+
```
murrayrm, line 1717: As above.

control/timeresp.py
```diff
@@ -1704,9 +1712,14 @@ def initial_response(sys, T=None, X0=0., input=0, output=None, T_num=None,
...
     # Store the response without an input
     return TimeResponseData(
         response.t, response.y, response.x, None, issiso=issiso,
+        output_labels=output_index, input_labels=None,
+        state_labels=sys.state_index,
```
murrayrm, line 1722: As above.

control/timeresp.py
```diff
@@ -1869,8 +1882,14 @@ def impulse_response(sys, T=None, X0=0., input=None, output=None, T_num=None,
     # Figure out if the system is SISO or not
     issiso = sys.issiso() or (input is not None and output is not None)
 
+    # Select only the given input and output, if any
+    input_index = sys.input_index if input is None else {k: 0 for k, v in sys.input_index.items() if v == input}
+    output_index = sys.output_index if output is None else {k: 0 for k, v in sys.output_index.items() if v == output}
+
```
murrayrm, line 1888: As above.

control/timeresp.py
```diff
@@ -1869,8 +1882,14 @@ def impulse_response(sys, T=None, X0=0., input=None, output=None, T_num=None,
...
     return TimeResponseData(
         response.time, yout, xout, uout, issiso=issiso,
+        output_labels=output_index, input_labels=input_index,
+        state_labels=sys.state_index,
```
murrayrm, line 1892: As above.

murrayrm (commented):
Thanks for this nice contribution.  I suggest using `*_labels` instead of `*_index`, which I think makes the code a bit more readable.

joaoantoniocardoso:
Hi @murrayrm, I did the changes regarding using `_labels` instead of `_labels`, can you take a look again? Thanks!

control/timeresp.py
```diff
@@ -1869,8 +1885,16 @@ def impulse_response(sys, T=None, X0=0., input=None, output=None, T_num=None,
     # Figure out if the system is SISO or not
     issiso = sys.issiso() or (input is not None and output is not None)
 
+    # Select only the given input and output, if any
+    input_labels = sys.input_labels if input is None \
+        else sys.input_labels[:1]
+    output_labels = sys.output_labels if output is None \
+        else sys.output_labels[:1]
```
joaoantoniocardoso, line 1892: Now I'm thinking if it should be
```suggestion
        else sys.output_labels[output]
```

... I'll add a test case to check it

control/timeresp.py
```diff
@@ -1358,6 +1358,12 @@ def step_response(sys, T=None, X0=0., input=None, output=None, T_num=None,
     xout = np.empty((sys.nstates, ninputs, np.asarray(T).size))
     uout = np.empty((ninputs, ninputs, np.asarray(T).size))
 
+    # Get labels for the inputs/outputs (prior to mimo -> simo conversion)
+    input_labels = sys.input_labels if input is None \
+        else [sys.input_labels[input]]
+    output_labels = sys.output_labels if output is None \
+        else [sys.output_labels[output]]
+
```
joaoantoniocardoso, line 1366: I was thinking about it. Should we move it to here, or should we make the mimo/siso/simo conversions to also propagate the labels (and name)?

murrayrm: Either way works, but because `_get_ss_simo` is an internal function, it is fine for it to ignore labels.

joaoantoniocardoso: take a look at [this commit](https://github.com/python-control/python-control/commit/a711f5530eec8292729b565e333f4372887b9686)

murrayrm: It's your PR, so feel free to revert the moves that I made and put in your proposed commit.  I was just trying to get something that passed the unit tests, since they were failing.

murrayrm: Also: I think it should work to do `sys.output_labels[output]` instead of `[sys.output_labels[output]]` (eg, remove the outer list).  The code underneath is smart enough to deal with getting passed a string for a single input/output.

joaoantoniocardoso: > Also: I think it should work to do `sys.output_labels[output]` instead of `[sys.output_labels[output]]` (eg, remove the outer list). The code underneath is smart enough to deal with getting passed a string for a single input/output.

The _process_labels had a small bug, now it works with plain str instead of [str]

joaoantoniocardoso:
@murrayrm I'm good with the patch, let me know if you need any changes, otherwise feel free to merge it

Issue #901, opened 2023-06-03
Add rename system method to NamedIOSystems


murrayrm:
As discussed briefly in #884, it would be useful to have a generic method for renaming systems that are converted from one form to another (eg, time discretization, linearization, tf2ss, etc).  This is currently done with individual code that allows for prefix/suffix changes.

PR #903, opened 2023-06-04
Change name of converted LinearIOSystems

murrayrm:
This PR implements that change suggested in PR #884 and issue #901 regarding changing the system name when a transfer function is converted to a state space system and vice versa.  Converted systems now have "$converted" appended to the system name.  This is only done when you do an explicit conversation via `ss` or `tf`.  Implicit conversions from a `TransferFunction` to a `StateSpace` object, for example as part of using `interconnect` use the original name.

In addition, prefix/suffix processing has been moved into the `NamedIOSystem` class, under `_copy_names` and `_name_or_default` methods.

### DIFF
--- a/control/iosys.py
+++ b/control/iosys.py
@@ -591,13 +591,8 @@ class InputOutputSystem(NamedIOSystem):
                 DeprecationWarning)
 
         if copy_names:
-            linsys._copy_names(self)
-            if name is None:
-                linsys.name = \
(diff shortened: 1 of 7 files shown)


sawyerbfuller (approved):
The new method, `_copy_names` is something I have needed on a few different occasions and will put to use. This looks like a good solution.

Issue #893, opened 2023-05-22
Simulation Function Names Poorly Chosen...


frohro:
Hi,

Generally, the forced response is also known as the zero state response and is the response due to the forcing function (the input) alone (zero initial conditions).  The control.forced_response() is a linear simulation for a linear system, and it can give the response due to the input and the initial condition all at once, not just the forced response.  You just put in the input and initial condition.  It is similar to lsim in Matlab. 

This is made more confusing because you have a control.initial_response(), where the input parameter is actually there, but the docs say it is ignored!  This is also known as the zero input response, or natural response. 

The control.input_output_response() is a nonlinear simulation of the system, and could use a name that says that.  In [this location in the docs](https://python-control.readthedocs.io/en/0.9.3.post2/generated/control.input_output_response.html#control.input_output_response), it has no mention of it being a nonlinear simulation.

There are probably other similar issues.  These are just the ones that caused me grief in the last couple of weeks.

This may be a hard thing to change, given the president already set, but you could depreciate the present function names, and make the option of using some more descriptive ones. Eventually, everyone will use the more descriptive function calls and you can remove the poorly named ones.

Thanks for this helpful package!

Rob

Issue #890, opened 2023-05-05
Using input_output_response(...), Simple System Gives Erroneous Output


frohro:
This system should not have the bump in the output you see about halfway through the simulation.
```
import matplotlib.pyplot as plt
import numpy as np
import control as ctrl

R = 22e3
C = 470e-6

circuit = ctrl.ss(-1/(R*C), 1/(R*C), 1, 0)
t, resp = ctrl.step_response(circuit)
plt.plot(t, resp)
plt.title('Step Response of the RC Circuit')
plt.grid()
print("The eigenvalue is: ",-1/(R*C))
print("numpy gives the eigenvalue, and eigen vector respectively as: ")
np.linalg.eig(circuit.A)
t = np.arange(0, 120, 0.01)
u = 3.3*148/256-3.3*40/256*np.heaviside(t-60,1)
plt.plot(t,u)
x0 = 3.3*128/256
t, vc = ctrl.input_output_response(circuit, t, u, x0)
plt.plot(t,vc)
plt.grid()
plt.title("Response of RC Circuit")
plt.legend(["u","$v_c$"])
plt.xlabel("time (seconds)")
plt.ylabel("Voltage (Volts)")
plt.show()
```

![image](https://user-images.githubusercontent.com/994565/236429698-d4990dcf-9371-4ef4-9958-47ed940c6ccb.png)

murrayrm:
Confirmed this behavior.  Using `forced_response` instead of `input_output_response` works correctly:
![Figure_1](https://github.com/python-control/python-control/assets/293362/eb4156fc-fcb4-49e5-beab-b3f48656de77)
So something odd going on in `input_output_response`...

frohro:
It appears to me that the input_output_response is using Runga-Kutta 45, and that the tolerance is set to 0.001 or so.  I get the same results with that if the rtol is set to low.   Here is my python script with the RK45 method.  Here is the plot.  
![image](https://github.com/python-control/python-control/assets/994565/9549744a-c01f-4614-93df-06bae5ae1236)
`# %% Imports
import numpy as np
import matplotlib.pyplot as plt
from scipy.integrate import solve_ivp

# %% Define independent function and derivative function
def x(t):
    return 1-np.heaviside(t-60, 1)

def f(t, y, c):
    R = 22e3
    C = 470e-6
    dydt = x(t)/R/C - y/R/C
    return dydt

# %% Define time spans, initial values, and constants
tspan = np.linspace(0, 120, 1000)
[...]

Issue #904, opened 2023-06-04
Releasing 0.9.4 and moving to 0.10.0


murrayrm:
I am planning to do a new release (0.9.4) in the coming days that will include all recent PRs _except_ #881, since that may have some incompatibilities due to the requirement to use lists (versus tuples) in some parts of the `interconnect` interface.

Are there any additional changes we want for 0.9.4?

For version 0.10.0, I was thinking that we might:

* Add in the new multivariable interconnect functionality
* Get rid of support for the NumPy `matrix` class (no longer recommended by NumPy, though not yet deprecated)
* Consider other changes that would require small rewrites of working user code.

Use this issue to post thoughts and suggestions.

sawyerbfuller:
I think I have everything that I hoped to for a .9.4 release already submitted.

As for .10, I am in favor of using that release to begin phasing out deprecations, most notably the matrix class. 

Another possible future change to consider for that time frame is merging `StateSpace` and `LinearIOSystem` classes. Are there any reasons to keep them separate aside from the effort to refactor the code? There is also the matter of what the resulting class should be called. 

What's needed for a 1.0 release? Here are two possible features that come to mind that may be worth considering:
1. Time delays
2. Refactoring the code to separate plotting from computation enough to enable a plot-free version of python control as suggested in #808, which would be helped by #645

sawyerbfuller:
As long as we are talking about architecture, I had a question about the use of the dictionaries `input_index` and `output_index`. Is there any chance we could get away with a simplification in which signal names were stored entirely in `input_labels` and `output_labels`, which are lists of signal names? The location in the list could encode the index of the signal. To find the index associated with a name/label, you would use the syntax `sys.input_labels.index('u')` rather than `sys.input_index['u']` -- a few more characters, but fewer fields to keep around and keep in sync. But I really don't know all the ways these are used -- especially as we move toward multivariable interconnect -- so I may be missing something.

murrayrm:
There are not actually two separate lists.  Input `input_labels` (and the others) is a _property_ that just returns `input_labels.keys()`.  So no syncing issues (and you can use `sys.input_labels.index('u')` if you like -:).

murrayrm:
Regarding system classes, here is what we have now:

![classes-0 9 3 post2](https://github.com/python-control/python-control/assets/293362/35171b41-ed44-46ef-ba81-881834e5b7bc)

I think we can do the following:
[...]

PR #905, opened 2023-06-05
add/cleanup documentation on simulation functions

murrayrm:
This PR updates the user guide and docstrings to provide more information on simulation functions, addressing issues #890 and #893.

I also added the missing `OptimalEstimationProblem` and `OptimalEstimationResult` classes to the documentation.

### DIFF
--- a/control/__init__.py
+++ b/control/__init__.py
@@ -54,10 +54,15 @@ The docstring examples assume that the following import commands::
 
 Available subpackages
 ---------------------
-flatsys
-    Differentially flat systems
-optimal
-    Optimization-based control
+
+The main control package includes the most commpon functions used in
+analysis, design, and simulation of feedback control systems.  Several
+additional subpackages are available that provide more specialized
+functionality:
+
+* :mod:`~control.flatsys`: Differentially flat systems
+* :mod:`~control.matlab`: MATLAB compatibility module
+* :mod:`~control.optimal`: Optimization-based control
 
 """
 
--- a/control/flatsys/systraj.py
+++ b/control/flatsys/systraj.py
@@ -40,7 +40,7 @@ import numpy as np
 from ..timeresp import TimeResponseData
 
 class SystemTrajectory:
-    """Class representing a system trajectory.
+    """Class representing a trajectory for a flat system.
 
     The `SystemTrajectory` class is used to represent the
     trajectory of a (differentially flat) system.  Used by the
--- a/control/iosys.py
+++ b/control/iosys.py
@@ -1704,6 +1704,15 @@ def input_output_response(
        start with zero initial condition since this can be specified as
        [xsys_0, 0].  A warning is issued if the initial conditions are padded
        and and the final listed initial state is not zero.
+    
+    2. If discontinuous inputs are given, the underlying SciPy numerical
+       integration algorithms can sometimes produce erroneous results due
+       to the default tolerances that are used.  The `ivp_method` and
+       `ivp_keywords` parameters can be used to tune the ODE solver and
+       produce better results.  In particular, using 'LSODA' as the
+       `ivp_method` or setting the `rtol` parameter to a smaller value
+       (e.g. using `ivp_kwargs={'rtol': 1e-4}`) can provide more accurate
+       results.
 
     """
     #
--- a/control/optimal.py
+++ b/control/optimal.py
@@ -3,7 +3,7 @@
 # RMM, 11 Feb 2021
 #
 
-"""The :mod:`~control.optimal` module provides support for optimization-based
+"""The :mod:`control.optimal` module provides support for optimization-based
 controllers for nonlinear systems with state and input constraints.
 
 The docstring examples assume that the following import commands::
--- a/control/timeresp.py
+++ b/control/timeresp.py
@@ -820,7 +820,7 @@ def _check_convert_array(in_obj, legal_shapes, err_msg_start, squeeze=False,
 # Forced response of a linear system
 def forced_response(sys, T=None, U=0., X0=0., transpose=False,
                     interpolate=False, return_x=None, squeeze=None):
-    """Simulate the output of a linear system.
+    """Compute the output of a linear system given the input.
 
     As a convenience for parameters `U`, `X0`:
     Numbers (scalars) are converted to constant arrays with the correct shape.
@@ -1616,7 +1616,7 @@ def step_info(sysdata, T=None, T_num=None, yfinal=None,
 def initial_response(sys, T=None, X0=0., input=0, output=None, T_num=None,
                      transpose=False, return_x=False, squeeze=None):
     # pylint: disable=W0622
-    """Initial condition response of a linear system
+    """Compute the initial condition response for a linear system.
 
     If the system has multiple outputs (MIMO), optionally, one output
     may be selected. If no selection is made for the output, all
--- a/doc/classes.rst
+++ b/doc/classes.rst
@@ -58,3 +58,8 @@ Additional classes
    flatsys.SystemTrajectory
    optimal.OptimalControlProblem
    optimal.OptimalControlResult
+   optimal.OptimalEstimationProblem
+   optimal.OptimalEstimationResult
+
+The use of these classes is described in more detail in the
+:ref:`flatsys-module` module and the :ref:`optimal-module` module
--- a/doc/conventions.rst
+++ b/doc/conventions.rst
@@ -11,7 +11,6 @@ way that different types of standard information used by the library.
 Throughout this manual, we assume the `control` package has been
 imported as `ct`.
 
-
 LTI system representation
 =========================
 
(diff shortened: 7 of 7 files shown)


murrayrm:
It should be possible to make all of the systems work for a NonlinearIOSystem.

For the impulse response, we could take the linearization of the system at the initial condition and then use that to reset the initial condition, though I'd want to think that through make sure that is the right thing to do mathematically.

sawyerbfuller (approved):
Happy to add nonlinear system compatibility to time simulation functions to a future PR.  LGTM

frohro:
@sawyerbfuller,
Here are two comments for your questions.  
The terms zero state (forced response) and zero input response (initial condition response) are really only meaningful in a linear system, because, in a linear system, they may be added to get the full response.  If the system is nonlinear, then they make no sense.  So my advice is do not do anything with them in the non-linear domain.  
Likewise, there is no reason to handle an impulse response in a nonlinear way, because its importance comes through superposition, where you can use the impulse response to obtain the total response (the convolution integral).  So I don't think you should think about computing it for a nonlinear system.  You will just be leading folk astray in their understanding of linear systems.  
Rob

sawyerbfuller:
hi @frohro - 

These four simulation functions derive their names from the matlab equivalents, `initial`, `impulse`, `step`, and `lsim`, updated somewhat to have more pythonic or descriptive names, especially for `lsim`. My sense is that their meaning is widely understood to be what you describe - no input or zero state respectively. Is your sense that they are mis-named in Matlab as well? I don't think I have encountered the precise terminology "zero state" and "zero input", but I see what they mean. 

At the very least, there should probably be a warning in `initial` if the user supplies an input that is ignored. 

As for nonlinear simulation, it seems like it would be useful to have functionality available in the library to do the same for nonlinear systems, even if superposition doesn't hold for them. An example that comes to mind is simulating a nonlinear pendulum system from an arbitrary initial state. 

Granted, complexities may arise in the extensions of such functions to nonlinear systems because in many cases we will not be able to auto-compute a reasonable time window, but I will assume they can be handled somehow.

frohro:
Hi @sawyerbfuller,
My beef is with the naming of these simulation functions.  It would be nice if it was as you claim, but it does **not** match MATLAB. 
 I'm sorry.  I must not have made my point of view clear.  Let me try again:

1. control.forced_response() should be named in a way that it is clear it is a general linear simulation, like MATLAB, which calls it lsim().  The term "forced response" of a linear system is also known as the zero-state response, which means X0 = 0.0, so the initial condition option for the user should not be an option in a function called forced_response!  I am unaware of a MATLAB simulation that computes only the forced or zero-state response.  Maybe we should call it lsim().
2. control.input_output_response() already computes the response in a way that works for linear and nonlinear systems.  Its name is not quite so poorly chosen, though I would prefer the name to emphasize that it works for both linear and nonlinear systems.  In my opinion, we should leave the linear simulation alone, and not add the nonlinear feature to it, because we have a nonlinear one already.  It should perhaps be more clear that it is a more general simulation.
3. control.initial_response() has a parameter, input, which the docs claim is for compatibility with the other simulations, but it is ignored.  From my perspective, the confusion of having "input" as a parameter that is ignored is worse than any benefit coming from having the same arguments as other simulations.  MATLAB certainly does not do anything like this in its initial function.  They have no problem having functions with different arguments if it cuts the confusion.  When a casual reader sees a parameter in a function, they will assume it is used in the function.  Why would you want two functions to have the same arguments if they perform different tasks, especially when the parameter is ignored in one?  Every function would be harder to use, understand, etc.!  It just makes the user think you have to take into account an input, even if it is zero in a linear system, and that is false.
I use the Python control package as a teaching tool.  I used to use MATLAB/octave, but I'm switching to Python, and I don't want the control package to give my students the wrong ideas.  :-)

I understand that changing function names and parameters are painful, but Python is famous for changing those things, and it can probably be done in a less painful way if careful thought is put into it.  Hopefully, it is more clear what bothers me about these things.  Thanks for thinking about this.

Rob

ilayn:
Dealing with matlab legacy almost daily on scipy, I think lsim is the worst possible name that they can choose when they were dealing with 8.3 fortran name silliness. 

forced response is exactly what is happening. Zero state response has nothing to do with it since initial condition is an impulse response super imposed to the input function. So I don't quite get the argument there. Initial condition is a forcing if you let a mass spring damper system from a nonzero condition. It only hides the fact that somebody/something pumped energy to the system prior to t=0. So in that sense I disagree that initial conditions are independent from the forcing signal that is u. I think this should have been the case for matlab but they did poor naming themselves decades ago. 

linear or not input output response tells you that there is this input and and there will be an output. I am not a big fan but any other complexity will make things more magical than what it seems. Again, I don't quite get the distinction between linear and nonlinear it's an operator and if you disagree with that signal processing view than you should have bigger problems that control theory treats systems not this naming. 

I believe, somewhat strongly that all this is coming from terrible matlab naming schemes and it's not the fault of python-control. Folks should stop treating matlab as the canon instead and move on.

frohro:
@ilayn,
It seems to me that the initial condition can be caused by an impulse or by a lot of other functions driving it before the simulation is started. I like that we have a set of simulations that are strictly linear.  The great thing about linear systems is you can use superposition and proportionality.  This allows things like convolution, Laplace Transforms, and Fourier series.  You can use ideas from linear algebra, which are so helpful in solving real-world problems.  The EE students here get several courses in linear systems, Circuit Analysis, Linear Systems Analysis, Signals and Systems, and Linear Algebra just as an undergraduate.  If a system isn't linear, we often linearize it to make a solution easier.  So I think understanding linear systems is important, and the terms we use to describe them are also important because they help us all communicate better and more easily.  I am not pushing this language, because MATLAB does, but because it is the language EE's have been using, at least since I was an undergrad 40 years ago. 
Rob

ilayn:
Yes I agree linear systems is the key and they play important role but your argument is basically your last sentence, and if I am understanding correctly, just keep things the same don't change them which I cannot participate in sharing your enthusiasm about it. What is it that goes against the grain here? The initial conditions or matlab incompatibility. I don't quite understand the argument here. 

The matlab syntax is/was clearly wrong for many decades why should we continue that tradition?

frohro:
@ilayn, I guess I don't understand why MATLAB was wrong and the present control package is right.  Can you help me?  Maybe it would help to answer my three points, with proofs or citations why they and MATLAB or wrong?  Sorry I'm so dense.  :-)

murrayrm:
Just to weigh in a bit:

* We have a way to handle compatibility with MATLAB: the `control.matlab` module implements things in the same way as MATLAB for those who prefer that interface (including `lsim`).  If there are places in `control.matlab` that are not compatible with MATLAB, we should fix those.
* Some of the problems we are arguing about relate to the history of how these functions came about.  In particular, `input_output_response` was added when we put in nonlinear models and (at that time) state space system were a separate class (now they are effectively a subclass) => we needed both a method for linear systems (`forced_response`) and nonlinear systems (`input_output_response`).
* Related to that, part of the reason that `forced_response` allows the initial condition to be given is that prior to `input_output_response`, there was no other function that allowed you to specify both an input function and a non-zero initial condition.  I personally agree that `lsim` is not a great name and that the `control` module (versus the `control.matlab` module) should be Pythonic rather that "MATLABic" and the convention we have been using is to write things out (`bode_plot`, `initial_response`, etc).

Having said all of that, as we move into 0.10 (see #904), I think we can do the following:
* Allow `initial_response` and `step_response` to work for either linear or nonlinear systems.
* Allow `forced_response` to take a nonzero X0 as an optional argument, but perhaps document it as mainly being there for the "pure" forced response, which at least one textbook defines as corresponding to X0=0 (FBS, page 6-3).
* Use `input_output_response` as the general function for simulating a linear or nonlinear system for a specified input and initial condition (suggestions welcome for a better name).
* Allow `impulse_response` only for linear systems (mainly because I think it is mainly useful for educational purposes when teaching linear control theory, convolution, etc).

ilayn:
>  Sorry I'm so dense. :-)
I did not imply anything personal with my remarks, and I don't want the discussion to derail to such direction. 

>  Maybe it would help to answer my three points, with proofs or citations why they and MATLAB or wrong?

If you bother to check matlab's documentation on [lsim](https://www.mathworks.com/help/control/ref/dynamicsystem.lsim.html) it accepts both linear and nonlinear models, LPV and identified models of every kind. Thus if we were to be hygenic about linear nonlinear distinction, matlab itself is contradicting its rule. That is what I mean by wrong naming/syntax.

What you probably have in mind for "forced response" is the differential equation solutions; homogeneous solution when you set input term to zero (complementary equation) and non-homogeneous  solutions and so on. That is not the right terminology to use if you call nonhomogeneous a forced response. Because like I mentioned, initial conditions **are** also forcing terms or say inputs happened before t<0. "Forced response" term for zero RHS is a sloppy usage in that respect. 

The common terminology for such things are, that I am sure you know vrey well, typically zero-input response (ZIR) (which is called `initial` in matlab historically) and Zero state response ZSR which initial condition is zero and only input is driving to denote exactly what you are taking as 0.

This in fact opens up for nonlinear systems since we have concepts like zero dynamics and so on.

PR #921, opened 2023-07-06
Add final point to docstrings, PEP257

KybernetikJo:
Add final point to docstrings. According to [PEP257](https://peps.python.org/pep-0257/), "The docstring is a phrase ending in a period."

Makes the documentation consistent.


murrayrm:
Thanks for the fixes!

PR #922, opened 2023-07-11
Doc remove linear io system

KybernetikJo:
- Replace LinearIOSystem with StateSpace and LinearICSystem in docstrings
- Reruns jupyter notebooks

I have a question:
When a PR is merged, the sphinx help is built, but jupyter notebooks are not executed again? Is that right?

I am not sure if the second commit (Rerun jupyter ....) is really necessary.


murrayrm:
@KybernetikJo I don't believe that the Jupyter notebooks are regenerated when generating Sphinx output on readthedocs.  So what you did is required if you want readthedocs to show the output correctly.

Issue #929, opened 2023-09-06
Constrained MPC: Differences to MATLAB and Numeric Issues for different OS


JMK593:
In the context of a masters course on constrained optimal control, we were looking to implement the classic cart-pendulum example, with a hard state constraint on the angle of the pendulum. Our implementation followed the constrained MPC [aircraft model](https://python-control.readthedocs.io/en/0.9.4/mpc_aircraft.html) example.

During our implementation we observed that the output control sequences from the “optimal.solve_ocp” were different than from MATLAB’s “mpcActiveSetSolver” for the same model, stage cost, and prediction horizon. The following are the control sequences at time 0:
solve_ocp: 8.085e-01  1.464e-02  4.912e-01  7.621e-01  5.672e-01 5.080e-01  4.151e-01  2.398e-01  0.000e+00  0.000e+00
MATLAB: 0.8750 -0.1281 0.6964 0.7551 0.7363 0.7937 0.6934 0.6644 0.6041 0.4130

We also observed that the output from “input_output_response” simulation differed based on platform (Windows vs Linux/MacOS). On Linux/MacOS the optimiser would fail to meet the state constraints, while this behaviour does not occur on Windows.

I have included an implementation of our code in both Python and in JupyterLab, environment package lists, plots of the constrained state, and a comparison version in MATLAB: [MPC_PythonControl_Analysis.zip](https://github.com/python-control/python-control/files/12534268/MPC_PythonControl_Analysis.zip)

Are there any plans for classic MPC implementation for the cost function to directly compute the state predictions using the discrete-time linear dynamics rather than numerically computed using the trapezoid rule (line 311-322 in optimal.py)?
 
Are you able to provide any advice or guidance on why the numeric simulation result is different between OS versions?

murrayrm:
Thanks for your comments!

The `control.optimal` module  is just a wrapper around [`scipy.optimize.minimize`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html), so the differences in platforms probably trace back to SciPy (nothing in python-control is OS dependent, since it is pure Python).

>During our implementation we observed that the output control sequences from the “optimal.solve_ocp” were different than from MATLAB’s “mpcActiveSetSolver” for the same model, stage cost, and prediction horizon.

It's odd that the results seem to be very different from MATLAB, since they should be solving the same underlying optimization problem (though with different solvers).  One thing to try might be using different solvers (using the `minimize_method` keyword) to see if that changes the results.

[...]

PR #930, opened 2023-09-10
create_statefbk_iosystem and optimal control enhancements

murrayrm:
This PR has two primary changes. motivated by some debugging I was doing in #929:

* The `create_statefbk_iosystem` function now allows you to pass an I/O system instead of the a gain (or gain schedule) for the controller.  This allows a more general state feedback controller (like an MPC controller) to be created using `create_statefbk_iosystem`.  (I needed this so that I could gain easy access to the inputs in issue [#929](https://github.com/python-control/python-control/issues/929), since the `feedback` function does not pass through the controller inputs as a system output, but `create_statefbk_iosystem` does.)
* The `create_mpc_iosystem` function now properly allows keyword parameters that go to the underlying `OptimalControlProblem`, for example setting the minimization method.  Prior to this, the keyword parameters were passed to `OptimalControlProblem` but were also passed to `NonlinearIOSystem`, where they would generate an error.
* Unit tests and documentation updated as well.  More specifically, the documentation for the `solve_ocp` function has been updated to make more clear how the final time point is handled (see issue [#929](https://github.com/python-control/python-control/issues/929) for some discussion on this).
* Had to modify the `install_examples.yml` workflow to explicitly include the latest python version, otherwise an early version of `matplotlib` got installed that was not current enough to work with current code (not sure why this was suddenly required...).


doc/optimal.rst
```diff
@@ -65,6 +65,13 @@ can be on the input, the state, or combinations of input and state,
 depending on the form of :math:`g_i`.  Furthermore, these constraints are
 intended to hold at all instants in time along the trajectory.
 
+For a discrete time system, the same basic formulation applies except
+that the cost function is given by
+
+.. math::
+
+  J(x, u) = \sum_{k=0}^{N-1} L(x_k, u_k)\, dt + V \bigl( x_N \bigr).
```
sawyerbfuller, line 73: why does V(x_N) need big parens but L(.) does not?

sawyerbfuller: above,  V(.) gets regular sized parens.

sawyerbfuller:
These seem like obvious improvements to make, so I see no issue with them. Just one tiny note in the comments.

sawyerbfuller (approved):

Issue #935, opened 2023-10-17
tf2ss() returns unstable system


pooyasf:
I have the following transfer function:

`_sys1.den=[[array([9.43513863e-11, 6.05312352e-08, 7.92752628e-07, 5.23764693e-06,
         1.82502556e-05, 1.24355899e-05, 8.68206174e-06, 2.73818482e-06,
         4.29133144e-07, 3.85554417e-08, 1.62631575e-09, 8.41098151e-12,
         9.85278302e-15, 4.07646645e-18, 5.55496497e-22, 3.06560494e-26,
         5.98908988e-31])]]`

'_sys1.num=[[array([9.94004350e-13, 2.67602795e-11, 2.31058712e-10, 1.15119493e-09,
         5.04635153e-09, 1.34066064e-08, 2.11938725e-08, 2.39940325e-08,
         2.05897777e-08, 1.17092854e-08, 4.71236875e-09, 1.19497537e-09,
         1.90815347e-10, 1.00655454e-11, 1.47388887e-13, 8.40314881e-16,
         1.67195685e-18])]]'

asking for poles:

`_sys1.poles()=array([-6.28318525e+02+0.00000000e+00j, -6.28318531e+00+0.00000000e+00j,
       -3.14159265e+00+5.44139809e+00j, -3.14159265e+00-5.44139809e+00j,
       -1.19380521e-01+5.84842723e-01j, -1.19380521e-01-5.84842723e-01j,
       -1.88495559e-01+0.00000000e+00j, -6.63225116e-02+9.92624461e-02j,
       -6.63225116e-02-9.92624461e-02j, -1.00530965e-01+0.00000000e+00j,
       -4.32233973e-03+0.00000000e+00j, -6.28318476e-04+0.00000000e+00j,
       -6.28318592e-04+0.00000000e+00j, -6.29820167e-05+2.60763321e-07j,
       -6.29820167e-05-2.60763321e-07j, -6.25315230e-05+0.00000000e+00j])`

All negative. System is stable. But then when converting _sys1 to state space using tf2ss() returns a system which have positive poles (unstable). I am using Slycot.

I also repeated the same conversion but using scipy and it gives me a stable state space as a result.

python-control version: '0.9.3.post2'

murrayrm:
[...]

Issue #222, opened 2018-07-07
Update conda-forge recipe to use 


murrayrm:
While releasing v0.8.0 on conda-forge, I got the following warning:
```
This project is currently running on CircleCI 1.0 which will no longer be supported after 
August 31, 2018. Please start migrating this project to CircleCI 2.0.
```
Not sure what we need to change (or where, but probably [conda-forge/control-feedstock](https://github.com/conda-forge/control-feedstock)), but flagging this here for now.

moorepants:
Probably just needed a rerender: https://conda-forge.org/docs/conda_smithy.html

moorepants:
0.8.0 is now on conda forge: https://github.com/conda-forge/control-feedstock/pull/3

PR #937, opened 2023-10-22
Add unit test illustrating issue #935 + add method keyword for tf2ss

murrayrm:
This PR puts in some unit tests associated with issue #935, which looks like it is a problem in slycot where the tf2ss functionality (in `td04ad`) doesn't work properly.  As a partial workaround, a `method` keyword is added to `tf2ss` (and `ss`) that allows `method='scipy'` to be used for SISO systems.  Docstrings updated for the new keyword and a note on the `slycot` issue.

A bug report has been created in the `slycot` repository ([#222](https://github.com/python-control/Slycot/issues/222) and the unit tests will fail when the bug is fixed (so we can go back and update the unit tests and documentation).

### DIFF
--- a/control/statesp.py
+++ b/control/statesp.py
@@ -1583,6 +1583,13 @@ def ss(*args, **kwargs):
     --------
     tf, ss2tf, tf2ss
 
+    Notes
+    -----
+    If a transfer function is passed as the sole positional argument, the
+    system will be converted to state space form in the same way as calling
+    :func:`~control.tf2ss`.  The `method` keyword can be used to select the
+    method for conversion.
+
     Examples
     --------
     Create a Linear I/O system object from matrices.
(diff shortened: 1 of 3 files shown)


control/statesp.py
```diff
@@ -2250,6 +2274,8 @@ def _convert_to_statespace(sys, use_prefix_suffix=False):
                 A, B, C, D = \
```
sawyerbfuller, line 2274: Suggested change for the two lines above: 
```
if not issiso(sys):
    raise ControlMIMONotImplemented("MIMO system conversion not supported without Slycot")
```

control/statesp.py
```diff
@@ -2250,6 +2274,8 @@ def _convert_to_statespace(sys, use_prefix_suffix=False):
...
                     sp.signal.tf2ss(squeeze(sys.num), squeeze(sys.den))
                 newsys = StateSpace(A, B, C, D, sys.dt)
+        else:
+            raise ValueError(f"unknown {method=}")
```
sawyerbfuller, line 2278: First instantiation of self-documenting F-strings in python-control? : )

sawyerbfuller (approved):

PR #940, opened 2023-11-08
Vectorize cost calculation

urpok23:
Small todo fix.


murrayrm:
Thanks for the fix @urpok23!

PR #941, opened 2023-11-25
Improved speed of ctrb and obsv functions

Jpickard1:
Current implementation of ctrb and obsv functions utilize `np.linalg.matrix_power` inside of a `for i in range(1,n)` loop. I modified these functions to reduce the number of matrix multiplications by `amat` essentially by utilizing the precomputed matrix exponentiation from the `i-1`the iteration of the loop when computing the `i`th iteration. This is similar to the MATLAB implementation of the similar functions.

Additionally, I added an optional parameter `t` to both functions that defaults to `None` for the case where the reduced observability and controllability matrices are needed. By default, the full observability and controllability matrices are created. 

This appears to pass all test cases.

### DIFF
--- a/control/statefbk.py
+++ b/control/statefbk.py
@@ -1027,7 +1027,7 @@ def ctrb(A, B, t=None):
     ctrb = np.zeros((n, t * m))
     ctrb[:, :m] = bmat
     for k in range(1, t):
-        ctrb[:, k * m:(k + 1) * m] = amat @ ctrb[:, (k - 1) * m:k * m]
+        ctrb[:, k * m:(k + 1) * m] = np.dot(amat, ctrb[:, (k - 1) * m:k * m])
 
     return _ssmatrix(ctrb)
 
@@ -1070,7 +1070,7 @@ def obsv(A, C, t=None):
     obsv[:p, :] = cmat
         
     for k in range(1, t):
-        obsv[k * p:(k + 1) * p, :] = obsv[(k - 1) * p:k * p, :] @ amat
+        obsv[k * p:(k + 1) * p, :] = np.dot(obsv[(k - 1) * p:k * p, :], amat)
 
     return _ssmatrix(obsv)
 
--- a/control/statefbk.py
+++ b/control/statefbk.py
@@ -1024,7 +1024,7 @@ def ctrb(A, B):
     return _ssmatrix(ctrb)
 
 
-def obsv(A, C):
+def obsv(A, C, t=None):
     """Observability matrix.
 
     Parameters
@@ -1050,10 +1050,17 @@ def obsv(A, C):
     amat = _ssmatrix(A)
     cmat = _ssmatrix(C)
     n = np.shape(amat)[0]
+    
+    if t is None:
+        t = n
 
     # Construct the observability matrix
-    obsv = np.vstack([cmat] + [cmat @ np.linalg.matrix_power(amat, i)
-                               for i in range(1, n)])
+    obsv = np.zeros((t * ny, n))
+    obsv[:ny, :] = c
+        
+    for k in range(1, t):
+        obsv[k * ny:(k + 1) * ny, :] = np.dot(obsv[(k - 1) * ny:k * ny, :], a)
+
     return _ssmatrix(obsv)
 
 
--- a/control/statefbk.py
+++ b/control/statefbk.py
@@ -998,7 +998,7 @@ def ctrb(A, B, t=None):
     A, B : array_like or string
         Dynamics and input matrix of the system
     t : None or integer
-        maximum time horizon of the controllability matrix, max = n
+        maximum time horizon of the controllability matrix, max = A.shape[0]
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


control/statefbk.py
```diff
@@ -1016,22 +1018,30 @@ def ctrb(A, B):
...
+        t = n
 
     # Construct the controllability matrix
-    ctrb = np.hstack(
-        [bmat] + [np.linalg.matrix_power(amat, i) @ bmat
-                  for i in range(1, n)])
+    ctrb = np.zeros((n, t * m))
+    ctrb[:, :m] = bmat
+    for k in range(1, t):
+        ctrb[:, k * m:(k + 1) * m] = np.dot(amat, ctrb[:, (k - 1) * m:k * m])
```
sawyerbfuller, line 1030:
```suggestion
        ctrb[:, k * m:(k + 1) * m] = amat @ ctrb[:, (k - 1) * m:k * m]
```

control/statefbk.py
```diff
@@ -1050,10 +1060,18 @@ def obsv(A, C):
...
+        t = n
 
     # Construct the observability matrix
-    obsv = np.vstack([cmat] + [cmat @ np.linalg.matrix_power(amat, i)
-                               for i in range(1, n)])
+    obsv = np.zeros((t * p, n))
+    obsv[:p, :] = cmat
+        
+    for k in range(1, t):
+        obsv[k * p:(k + 1) * p, :] = np.dot(obsv[(k - 1) * p:k * p, :], amat)
```
sawyerbfuller, line 1073:
```suggestion
        obsv[k * p:(k + 1) * p, :] = obsv[(k - 1) * p:k * p, :] @ amat
```

sawyerbfuller (changes requested):
Thanks for the contribution!

A couple of comments: 

1. Have you benchmarked this to confirm it's actually faster? Also, I am not aware of a use case where the obsv or ctrb matrix must be repeatedly evaluated - is there one? 
2. can you add a unit test for obsv and ctrb for the t < n case? 
3. can you define n in the docstrings?

control/statefbk.py
ilayn: If performance is needed here, you can use the [`numpy.matmul`](https://numpy.org/doc/stable/reference/generated/numpy.matmul.html) with `out=ctrb[:, k * m:(k + 1) * m]` to avoid the interim array.

ilayn:
This is actually not only for performance but for accuracy reasons, the right way to go. Hence it is an improvement anyways. It is basically a Krylov space argument that leads to repeated powers of `A` and not that the powers have to be calculated (which is for some reason, always skipped though makes the exposition very accessible).

However, this test is fundamentally broken for production code; i) it is only a theoretical tool since the powers blow up regardless ii) rank decision is a very tricky thing to nail it down without false positives and false negatives iii) as if it is not enough, the answer is a binary yes or no. It does not give any indication of how close we are dancing to uncontrollability/unobservability. 

A typical and best bang for buck candidate  is D. Boley's test if you are really looking for a better test which computes the cancellation distance in our context (but originally checks for the rank deficiency metric) Then you can put a limit on the distance and you can tie it to the matrix norms and so on. In my opinion a much better candidate instead of a lower semi-continuous function.

I have the [regular version here](https://github.com/ilayn/harold/blob/main/harold/_system_funcs.py#L225-L294) in case you want to check it out. The article is in the docstring. 

I hope academia emphasizes a bit more that the Kalman test is just a theoretical tool and we use it just on the homework problems.

Jpickard1:
Dear Ilhan,

Thank you. I agree there are several issues with utilizing the Kalman test for determining controllability or observability, and thank you for sharing the test by Boley.

My particular motivation for this change is to use Gramians to determine observability. This is advantageous to me because in addition to providing a Yes/No test, the Gramian provides information on the principal directions of controllability/observability. Since discrete, finite horizon Gramians can be computed in terms of `obsv(A,C,t=T)`, it seems like adding `t` as an argument would be a nice functionality to include in the code.

While there are these limitations that you have noted, it seems to me that we are in agreement this is a beneficial argument to add for the `ctrb` and `obsv` functions. Please share if you think I am misunderstanding your points above or if there are other changes that need to be made.

Thank you. Best,
Joshua Pickard

Jpickard1:
Dear all,

It appears that this is failing with the below error related to `Doctest / doctest-linux (pull_request)`, and I am hoping I could have some pointers to fix this. This error appears to occur surrounding the doctest for `control.forced_response`, which is a section of code that I do not believe was modified on my fork or for this pull request. I am not familiar with the doctests for this repo, so could you please advise if there are changes that I should make on my end to resolve this error or if the error exist more broadly in the repo.

Thank you. Best,
Joshua Pickard
```
ERROR: [numpydoc] While processing docstring for 'control.forced_response'

Extension error (numpydoc.numpydoc):
Handler <function mangle_docstrings at 0x7fa850237ec0> for event 'autodoc-process-docstring' threw an exception (exception: Expected a class or None, but got <function forced_response at 0x7fa853992ac0>)
make: *** [Makefile:32: html] Error 2
reading sources... [ 26%] generated/control.forced_response
Error: Process completed with exit code 2.
```

bnavigator:
It's failing in parts of the code unaffected by your change, so don't worry about in this PR.

sawyerbfuller:
I tried re-running the failing test once or twice but it kept failing. I agree that it has nothing to do with this PR, so I'll merge it.

Issue #950, opened 2023-12-24
gain offset using control.matlab.c2d using method='matched' ?


JonHowMIT:
Compared results of 
sysc = ct.TransferFunction([1, z], [1, p])
sysd = ct.matlab.c2d(sysc, Ts, method='bilinear')
sysd1 = ct.matlab.c2d(sysc, Ts, method='matched')

yields a sysd1 with a gain that is about 10x that of sysd - is that expected?

sawyerbfuller:
Hi John great to hear from you and happy holidays. Glad to see you are using or at the very least evaluating the library. 

The matched c2d is code I have never used before and it doesn’t look like we have a unit test for it, so there is a decent chance it is just a bug. Thanks for the report! 

One q - when you refer to gain, do you mean the DC gain of the two systems?

murrayrm:
I did a quick check and I agree there is something wrong in the matched transformation:
```
import control as ct
sysc = ct.tf([1, 2], [1, 3], name='sysc')
sysd_bilinear = ct.sample_system(sysc, 0.1, method='bilinear', name='sysd_bilinear')
sysd_matched = ct.sample_system(sysc, 0.1, method='matched', name='sysd_matched')
ct.bode([sysd_bilinear, sysd_matched, sysc])
```

This produces the following plot:
![Figure_1](https://github.com/python-control/python-control/assets/293362/529611c2-5a08-487c-9342-be9034b20c4e)

A couple of things to fix:
* Zero frequency gain for the 'matched' transformation is incorrect.
* The name of the system is lost when doing the 'matched' transformation.
* The order of the system names is incorrect in the figure title

murrayrm:
Fixed in #951.  New plot:

![Figure_2](https://github.com/python-control/python-control/assets/293362/98d0d19e-68a6-4f3e-8bbb-e8e9b26fe1b1)

PR #951, opened 2023-12-26
Fix bug in matched transformation + address other issues in #950

murrayrm:
This PR fixes the calculation of the DC gain in the 'matched' transformation from continuous time to discrete time.  It also fixes some other issues identified in #950.  Also adds unit test for `sample_system` that catches the original error.

### DIFF
--- a/control/xferfcn.py
+++ b/control/xferfcn.py
@@ -1181,7 +1181,9 @@ class TransferFunction(LTI):
         if not self.issiso():
             raise ControlMIMONotImplemented("Not implemented for MIMO systems")
         if method == "matched":
-            return _c2d_matched(self, Ts)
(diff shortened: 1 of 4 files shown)


control/xferfcn.py
```diff
@@ -1284,9 +1286,12 @@ def _isstatic(self):
 
 
 # c2d function contributed by Benjamin White, Oct 2012
-def _c2d_matched(sysC, Ts):
+def _c2d_matched(sysC, Ts, **kwargs):
+    if sysC.ninputs > 1 or sysC.noutputs > 1:
```
sawyerbfuller, line 1290:
```suggestion
    if not sysC.issiso():
            raise ControlMIMONotImplemented("Not implemented for MIMO systems")
```

PR #953, opened 2023-12-29
Update pole/zero and root locus plots to use _map/_plot pattern

murrayrm:
This PR updates pole/zero and root locus plots to use the _map/_plot calling pattern described in #645. The following patterns will work:
```
rldata = ct.root_locus_map(syslist[, klist])
lines = rldata.plot(plot_options)

lines = ct.root_locus_plot(response[, klist, plot_options])

ct.root_locus(sys)
ct.rlocus(sys)

ct.pole_zero_map(sys).plot()
ct.pole_zero_plot(sys)
ct.pzmap(sys)
```

Everything is mostly backwards compatible except that the outputs from a _plot function are now an array of Line2D objects instead of the roots and gains.  You can still get the original data using the _map function and there is some legacy processing if you used the `plot` keyword to try let some code work without changes.

Summary of changes:
* All pole/zero and root locus plots now have a `_map` function and a `_plot` function.  You can access the latter via the `.plot()` method on the response.
* The `_plot` functions accept either the output of the `_map` function or a list of systems (in which case the `_map` function is called internally).  This allows the common pattern of `pole_zero_plot(sys)`, `root_locus_plot(sys)` to work as expected.
* The interactive mode for root locus plots is on by default (it used to be there only within `sisotool`).  Clicking on a portion of the root locus diagram will generate markers at the locations on the loci corresponding to that gain and add a message above the plot giving the frequency and damping ratio for the point that was clicked.
* The `print_gain` keyword is replaced with `interactive`.
* The (deprecated) `Plot` keyword is now `plot` (and generates a warning, since it triggers the legacy return values).
* Recalculation of the root locus plot when zooming no longer works (you can still zoom in and out, you just don't get a recalculated curve).
* The short versions `pzmap` and `rlocus` are still there.
* The `control.matlab` version of `rlocus` returns `roots, gains` (compatible with MATLAB)
* Added unit tests plus user documentation.
* Some other small fixes, code streamlining, etc along the way.

**This PR is going to break existing code.** It would be great if a few people could try this out so that we can make sure we are OK with the changes here.

Pole/zero maps and root locus diagrams provide insights into system response based on the locations of system poles and zeros in the complex plane. The `pole_zero_map()` function returns the poles and zeros and can be used to generate a pole/zero plot:
```
sys = ct.tf([1, 2], [1, 2, 3], name='SISO transfer function')
response = ct.pole_zero_map(sys)
ct.pole_zero_plot(response)
```
pzmap-siso_ctime-default

A root locus plot shows the location of the closed loop poles of a system as a function of the loop gain:
```
ct.root_locus_map(sys).plot()
```

rlocus-siso_ctime-default

The grid in the left hand plane shows lines of constant damping ratio as well as arcs corresponding to the frequency of the complex pole. The grid can be turned off using the `grid` keyword. Setting grid to `Fals`e will turn off the grid but show the real and imaginary axis. To completely remove all lines except the root loci, use `grid=’empty’`.

On systems that support interactive plots, clicking on a location on the root locus diagram will mark the pole locations on all branches of the diagram and display the gain and damping ratio below the plot title:

rlocus-siso_ctime-clicked

Root locus diagrams are also supported for discrete time systems, in which case the grid is show inside the unit circle:
```
sysd = sys.sample(0.1)
ct.root_locus_plot(sysd)
```

rlocus-siso_dtime-default

Lists of systems can also be given, in which case the root locus diagram for each system is plotted in different colors:
```
sys1 = ct.tf([1], [1, 2, 1], name='sys1')
sys2 = ct.tf([1, 0.2], [1, 1, 3, 1, 1], name='sys2')
ct.root_locus_plot([sys1, sys2], grid=False)
```

rlocus-siso_multiple-nogrid


control/sisotool.py
```diff
@@ -1,5 +1,10 @@
 __all__ = ['sisotool', 'rootlocus_pid_designer']
 
+import numpy as np
+import matplotlib.pyplot as plt
+import warnings
+from functools import partial
```
bnavigator, line 6: Worth running this throug `isort`?

`warnings` and `functools` are stdlib, Numpy and Pyplot external.

control/tests/pzmap_test.py
```diff
@@ -78,12 +83,47 @@ def test_pzmap(kwargs, setdefaults, dt, editsdefaults, mplcleanup):
...
+    # Legacy return format
+    with pytest.warns(DeprecationWarning, match=".* values .* deprecated"):
+        poles, zeros = ct.pole_zero_plot(pzdata, plot=False)
+    np.testing.assert_equal(poles, sys.poles())
+    np.testing.assert_equal(zeros, sys.zeros())
+
+    with pytest.warns(DeprecationWarning, match=".* values .* deprecated"):
+        poles, zeros = ct.pole_zero_plot(pzdata, plot=True)
+    np.testing.assert_equal(poles, sys.poles())
+    np.testing.assert_equal(zeros, sys.zeros())
```
bnavigator, line 106: Parametrize?

bnavigator (approved):
This is large, without too extensive review, I would say LGTM.

PR #955, opened 2023-12-29
Fix CI warnings 

murrayrm:
This PR makes a bunch of small changes to remove warning messages in the GitHub actions CI runs:
* Access scalar NumPy arrays using `item()` (old usage is newly deprecated in NumPy).
* Change `set_tight_layout(True)` to `fig.set_layout_engine('tight')` (newly deprecated in Matplotlib).
* Update unit tests using deprecated functionality to use current functionality.
* Add warning filters for internal calls to `connect` (which are OK for now).
* Avoid calling tf2zpk with zero in the numerator (SciPy generates warning).
* Fix up some strings that had newly deprecated escape sequences (by using raw strings)
* Turn off warnings for optimal control problems that fail to hit desired precision (but still return valid answers)

Note: it is likely that this PR will conflict with #953.  We should probably merge #953 first and then rebase this PR since the changes here are pretty straightforward.

### DIFF
--- a/control/xferfcn.py
+++ b/control/xferfcn.py
@@ -470,7 +470,7 @@ class TransferFunction(LTI):
                 elif self.display_format == 'zpk':
                     num = self.num[no][ni]
                     if num.size == 1 and num.item() == 0:
-                        # Catch a special case taht SciPy doesn't handle
+                        # Catch a special case that SciPy doesn't handle
                         z, p, k = tf2zpk([1.], self.den[no][ni])
                         k = 0
                     else:
--- a/control/xferfcn.py
+++ b/control/xferfcn.py
@@ -468,7 +468,13 @@ class TransferFunction(LTI):
                     numstr = _tf_polynomial_to_string(self.num[no][ni], var=var)
                     denstr = _tf_polynomial_to_string(self.den[no][ni], var=var)
                 elif self.display_format == 'zpk':
-                    z, p, k = tf2zpk(self.num[no][ni], self.den[no][ni])
+                    num = self.num[no][ni]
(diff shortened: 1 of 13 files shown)


control/xferfcn.py
```diff
@@ -468,7 +468,13 @@ def __str__(self, var=None):
                     numstr = _tf_polynomial_to_string(self.num[no][ni], var=var)
                     denstr = _tf_polynomial_to_string(self.den[no][ni], var=var)
                 elif self.display_format == 'zpk':
-                    z, p, k = tf2zpk(self.num[no][ni], self.den[no][ni])
+                    num = self.num[no][ni]
+                    if num.size == 1 and num.item() == 0:
+                        # Catch a special case taht SciPy doesn't handle
```
slivingston, line 473:
```suggestion
                        # Catch a special case that SciPy doesn't handle
```

bnavigator:
> Note: it is likely that this PR will conflict with #953. We should probably merge #953 first and then rebase this PR since the changes here are pretty straightforward.

Either this or the other way round. Python 3.12 requires numpy 1.26 on conda, already triggering the deprecation warnings. I had to disable them for the Slycot CI: https://github.com/python-control/Slycot/pull/227/files#diff-b2cf1d5b6cad7a0f6a749da416cb8a4e34040b2164a84a651783692d4b60d608R17

murrayrm:
Looks like there were no conflicts, so going ahead and merging.

@bnavigator: FYI (and not really for this PR), I tried updating GitHub actions to use Python 3.12 but mamba caused problems, so I'm waiting until that clears.  You can find the attempt [here](https://github.com/murrayrm/python-control/actions/runs/7441215388).

PR #964, opened 2024-01-22
Add pandas to pyproject.toml

KybernetikJo:
Adds pandas to pyproject.toml.

Allows optional install

```bash
pip install control[pandas]
pip install control[pandas,slycot]
...
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,7 @@ test = ["pytest", "pytest-timeout"]
 slycot = [ "slycot>=0.4.0" ]
 cvxopt = [ "cvxopt>=1.2.0" ]
 pandas = [ "pandas>=1.5.0" ]
+full = [ "control[slycot,cvxopt,pandas]" ]
 
 [project.urls]
 homepage = "https://python-control.org"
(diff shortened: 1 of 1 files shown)


bnavigator:
I don't really see a big enough benefit of adding a single optional **well known** package with the same name as the package.

What's the effective difference between `pip install control pandas` and `pip install 'control[pandas]'`?

bnavigator:
FTR, I am guilty of starting that kind of extra with Slycot in #678. Probably would not do it again.

KybernetikJo:
> I don't really see a big enough benefit of adding a single optional **well known** package with the same name as the package.
> 
> What's the effective difference between `pip install control pandas` and `pip install 'control[pandas]'`?

These are valid points, so I looked for other projects that use the pattern. I was able to find two well known packages:
- polars https://github.com/pola-rs/polars/blob/main/py-polars/pyproject.toml
- cvxpy https://github.com/cvxpy/cvxpy/blob/master/pyproject.toml

The reason could be
- to document all dependencies (also optional) in pyproject.toml
- to indicate that single options are valid installations `pip install control[pandas]` 
- to ensure the right version of optional packages
  - `pip install control[slycot]` is actually `pip install "slycot>=0.4.0" control`
  - `pip install control[pandas]` would be `pip install "pandas>=1.5.0" control`
  
 
Remark:
Well, the benefits don't seem to be big enough. I will keep the PR open for a few weeks, and then I will close it.

KybernetikJo:
Well, the benefits were not big enough.

PR #960, opened 2024-01-08
System norms

henriks76:
Hello,

I have written a new function norm to compute H_2 and L_infinity system norms of LTI objects.

Please do let me know if you have any suggested improvements.

Best regards,
Henrik

PR #971, opened 2024-02-17
System norms 2

henriks76:
I believe all earlier feedback on `sysnorm.py` has been addressed now.

### DIFF
--- a/control/tests/sysnorm_test.py
+++ b/control/tests/sysnorm_test.py
@@ -8,40 +8,51 @@ Author: Henrik Sandberg
 
 import control as ct
 import numpy as np
+import pytest
+
 
 def test_norm_1st_order_stable_system():
     """First-order stable continuous-time system"""
     s = ct.tf('s')
+    
     G1 = 1/(s+1)
-    assert np.allclose(ct.norm(G1, p='inf', tol=1e-9, print_warning=False), 1.0) # Comparison to norm computed in MATLAB
-    assert np.allclose(ct.norm(G1, p=2, print_warning=False), 0.707106781186547) # Comparison to norm computed in MATLAB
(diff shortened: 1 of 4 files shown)


control/tests/sysnorm_test.py
```diff
@@ -0,0 +1,74 @@
...
+import pytest
+
+
+def test_norm_1st_order_stable_system():
+    """First-order stable continuous-time system"""
+    s = ct.tf('s')
+    
+    G1 = 1/(s+1)
+    assert np.allclose(ct.norm(G1, p='inf', tol=1e-9 ), 1.0) # Comparison to norm computed in MATLAB
+    assert np.allclose(ct.norm(G1, p=2), 0.707106781186547) # Comparison to norm computed in MATLAB
```
bnavigator, line 20: See https://github.com/python-control/python-control/pull/960#discussion_r1446462119

Please don't restrict the tolerance to 1e-9. It will create problems like the one solved in #366.

bnavigator:
I have fixed the futurewarning test. Please rebase.

control/tests/sysnorm_test.py
bnavigator: That's maybe a bit coarse. I'd say we start with 1e-6 and see if we get failures, e.g. in distribution packages like the openSUSE rpm or conda after the next release.

bnavigator:
Thank you @henriks76!

henriks76:
Thanks, @KybernetikJo and @bnavigator, for all the excellent feedback! I'll look into the documentation as soon as possible.

henriks76:
Is it correct that the sphinx documentation will be automatically generated from the function docstring, or would you like me to copy it somewhere? This is all very new to me...

I have already tried to follow the conventions [here](https://github.com/python-control/python-control/wiki/Coding-conventions), but let me know if something needs fixing.

PR #976, opened 2024-03-03
norm added to the sphinx documentation

henriks76:
Added `norm` to the function reference in the sphinx documentation under Utility functions and conversions.

See #971

### DIFF
--- a/control/sysnorm.py
+++ b/control/sysnorm.py
@@ -94,13 +94,13 @@ def norm(system, p=2, tol=1e-6, print_warning=True, method=None):
         Type of norm to be computed. ``p=2`` gives the H2 norm, and ``p='inf'`` gives the L-infinity norm.
     tol : float
         Relative tolerance for accuracy of L-infinity norm computation. Ignored
-        unless p='inf'.
+        unless ``p='inf'``.
     print_warning : bool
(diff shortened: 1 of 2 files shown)


bnavigator (changes requested):
Need to fix the tolerance in the docstring example

bnavigator:
```
Document: generated/control.norm
--------------------------------
**********************************************************************
File "../control/sysnorm.py", line ?, in default
Failed example:
    ct.norm(Gc, 'inf', tol=1e-4, method='scipy')
Expected:
    1.00006103515625
Got:
    0.99993896484375
**********************************************************************
1 items had failures:
   1 of   3 in default
3 tests in 1 items.
2 passed and 1 failed.
***Test Failed*** 1 failures.
```

You won't get any futher by setting the example to 0.99993896484375. Different platform yield different values. Use an ellipsis or round the output:

https://docs.python.org/3/library/doctest.html#warnings

bnavigator:
Or go to pytest-doctestplus: https://github.com/scientific-python/pytest-doctestplus?tab=readme-ov-file#floating-point-comparison

bnavigator (approved):
LGTM

PR #978, opened 2024-03-20
remove np.int

AbhiramShenoiKodiak:
In `numpy` versions >1.24 the usage of `np.int` has been deprecated. This is the only file that uses this.

This PR makes a change to allow `python-control` to be compatible with `numpy`  > 1.24.

https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

### DIFF
--- a/external/yottalab.py
+++ b/external/yottalab.py
@@ -510,7 +510,7 @@ def placep(A,B,P):
     wrka = zeros((n,m))
     wrk1 = zeros(m)
     wrk2 = zeros(m)
-    iwrk = zeros((m),np.int)
+    iwrk = zeros((m),np.int64)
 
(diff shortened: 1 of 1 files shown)


bnavigator:
Thanks, but this is unused old code. This should be "fixed" when actually merged into python-control or be removed.

An thoughts @robertobucher, @murrayrm?

bnavigator:
Also, if we fix it, Plain `int` instead of `np.int64` should be used.

murrayrm:
I think we can probably remove the `external/` directory.  It was there from a time when we were missing some code and I thought people might want to see other implementations.  We don't actively use it and it is probably mostly out of date.

I'll post a short PR to remove it and close out this PR.

PR #980, opened 2024-03-28
Reimplementation of 2D phase plots

murrayrm:
This PR reimplements the phase plot functionality of python-control by creating a new function `phase_plane_plot` and creating a module `phaseplot` that provides more specialized functionality.  The legacy `phase_plot` function is still available (with a deprecation warning).

From the documentation:

The default method for generating a phase plane plot is to provide a 2D dynamical system along with a range of coordinates and time limit:

    sys = ct.nlsys(
        lambda t, x, u, params: np.array([[0, 1], [-1, -1]]) @ x, 
        states=['position', 'velocity'], inputs=0, name='damped oscillator')
    axis_limits = [-1, 1, -1, 1]
    T = 8
    ct.phase_plane_plot(sys, axis_limits, T)

phaseplot-dampedosc-default

By default, the plot includes streamlines generated from starting points on limits of the plot, with arrows showing the flow of the system, as well as any equilibrium points for the system.  A variety of options are available to modify the information that is plotted, including plotting a grid of vectors instead of streamlines and turning on and off various features of the plot.

To illustrate some of these possibilities, consider a phase plane plot for an inverted pendulum system, which is created using a mesh grid:

        return [x[1], -b/m * x[1] + (g * l / m) * np.sin(x[0]) + u[0]/m]
    invpend = ct.nlsys(invpend_update, states=2, inputs=1, name='invpend')
        
    ct.phase_plane_plot(
        invpend, [-2*pi, 2*pi, -2, 2], 5,
        gridtype='meshgrid', gridspec=[5, 8], arrows=3,
        plot_equilpoints={'gridspec': [12, 9]},
    plt.xlabel(r"$\theta$ [rad]")
    plt.ylabel(r"$\dot\theta$ [rad/sec]")
    
phaseplot-invpend-meshgrid

This figure shows several features of more complex phase plane plots: multiple equilibrium points are shown, with saddle points showing separatrices, and streamlines generated along a 5x8 mesh of initial conditions.  At each mesh point, a streamline is created that goes 5 time units forward and backward in time.  A separate grid specification is used to find equilibrium points and separatrices (since the course grid spacing of 5x8 does not find all possible equilibrium points).  Together, the multiple features in the phase plane plot give a good global picture of the topological strucrure of solutions of the dynamical system.

Phase plots can be buit up by hand using a variety of helper functions that are part of the `control.phaseplot` (pp) module:

    import control.phaseplot as pp

    def oscillator_update(t, x, u, params):
        return [x[1] + x[0] * (1 - x[0]**2 - x[1]**2),
                -x[0] + x[1] * (1 - x[0]**2 - x[1]**2)]
    oscillator = ct.nlsys(
        oscillator_update, states=2, inputs=0, name='nonlinear oscillator')
  
    ct.phase_plane_plot(oscillator, [-1.5, 1.5, -1.5, 1.5], 0.9)
    pp.streamlines(
        oscillator, np.array([[0, 0]]), 1.5,
        gridtype='circlegrid', gridspec=[0.5, 6], dir='both')
    pp.streamlines(
        oscillator, np.array([[1, 0]]), 2*pi, arrows=6, color='b')
    plt.gca().set_aspect('equal')

phaseplot-oscillator-helpers

The following helper functions are available:

* pp.equilpoints
* pp.separatrices
* pp.streamlines
* pp.vectorfield

The `phase_plane_plot` function calls these helper functions based on the options it is passed.

Note that unlike other plotting function, phase plane plots do not involve computing a response and then plotting the result via a `plot()` method. Instead, the plot is generated directly be a call to the `phase_plane_plot` function (or one of the `ct.phaseplot` helper functions.


control/phaseplot.py
```diff
@@ -1,61 +1,912 @@
...
+# TODO
+# * Allow multiple timepoints (and change timespec name to T?)
+# * Update linestyles (color -> linestyle?)
+# * Check for keyword compatibility with other plot routines
+# * Set up configuration parameters (nyquist --> phaseplot)
 
-import numpy as np
-import matplotlib.pyplot as mpl
+"""The :mod:`control.phaseplot` module contains functions for generating 2D
+phase plots.
```
slivingston, line 13: Convention is for summary to be 1 line (<https://peps.python.org/pep-0257/#multi-line-docstrings>).

examples/phase_plane_plots.py
```diff
@@ -0,0 +1,215 @@
+# phase_portraits.py - phase portrait examples
```
slivingston, line 1:
```suggestion
# phase_plane_plots.py - phase portrait examples
```

doc/Makefile
```diff
@@ -29,6 +30,9 @@ freqplot-siso_bode-default.png: ../control/tests/freqplot_test.py
 rlocus-siso_ctime-default.png: ../control/tests/rlocus_test.py
 	PYTHONPATH=.. python $<
 
+phaseplot-dampedosc-default.png: ../control/tests/phaseplot_test.py
+	PYTHONPATH=.. python $<
```
slivingston, line 34: Should the other images generated by phaseplot_test.py (`phaseplot-oscillator-helpers.png`, `phaseplot-invpend-meshgrid.png`) be included here? It does not seem important because the PNG files are being committed, but I note this here in case you wanted to.

control/phaseplot.py
```diff
@@ -1,61 +1,912 @@
...
+
+    # Make sure all keyword arguments were processed
+    if check_kwargs and kwargs:
+        raise TypeError("unrecognized keywords: ", str(kwargs))
+
+    # Generate phase plane (quiver) data
+    vfdata = np.zeros((points.shape[0], 4))
+    for i, x in enumerate(points):
+        vfdata[i, :2] = x
+        vfdata[i, 2:] = sys.dynamics(0, x, 0, params=params)
```
sawyerbfuller, line 259: I think this wouldn't work if `sys` is just a function - is the callable intended to be a function if it is not a system?

sawyerbfuller (changes requested):
Fixes some very outdated code. Very nice!

doc/Makefile
murrayrm: Agree that they could be listed, but I didn't list them all out since triggering on the one file will update all files => if the test generation script is updated, all PNG files are updated.

murrayrm:
In addition to adressing comments, functionality for handling callables (versus systems) was missing; now added.

doc/Makefile
slivingston: Good point! I will complete review tomorrow.

control/phaseplot.py
```diff
@@ -1,61 +1,934 @@
...
+    color = _get_color(kwargs, ax)
+
+    # Make sure all keyword arguments were processed
+    if check_kwargs and kwargs:
+        raise TypeError("unrecognized keywords: ", str(kwargs))
+
+    # Create reverse time system, if needed
+    if dir != 'forward':
+        revsys = NonlinearIOSystem(
+            lambda t, x, u, params: -np.array(sys.updfcn(t, x, u, params)),
```
sawyerbfuller, line 362:
```suggestion
            lambda t, x, u, params: -np.asarray(sys.updfcn(t, x, u, params)),
```
This won't re-create the array if it is already one. 

relatedly, and this is more of a question: I am confused about the use of each of these methods, all of which do roughly the same thing. Is this correct:  `_rhs` fast and low-level, `updfcn` just a reference to the function that was passed in, and `dynamics`: user-friendly function that may not be as fast? Just checking that this is the right one (speed is kind of important in this application)

sawyerbfuller (approved):
LGTM

control/phaseplot.py
murrayrm: Correct on the various versions.  In general:
* _updfcn() should not be called directly (though OK here, because we are creating a new system).
* _rhs() should be called inside loops (with a call to _update_params() before the loop starts, which sets the parameter values)
* dynamics() is the user-callable function (which calls _update_params and then _rhs)

control/phaseplot.py
```diff
@@ -1,61 +1,935 @@
...
-# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
-# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
-# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
-# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
-# IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
-# POSSIBILITY OF SUCH DAMAGE.
 
-import numpy as np
-import matplotlib.pyplot as mpl
+# Create a system form a callable
```
slivingston, line 705:
```suggestion
# Create a system from a callable
```

control/phaseplot.py
```diff
@@ -1,61 +1,935 @@
...
+    if isinstance(pointdata, np.ndarray) and gridtype is None:
+        pointdata = np.atleast_2d(pointdata)
+        if pointdata.shape[1] == 2:
+            # Given a list of points => no action required
+            return pointdata, None
+
+    # Utility function to parse (and check) input arguments
+    def _parse_args(defsize):
+        # if not isinstance(pointdata, (list, tuple)) or len(pointdata) != 4:
+        #     raise ValueError("invalid grid data specification")
```
slivingston, line 793:
```suggestion
```
This can be deleted, or the check in the comment can be expanded to handle case of `isinstance(pointdata, np.ndarray)`.

slivingston (approved):
Great work! I found a few more misprints, but nothing critical. Locally I merged this into the current tip of `main` branch and successfully ran the tests and example code, and built the documentation.

murrayrm:
Thanks for the careful reviews, @sawyerbfuller and @slivingston!  Will merge as soon as checks finish up (just in case).

Issue #987, opened 2024-04-03
Wrong parameter name in documentation


Davide-sd:
I'm adapting my codes to `control 0.10.0`. In particular I'm switching from `ct.root_locus` to `ct.root_locus_map` in order to compute the root locus data. An error was raised in one of my tests, which led me to discover this. The documentation associated to `root_locus_map` is:

```py
def root_locus_map(sysdata, gains=None):
    """Compute the root locus map for an LTI system.

    Calculate the root locus by finding the roots of 1 + k * G(s) where G
    is a linear system with transfer function num(s)/den(s) and each k is
    an element of kvect.

    Parameters
    ----------
    sys : LTI system or list of LTI systems
        Linear input/output systems (SISO only, for now).
    kvect : array_like, optional
        Gains to use in computing plot of closed-loop poles.
    ...
    """
```

Note that (the new) `gains` is used as a function parameter, but `kvect` is documented (the old parameter from `ct.root_locus`).

bnavigator:
Thanks! I made a small PR in #988. If there is more, please tell us or create your own PR!

PR #989, opened 2024-04-05
Update documentation, processing of root_locus kvect keyword

murrayrm:
This PR addresses issue #987, which was partially fixed in #988, by consistently using the keyword `gains` instead of `kvect` in the root locus commands.  In addition, if the `kvect` keyword is used, a warning is generated indicating that `kvect` has been deprecated and `gains` should be used instead.

Also include some minor reformatting of docstrings in rlocus.py.

### DIFF
--- a/control/rlocus.py
+++ b/control/rlocus.py
@@ -146,9 +146,8 @@ def root_locus_plot(
 
     roots, gains : ndarray
         (legacy) If the `plot` keyword is given, returns the closed-loop
-        root locations, arranged such that each row corresponds to a gain
-        in gains, and the array of gains (ame as gains keyword argument if
-        provided).
+        root locations, arranged such that each row corresponds to a gain,
+        and the array of gains (same as `gains` keyword argument if provided).
 
     Notes
     -----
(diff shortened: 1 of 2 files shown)


control/rlocus.py
```diff
@@ -145,10 +145,10 @@ def root_locus_plot(
         * lines[idx, 2]: loci
 
     roots, gains : ndarray
-        (legacy) If the `plot` keyword is given, returns the
-        closed-loop root locations, arranged such that each row
-        corresponds to a gain in gains, and the array of gains (ame as
-        kvect keyword argument if provided).
+        (legacy) If the `plot` keyword is given, returns the closed-loop
+        root locations, arranged such that each row corresponds to a gain
+        in gains, and the array of gains (ame as gains keyword argument if
```
bnavigator, line 150:
```suggestion
        in gains, and the array of gains (same as `gains` keyword argument if
```

slivingston (approved):
This is good to merge, except for the misprint identified by @bnavigator (the misprint existed before this PR, but might as well fix it).

Issue #992, opened 2024-04-16
Plotting nyquist plot from FrequencyResponseData fails


wueestry:
When trying to create the nyquist plot from an FRD system
```python
freq_response = ct.FRD(response, frequencies)
ct.nyquist_plot(freq_response)
```
the plotting fails with the following error
`AttributeError: 'FrequencyResponseData' object has no attribute 'poles'` during the call of `_default_frequency_range`

version: 0.10.0

slivingston:
Is it possible to make a Nyquist plot given only frequency response data?

murrayrm:
You can get a valid Nyquist plot from the frequency response in some instances and with some assumptions.  Basically, if the open loop dynamics are assumed stable, the zero frequency gain is finite, and the high frequency gain is zero, it *should* work.

The error above comes when it tries to compute the default frequency range for the plot, but in the case of frequency response data you don't actually need to do that.

A quick fix is to change the call to be `ct.nyquist_plot(freq_response, freq_response.omega)`, which bypasses the computation of the default frequency range.

murrayrm:
Fixed in #996.

PR #996, opened 2024-04-23
Fix nyquist plotting from FrequencyResponseData

wueestry:
This PR resolves the issue #992, by adding a check for FrequencyResponseData during plotting of a nyquist plot and setting omega if not yet set correctly

### DIFF
--- a/control/freqplot.py
+++ b/control/freqplot.py
@@ -2469,6 +2469,14 @@ def _determine_omega_vector(syslist, omega_in, omega_limits, omega_num,
     omega_range_given = True
 
     if omega_in is None:
+        for sys in syslist:
+            if isinstance(sys, FrequencyResponseData):
+                # FRD already has predetermined frequencies
+                if omega_in is not None and not np.all(omega_in == sys.omega):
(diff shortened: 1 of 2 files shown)


murrayrm:
This is a good fix, but I think we could add a bit better support if there are multiple FRD systems with different frequencies and not explicit frequency list given.  For example, try this:
```
import control as ct
sys = ct.rss(4, 1, 1)
sys1 = ct.frequency_response(sys, np.logspace(-1, 1, 10))
sys2 = ct.frequency_response(sys, np.logspace(-2, 2, 10))
ct.bode_plot([sys1, sys2])                     # works OK
ct.nyquist_plot(sys1, np.logspace(-1, 1, 10))  # works OK
ct.nyquist_plot(sys2, np.logspace(-1, 1, 10))  # generates error (as it should)
ct.nyquist_plot([sys1, sys2])                  # generates error (does it need to?)
```
It seems to me that the last case should work OK and just plot each system independently, without needing a common frequency range (as it does for Bode plots).

PR #997, opened 2024-04-24
Add extra information to nyquist plots

wueestry:
Adds the possibility to add the unit circle (to get gain crossover frequency),  Ms-circle (sensitivity) and Mt-circle (complementary sensitivity) to nyquist plots similar to the approach taken in [JuliaControl](https://github.com/JuliaControl/ControlSystems.jl/blob/master/lib/ControlSystemsBase/src/plotting.jl)

image

### DIFF
--- a/control/freqplot.py
+++ b/control/freqplot.py
@@ -1060,7 +1060,9 @@ _nyquist_defaults = {
     'nyquist.max_curve_magnitude': 20,          # clip large values
     'nyquist.max_curve_offset': 0.02,           # offset of primary/mirror
     'nyquist.start_marker': 'o',                # marker at start of curve
-    'nyquist.start_marker_size': 4,             # size of the maker
+    'nyquist.start_marker_size': 4,             # size of the marker
+    'nyquist.circle_style':                     # style for unit circles
+      {'color': 'black', 'linestyle': 'dashed', 'linewidth': 1}
 }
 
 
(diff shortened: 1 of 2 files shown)


murrayrm:
It would be nice to have unit tests, even if they call the code to get coverage (better would be to check to make sure the lines are in the right place).

If the system is discrete time, are these still the right circles to draw?

Any interaction with axis scaling?  Eg, if all of the action is very near the origin, might this reset the axes in a way that makes it harder to see the Nyquist plot?  If so, is there a better way to do that (eg, fix the axes and let the circle disappear at the edges of the axes)?

wueestry:
> It would be nice to have unit tests, even if they call the code to get coverage (better would be to check to make sure the lines are in the right place).
> 
> If the system is discrete time, are these still the right circles to draw?
> 
> Any interaction with axis scaling? Eg, if all of the action is very near the origin, might this reset the axes in a way that makes it harder to see the Nyquist plot? If so, is there a better way to do that (eg, fix the axes and let the circle disappear at the edges of the axes)?

This works as well for discrete systems. And tests have now been added.
Not sure yet if fixing the axis is the best idea, as it would result in issues when plotting a second system in the same plot afterwards

control/freqplot.py
```diff
@@ -1527,6 +1527,15 @@ def nyquist_plot(
     return_contour : bool, optional
         If 'True', return the contour used to evaluate the Nyquist plot.
 
+    unit_circle : bool, optional
+        If 'True', displays the unit circle, to read gain crossover frequency.
+
+    mt_circles : array_like, optional
+        draws circles corresponding to the given magnitudes of sensitivity
```
murrayrm, line 1534: For consistency, start phrase with a capital letter and end in a period.

control/freqplot.py
```diff
@@ -1527,6 +1527,15 @@ def nyquist_plot(
...
+    
+    ms_circles : array_like, optional
+        draws circles corresponding to the given magnitudes in complementary sensitivity
```
murrayrm, line 1537: For consistency, start phrase with a capital letter and end in a period.

murrayrm:
I rebased this onto the current main branch and added the ability to set the line style for the circles using `config.defaults`.

PR #999, opened 2024-05-10
klist->gains kwarg for matlab.wrappers.rlocus

### LINKED ISSUE #998: inconsistent kwarg klist/kvect/gains in rlocus doc
The first line of the docstring for `rlocus` suggests that there's a kwarg `klist`,

    """rlocus(sys[, klist, xlim, ylim, ...])
https://github.com/python-control/python-control/blob/e4a03e8419231b040d846ef089babd029e18168c/control/matlab/wrappers.py#L200

 but in the list of parameters it's `kvect`.  

https://github.com/python-control/python-control/blob/e4a03e8419231b040d846ef089babd029e18168c/control/matlab/wrappers.py#L212

They're all just passed on to `root_locus_plot` anyway 

https://github.com/python-control/python-control/blob/e4a03e8419231b040d846ef089babd029e18168c/control/matlab/wrappers.py#L247

and it's expecting `kvect`, or, really, `gains`

https://github.com/python-control/python-control/blob/e4a03e8419231b040d846ef089babd029e18168c/control/rlocus.py#L163

gdmcbain:
This fixes #998.

It wasn't the actual problem there, with `control.matlab.wrappers.rlocus` not being called (rather `control.rlocus.rlocus = control.rlocus.root_locus_plot`), but still this does forward kwargs to that function which is expecting `gains` rather than `klist` or `kvect` (though it does have a `control._process_legacy_keyword` in place to catch the latter).

Feel free to ignore or use this.

### DIFF
--- a/control/matlab/wrappers.py
+++ b/control/matlab/wrappers.py
@@ -197,19 +197,19 @@ def _parse_freqplot_args(*args):
 
 # TODO: rewrite to call root_locus_map, without using legacy plot keyword
 def rlocus(*args, **kwargs):
-    """rlocus(sys[, klist, xlim, ylim, ...])
+    """rlocus(sys[, gains, xlim, ylim, ...])
 
     Root locus diagram.
 
     Calculate the root locus by finding the roots of 1 + k * G(s) where G
     is a linear system with transfer function num(s)/den(s) and each k is
-    an element of kvect.
+    an element of gains.
 
     Parameters
     ----------
     sys : LTI object
         Linear input/output systems (SISO only, for now).
-    kvect : array_like, optional
+    gains : array_like, optional
         Gains to use in computing plot of closed-loop poles.
     xlim : tuple or list, optional
         Set limits of x axis, normally with tuple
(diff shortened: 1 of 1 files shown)


slivingston:
Thanks, @gdmcbain ! While reviewing this, I noticed another `kvect` in the same docstring, so I fixed it, too.

By the way, it looks like the email address you are using in the commit is not associated with your GitHub account because your commit appears without an avatar above. Is that intentional, or did you want to use a different address?

slivingston (approved):

bnavigator:
How about mapping MATLAB's parameter `k` to the `gains` kwarg of `root_locus_plot()`? https://www.mathworks.com/help/control/ref/dynamicsystem.rlocus.html

gdmcbain:
> By the way, it looks like the email address you are using in the commit is not associated with your GitHub account because your commit appears without an avatar above. Is that intentional, or did you want to use a different address?

Oops.  I assume it's too late to fix now; never mind.

> How about mapping MATLAB's parameter `k` to the `gains` kwarg of `root_locus_plot()`? https://www.mathworks.com/help/control/ref/dynamicsystem.rlocus.html

Sure, why not.  Let's.

PR #1001, opened 2024-05-11
Better I/O systems support for phase plots

murrayrm:
This PR provides better I/O systems support for making phase plots:
* Allow I/O systems with no inputs and/or no outputs to be specified.  This is useful when generating phase plots, where you might just have a simple dynamical systems but still want to use the I/O system objects.
* Improve the error messages in `input_output_response` when the number of states, inputs, or outputs are incompatible with the system size by telling you which one didn't match.
* Generate warnings when simulations fail for individual initial conditions and drop individual trace (rather than terminating).  This yields much nicer phase plane plots by allowing the system to generate lines when it can and ignoring the rest.
* Added Jupyter notebook analyzing open and closed loop dynamics of an inverted pendulum, showing off some of the capabilities of the package.
* Changed the way plot titles are created, using axes.set_title (centers title over axes) instead of fig.suptitle (centers over figure, which is good for multi-axes plots but otherwise looks funny).  I'll propagate this same change to time plots and frequency plots in separate PRs.
* Did some tweaking on arrow placement so that very short lines have zero or one arrows.  Could use more work, but this is better than what we had.

As an example of the changes, here is the code to produce a phase portrait for the inverted pendulum example in FBS2e, Figure 5.4:
```
from math import pi
import control as ct
    return [x[1], -b/m * x[1] + (g * l / m) * np.sin(x[0])]
invpend = ct.nlsys(
    invpend_update, states=2, inputs=0, name='invpend',

ct.phase_plane_plot(invpend, [-2*pi - 1, 2*pi + 1, -2, 2], 10)
```

Before:
invpend_before

After:
invpend_after

The changes are not major, but if you look at the number and spacing of arrows along the streamlines, you'll see examples of the changes there (some streamlines have no arrows in "before" plot).  Also note the change in location of the title.


slivingston:
I agree that the plots are better. After a quick review, I did not find any errors, but I need 1 day more to review the changes carefully.

slivingston (changes requested):
## Comments on the ipynb file

* "insighs" -> "insights"

* "which will will" -> "which we will"

* "resuling dynamics" -> "resulting dynamics"

* "reseting" -> "resetting"

* Example 3.3 in the book refers to b (γ there) as "rotational friction," but here you write "viscous friction." The equations are consistent, and Example 3.2 (balance systems) refers to "viscous friction," so either choice works. Nonetheless I am calling it out for consistency with Example 3.3.

## Other comments

* While reviewing this, I noticed functions that return `TimeResponseData` (e.g., [step_response](https://github.com/python-control/python-control/blob/dbc998dedf8e18dcdcc15acf15303f22adf9368f/control/timeresp.py#L1335-L1355)) list some properties of it. Since those docstrings were written, more properties have been added (including 2 in this pull request), so users may not be aware of the new properties. After merging this, I am happy to add a note to relevant docstrings suggesting that users read the documentation of `TimeResponseData` for all properties.

* Thanks for the many code quality improvements (sorting imports, deleting trailing space, ...)

murrayrm:
Check failures are due to NumPy 2.0 and can be ignored (#1013 fixes this).

slivingston (approved):

slivingston:
@murrayrm thanks; I am reviewing #1013 now

PR #1003, opened 2024-05-13
process legacy rlocus keyword k->gains

gdmcbain:
As [requested ](https://github.com/python-control/python-control/pull/999#issuecomment-2105836061) by @bnavigator in review of #999.


bnavigator:
The coverage does not change much because the line is run unconditionally, but could you please add a unit test which passes the keywords in question and checks for their proper handling?

bnavigator:
Thank you! The doctest failure is unrelated.

PR #1004, opened 2024-05-15
add place_varga in API doc

pierre-haessig:
Hello,

This is a tiny PR proposal to add `place_varga` to the [Function reference/Control system synthesis](https://python-control.readthedocs.io/en/0.10.0/control.html#control-system-synthesis) documentation page.

I guess this was forgotten in https://github.com/python-control/python-control/pull/176 when the original `place` was renamed  `place_varga`. Unless it is intentional?

Also, this may fix the non working of link to `place_varga` in [place](https://python-control.readthedocs.io/en/0.10.0/generated/control.place.html#control.place) doc (in See also section: `acker` link is fine, but not `place_varga` ).

PR #1006, opened 2024-05-21
Improve `acker` pole placement doc

pierre-haessig:
Hello,

This is a follow-up of #1004, still on pole placement, but now turning to [`acker` docstring](https://python-control.readthedocs.io/en/0.10.0/generated/control.acker.html#control.acker).

As a first step, this adds a "See Also" section to `place` and `place varga`.

However, I have two more ideas to extend this PR:
- add reference to the Ackermann's formula. I've found § 2.3.5 in the "Linear Control Theory" book by F. W. Fairman, 1998, but I'm not sure it matches perfectly with what's implemented.
- to the best of my (quite limited) knownledge, Ackermann's formula is for *single input* systems (and I got some linear algebra error on a test with a system with 2 inputs). If correct, I suggest to add a `ValueError` if `B.shape[1] > 1`.

### DIFF
--- a/control/statefbk.py
+++ b/control/statefbk.py
@@ -151,7 +151,7 @@ def place_varga(A, B, p, dtime=False, alpha=None):
     K = place_varga(A, B, p, dtime=False, alpha=None)
 
     Required Parameters
-    ----------
+    -------------------
     A : 2D array_like
         Dynamics matrix
     B : 2D array_like
@@ -160,7 +160,7 @@ def place_varga(A, B, p, dtime=False, alpha=None):
         Desired eigenvalue locations
 
     Optional Parameters
-    ---------------
+    -------------------
     dtime : bool
         False for continuous time pole placement or True for discrete time.
         The default is dtime=False.
@@ -193,7 +193,7 @@ def place_varga(A, B, p, dtime=False, alpha=None):
     --------
     >>> A = [[-1, -1], [0, 1]]
     >>> B = [[0], [1]]
-    >>> K = place_varga(A, B, [-2, -5])
+    >>> K = ct.place_varga(A, B, [-2, -5])
 
     See Also
     --------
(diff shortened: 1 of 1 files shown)


pierre-haessig:
Oups, I didn't see PR #1007 and pushed my own commit to fix `place_varga` doctring. Now I have to fix the conflict!

pierre-haessig:
> Now I have to fix the conflict!

Done!

slivingston:
@pierre-haessig Thanks! I am going to squash-merge this because the [second commit](https://github.com/python-control/python-control/pull/1006/commits/4f0e9b7142b605979e84ef093fc3afdddbd3287d) is not used after you resolved the conflict.

slivingston:
Regarding the other two items in your opening post:
1. For widely known results like Ackermann's formula, I recommend to use an open access reference, if available. The book [Feedback Systems](https://fbswiki.org/wiki/index.php/Feedback_Systems:_An_Introduction_for_Scientists_and_Engineers) can be used for this. Ackermann's formula is given in equation (7.21) in Chapter 7.
2. Ackermann's formula is only for single-input systems. This aligns with the similar [acker function in Matlab](http://www.ece.northwestern.edu/local-apps/matlabhelp/toolbox/control/ref/acker.html). The function always raises an exception if the system has more than 1 input, but the error message may be confusing. We could add a check with `ValueError` as you suggest, or the documentation can add a note about single-input systems (or both). Feel free to create another pull request with it.

As an aside, I was not able to find a description of `acker` in the [official Matlab documentation](https://www.mathworks.com/help/control/referencelist.html). Perhaps they removed it?

pierre-haessig:
> As an aside, I was not able to find a description of `acker` in the [official Matlab documentation](https://www.mathworks.com/help/control/referencelist.html). Perhaps they removed it?

Neither did I. I only found an online copy of a unspecified Matlab version here: www.ece.northwestern.edu/local-apps/matlabhelp/toolbox/control/ref/acker.html (in the source code, the date 2002 is mentioned). I don't know when it was removed.

PR #1011, opened 2024-06-16
Frequency plot improvements

murrayrm:
This PR adds a number of improvements for frequency plots (Bode, Nyquist, etc).  The changes are motivated by my experience in using python-control for a course,  where I found a few things that were harder to do than they should have been.

Changes in this PR:

* Allow `label` keyword in frequency response commands to override default label generation.

* Fix bug in processing `indent_radius` keyword when `nyquist_plot` is passed a system (keyword was not getting sent through to `nyquist_response`).

* Restore functionality that allows omega to be specified as a list of 2 elements (indicating a range) in all frequency response/plotting routines.  This used to work for `nyquist` but got removed at some point.  It now works for all frequency response commands.
  - **Note:** this could break some code if someone meant to evaluate the frequency response at exactly two frequencies.  This case seems rare, but it can be fixed by passing the two elements as an array.

* Fix up the `ax` keyword processing to allow arrays or lists + uniform processing in all freqplot routines.

* Fix up `rcParam`s processing (uniformity + remove unnecessary code).  (This needs more work in a future pass to eliminate the rcParams keywords that are part of frequency plotting functions and to pull separate time and frequency plot parameters into a single place.  There are some TODOs in the code marking what needs to be done.)

* Added new suptitle() function to add titles that are better centered (on axes instead of figure). Using this for frequency domain plots for now; will update for time domain plots as well.

* Set up `frd` as factory function with keywords, including setting the signal/system names.  Calls to `ct.FRD` should how be replaces with `ct.frd`, though the former will still work.
    
* Allow Bode and Nyquist plots for FRD systems with different omega vectors (#996) as well as mixtures of FRD and other LTI systems.

* Added a notebook `examples/cds110-bode_nyquist.ipynb` that shows a lot of the new functionality in the context of frequency domain analysis.

* Minor: changed the way that `freq_label` (frequency axis label) is implemented so that it can be more easily overriden (use fstring with '{units}' instead of '%' => can put in custom units).

For the most part these changes just make it easier to do things that were already possible.  Example of some new code that makes use of these features (from `examples/cds110-bode_nyquist.ipynb`):
```
import control as ct
P = ct.tf([1], [1, 0.1, -1])  # normalized inverted pendulum
C = ct.tf([2, 10], [1])       # PD controller
L = P * C                     # Loop transfer function

# Bode and Nyquist plots
plt.figure(figsize=[7, 4])
ax1 = plt.subplot(2, 2, 1)
plt.title("Bode plot for L", size='medium')
[...]

PR #1018, opened 2024-06-29
Time response plot improvements

murrayrm:
This PR makes improvements to the code for plotting time responses, building on some of the improvements made for frequency response plots (#1011):

* Fix up the `ax` keyword processing to allow arrays or lists + uniform processing for all (time and frequency) plot routines.
* Slight refactoring of code to put common functions in ctrlplot.py (additional changes coming in future PR to further consolidate common code)
* Allow time responses for multiple systems with common time vector and inputs.  This enables commands like
   ```
   ct.step_response([sys1, sys2]).plot
   ```
   to find a single time interval (before you had to call `step_response` twice and could get different default time intervals).
* Fix sequential plotting so that different colors are used and plot title is updated (like Bode and Nyquist).
* Allow label keyword in various time response commands to override default label generation.
* Allow legends to be turned on and off using show_legend keyword.
* Added a file `examples/plot_gallery.py` that generates a PDF file with some standard plots to compare changes in basic plotting between PRs and releases.

To get a sense of the differences, consider the following code:
```
sys_mimo1 = ct.tf2ss(
    [[[1], [0.1]], [[0.2], [1]]],
    [[[1, 0.6, 1], [1, 1, 1]], [[1, 0.4, 1], [1, 2, 1]]], name="sys_mimo1")
sys_mimo2 = ct.tf2ss(
    [[[1], [0.1]], [[0.2], [1]]],
    [[[1, 0.2, 1], [1, 24, 22, 5]], [[1, 4, 16, 21], [1, 0.1]]],
    name="sys_mimo2")
ct.step_response(sys_mimo1).plot()
ct.step_response(sys_mimo2).plot()
```

Prior to this PR, the output from this code was

mimo_step-orig

Note the fact that both responses are in the same color.

With this PR, the output becomes

mimo_step-seq

Furthermore, you can compute a single time range using the command
```
ct.step_response([sys_mimo1, sys_mimo2])
```
which gives a single time range for both systems:

mimo_step-list


control/nlsys.py
```diff
@@ -1327,8 +1327,8 @@ def input_output_response(
 
     Parameters
     ----------
-    sys : InputOutputSystem
-        Input/output system to simulate.
+    sysdata : I/O system or list of I/O systems
```
slivingston, line 1330: It is useful to include a class name here because, in that case, docs generation results in a link to documentation for the class. Unless I missed something, this should be `NonlinearIOSystem`. This will work even when the type is written as "...or list of...", e.g., compare with the `data` parameter of `nyquist_plot`: <https://python-control.readthedocs.io/en/0.10.0/generated/control.nyquist_plot.html>

slivingston:
I will finish reviewing this tomorrow morning.

control/nlsys.py
```diff
@@ -1317,7 +1317,7 @@ def nlsys(
 
 
 def input_output_response(
-        sys, T, U=0., X0=0, params=None, ignore_errors=False,
+        sysdata, T, U=0., X0=0, params=None, ignore_errors=False,
```
slivingston, line 1320: Changing the parameter name `sys` to `sysdata` here and in other functions modified by this pull request is good for clarity to users, but it will break existing calls of the form `input_output_response(sys=ss, ...)`. As such, the next version probably should be 0.11.0 instead of 0.10.1 (https://github.com/python-control/python-control/issues/1020) to emphasize that existing code using `control` may need to be updated.

slivingston (commented):
Good improvements! This is ready to merge after you consider/apply my comments.

PR #1022, opened 2024-07-03
Improve markov function, add mimo support, change api to TimeResponseData

KybernetikJo:
This PR adds 
- MIMO support to the markov function and
- it uses the TimeResponseData Type.

```python
T = np.linspace(0, 10, 100)
U = np.ones((1, 100))
response = ct.forced_response(ct.tf([1], [1, 0.5], True), T, U)
H = ct.markov(response, 3)
```

~~This breaks old user code!~~
Does not break old user code!

```python
T = np.linspace(0, 10, 100)
U = np.ones((1, 100))
T, Y = ct.forced_response(ct.tf([1], [1, 0.5], True), T, U)
H = ct.markov(Y, U, 3, transpose=False)
```

### DIFF
--- a/control/modelsimp.py
+++ b/control/modelsimp.py
@@ -402,7 +402,7 @@ def era(YY, m, n, nin, nout, r):
     raise NotImplementedError('This function is not implemented yet.')
 
 
-def markov(*args, **kwargs):
+def markov(*args, m=None, transpose=False, dt=True, truncate=False):
     """markov(Y, U, [, m])
     
     Calculate the first `m` Markov parameters [D CB CAB ...]
@@ -420,12 +420,12 @@ def markov(*args, **kwargs):
     the input data is less than the desired number of Markov parameters (a
     warning message is generated in this case).
 
-    The function can be called with either 1, 2, or 3 arguments:
+    The function can be called with either 1, 2 or 3 arguments:
 
-    * ``K, S, E = lqr(response)``
-    * ``K, S, E = lqr(respnose, m)``
-    * ``K, S, E = lqr(Y, U)``
-    * ``K, S, E = lqr(Y, U, m)``
+    * ``H = markov(response)``
+    * ``H = markov(respnose, m)``
+    * ``H = markov(Y, U)``
+    * ``H = markov(Y, U, m)``
 
     where `response` is an `TimeResponseData` object, and `Y`, `U`, are 1D or 2D
     array and m is an integer.
@@ -446,26 +446,20 @@ def markov(*args, **kwargs):
         Number of Markov parameters to output.  Defaults to len(U).
     dt : True of float, optional
         True indicates discrete time with unspecified sampling time,
-        positive number is discrete time with specified sampling time.
-        It can be used to scale the markov parameters in order to match
-        the impulse response of this library.
-        Default values is True.
+        positive number is discrete time with specified sampling time.It
+        can be used to scale the markov parameters in order to match the
+        impulse response of this library. Default is True.
     truncate : bool, optional
-        Do not use first m equation for least least squares.
-        Default value is False.
+        Do not use first m equation for least least squares. Default is False.
+    transpose : bool, optional
+        Assume that input data is transposed relative to the standard
+        :ref:`time-series-convention`. For TimeResponseData this parameter
+        is ignored. Default is False.
 
     Returns
     -------
     H : ndarray
         First m Markov parameters, [D CB CAB ...]
-    
-
-    Notes
-    -----
-    It works for SISO and MIMO systems.
-
-    This function does comply with the Python Control Library
-    :ref:`time-series-convention` for representation of time series data.
 
     References
     ----------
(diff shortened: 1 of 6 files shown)


murrayrm:
Rather than breaking old code, it would be nice to do this in a way that preserved prior functionality.  So use `*args` processing to allow both of the following to work:
```
H = ct.markov(response, 3)
H = ct.markov(Y, U, 3)
```
It also seems a bit odd (to me) to have the return type be a `TimeResponseData` object.  It true that the Markov parameters are the impulse response, but are people going to be plotting that impulse response as the primary way they interact with the output of the `markov` command?

KybernetikJo:
> Rather than breaking old code, it would be nice to do this in a way that preserved prior functionality. So use `*args` processing to allow both of the following to work:
> 
> ```
> H = ct.markov(response, 3)
> H = ct.markov(Y, U, 3)
> ```
> 

I will update this one.

> It also seems a bit odd (to me) to have the return type be a `TimeResponseData` object. It true that the Markov parameters are the impulse response, but are people going to be plotting that impulse response as the primary way they interact with the output of the `markov` command?

I do not have strong feelings about that.  The idea was to simplify the joint use of markov and era:

```python
impulse_response = ct.markov(response, m=10)
sysd = ct.era(impulse_response, r=4)
```

Direct access of Markov parameters:

```python
_, H = ct.markov(response, m=10)
```

I can change the return value to an array. I think it is also better for the old API.
I guess I should also keep the old API for era.

```python
H = ct.markov(Y, U, m=10)
sysd = ct.era(H, r=4)
```

control/modelsimp.py
```diff
@@ -402,9 +402,11 @@ def era(YY, m, n, nin, nout, r):
...
 
 
-def markov(Y, U, m=None, transpose=False):
-    """Calculate the first `m` Markov parameters [D CB CAB ...]
-    from input `U`, output `Y`.
+def markov(*args, m=None, transpose=False, dt=True, truncate=False):
+    """markov(Y, U, [, m])
+    
+    Calculate the first `m` Markov parameters [D CB CAB ...]
+    from data
```
murrayrm, line 409: Should end in a period.

control/modelsimp.py
```diff
@@ -463,77 +477,80 @@ def markov(Y, U, m=None, transpose=False):
...
 
-    # If data is in transposed format, switch it around
-    if transpose:
-        Umat, Ymat = np.transpose(Umat), np.transpose(Ymat)
-
-    # Make sure the system is a SISO system
-    if Umat.shape[0] != 1 or Ymat.shape[0] != 1:
-        raise ControlMIMONotImplemented
+    # Get the system description
+    if (len(args) < 1):
```
murrayrm, line 482: Parenthesis are not needed here.  The more standard style convention would be
```
if len(args) < 1:
```

control/modelsimp.py
```diff
@@ -463,77 +477,80 @@ def markov(Y, U, m=None, transpose=False):
...
+        Ymat = np.array(args[0].outputs, ndmin=2)
+        transpose = args[0].transpose
+        if args[0].transpose and not args[0].issiso:
+            Umat, Ymat = np.transpose(Umat), np.transpose(Ymat)
+        if (len(args) == 2):
+            m = args[1]
+        elif (len(args) > 2):
+            raise ControlArgument("too many positional arguments")
+    else:
+        if (len(args) < 2):
```
murrayrm, line 496: OK to remove parens (here and below).

control/modelsimp.py
```diff
@@ -427,11 +439,22 @@ def markov(Y, U, m=None, transpose=False):
...
         Input data, arranged in the same way as `Y`.
+    data : TimeResponseData
+        Response data from which the Markov parameters where estimated.
+        Input and output data must be 1D or 2D array.
     m : int, optional
         Number of Markov parameters to output.  Defaults to len(U).
+    dt : True of float, optional
+        True indicates discrete time with unspecified sampling time,
+        positive number is discrete time with specified sampling time.It
+        can be used to scale the markov parameters in order to match the
```
murrayrm, line 450: Markov should be capitalized.

murrayrm, line 450: In the case that a TimeResponseData object is given, should `dt` default to the value coming from the time array?  You could set the default value of `dt` to `None` and then if the input is `Y, U` then you default to `dt=1` and the input is data, you default to `data.time[1] - data.time[0]` (perhaps with a check to make sure that time points are equally spaced?).

murrayrm (changes requested):
Overall this looks OK.  It would be good to add unit tests to cover some of the [lines that are not currently covered](https://coveralls.io/builds/68591443/source?filename=modelsimp.py#L483).

control/modelsimp.py
KybernetikJo: Done.

control/modelsimp.py
KybernetikJo: Done.

control/modelsimp.py
KybernetikJo: Done.

control/modelsimp.py
KybernetikJo: Done.

KybernetikJo: Done.

PR #1023, opened 2024-07-03
Fix impuse response, input data only contains one input[0]==1 in each trace

KybernetikJo:
Computation of impulse response was correct, but input in impulse response was not.

### DIFF
--- a/control/tests/timeresp_test.py
+++ b/control/tests/timeresp_test.py
@@ -536,6 +536,32 @@ class TestTimeresp:
         sysdt = sys.sample(dt, 'impulse')
         np.testing.assert_array_almost_equal(impulse_response(sys, t)[1],
                                              impulse_response(sysdt, t)[1])
+        
+    def test_discrete_time_impulse_input(self):
+        # discrete time impulse input, Only one active input for each trace
+        A = [[.5, 0.25],[.0, .5]]
+        B = [[1., 0,],[0., 1.]]
+        C = [[1., 0.],[0., 1.]]
+        D = [[0., 0.],[0., 0.]]
+        dt = True
+        sysd = ct.ss(A,B,C,D, dt=dt)
+        response = ct.impulse_response(sysd,T=dt*3)
+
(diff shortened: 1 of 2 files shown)


slivingston:
@KybernetikJo Can you add a regression test that demonstrates the error you are fixing?

KybernetikJo:
```python
import numpy as np
import control as ct
print(ct.__version__)

A = [[.5, 0.25],[.0, .5]]
B = [[1., 0,],[0., 1.]]
C = [[1., 0.],[0., 1.]]
D = [[0., 0.],[0., 0.]]
dt = True
sysd = ct.ss(A,B,C,D, dt=dt)
response = ct.impulse_response(sysd,T=dt*3)

U = np.zeros((2,2,4), dtype=float).astype(object)
U[0,0,0] = 1./dt
U[1,1,0] = 1./dt

print(np.array_equal(response.inputs,U))
display(U)
display(response.inputs)
```

Expected / New behavior
```
0.10.1.dev76+g4acc78b.d20240703
True
array([[[1.0, 0.0, 0.0, 0.0],
        [0.0, 0.0, 0.0, 0.0]],

       [[0.0, 0.0, 0.0, 0.0],
        [1.0, 0.0, 0.0, 0.0]]], dtype=object)

array([[[1.0, 0.0, 0.0, 0.0],
        [0.0, 0.0, 0.0, 0.0]],

       [[0.0, 0.0, 0.0, 0.0],
        [1.0, 0.0, 0.0, 0.0]]], dtype=object)
```

Old behavior
```
0.10.0
False
array([[[1.0, 0.0, 0.0, 0.0],
        [0.0, 0.0, 0.0, 0.0]],

       [[0.0, 0.0, 0.0, 0.0],
        [1.0, 0.0, 0.0, 0.0]]], dtype=object)

array([[[1.0, 0.0, 0.0, 0.0],
        [1.0, 0.0, 0.0, 0.0]],

       [[1.0, 0.0, 0.0, 0.0],
        [1.0, 0.0, 0.0, 0.0]]], dtype=object)
```

Issue #1002, opened 2024-05-11
How to convert finite step response model to state space representation?


csianglim:
I have some simulation data for a discrete LTI system which are provided as unit step response models between the 5 inputs (rows) and 12 outputs (columns).

I'd like to convert the model to a state space representation in `python-control` but I'm not sure how to do that.

Based on what I've read so far, it seems like the best way is to convert the step response models to impulse response first, then use `control.era` to get a state space representation. Am I on the right track?

![image](https://github.com/python-control/python-control/assets/7519832/e719ca34-61e6-480d-b647-2d905b0eaa38)

Thanks,
Siang

murrayrm:
This functionality does not yet exist in the python-control toolbox (the `era` function is raises an "Not Implemented" exception).  It would be great if you wanted to contribute this functionality via a pull request!

csianglim:
Thanks, I'll give it a try.

Issue #847, opened 2023-01-12
Sample code for ERA-OKID from UW is broken


RPellowski:
### What is the issue?
This code must have worked at one time but is now broken. Have no previous experience with it running correctly.

Two issues related to:

- http://databookuw.com ... DATA DRIVEN SCIENCE & ENGINEERING ... CODE.zip + DATA.zip (PYTHON CODE)

Notebook containing the broken code:

- ```CH09_SEC03_ERA_OKID.ipynb```

### Steps to Reproduce

1. Test on Python 3.10
2. Run the notebook

### Source changes to accommodate the failures
The following changes to the source were needed:
**Issue 1. Slice assignment**
Change 
    ```yFull[:,:,qi],t = impulse(sysFull,T=tspan,input=qi)```
to
    ```yFull[:,:,qi:qi+1],t = impulse(sysFull,T=tspan,input=qi)```

Change
    ```y1[:,:,qi],t1 = impulse(sysFull,np.arange(200),input=qi)```
    ```y2[:,:,qi],t2 = impulse(sysERA,np.arange(100),input=qi)```
    ```y3[:,:,qi],t3 = impulse(sysERAOKID,np.arange(100),input=qi)```
to
    ```y1[:,:,qi:qi+1],t1 = impulse(sysFull,np.arange(200),input=qi)```
    ```y2[:,:,qi:qi+1],t2 = impulse(sysERA,np.arange(100),input=qi)```
    ```y3[:,:,qi:qi+1],t3 = impulse(sysERAOKID,np.arange(100),input=qi)```

**Issue 2. Invalid np array shape**
Change
    ```yRandom = lsim(sysFull,uRandom,range(200))[0].T # Output```
to
    ```yRandom = lsim(sysFull,uRandom.T,range(200))[0].T # Output```

### Summary
Potential root causes could be due to changes in this library or in changes to Python or both.

murrayrm:
Converting this to a discussion topic so that others can find it in the future.

PR #1024, opened 2024-07-04
Implement ERA, change api to TimeResponseData

KybernetikJo:
This PR implements the ERA method. Related to #1002 and #847

- it uses the TimeResponseData Type.

### DIFF
--- a/control/modelsimp.py
+++ b/control/modelsimp.py
@@ -48,6 +48,7 @@ from .exception import ControlSlycot, ControlMIMONotImplemented, \
 from .iosys import isdtime, isctime
 from .statesp import StateSpace
 from .statefbk import gram
+from .timeresp import TimeResponseData
 
 __all__ = ['hsvd', 'balred', 'modred', 'era', 'markov', 'minreal']
 
@@ -368,8 +369,10 @@ def minreal(sys, tol=None, verbose=True):
     return sysr
 
 
-def era(data, r, m=None, n=None, dt=True):
-    """Calculate an ERA model of order `r` based on the impulse-response data.
+def era(arg, r, m=None, n=None, dt=True, transpose=False):
+    r"""era(YY, r)
+
+    Calculate an ERA model of order `r` based on the impulse-response data.
 
     This function computes a discrete time system
 
@@ -380,8 +383,19 @@ def era(data, r, m=None, n=None, dt=True):
 
     for a given impulse-response data (see [1]_).
 
+    The function can be called with 2 arguments:
+
+    * ``sysd, S = era(data, r)``
+    * ``sysd, S = era(YY, r)``
+
+    where `response` is an `TimeResponseData` object, and `YY` is 1D or 3D
+    array and r is an integer.
+
     Parameters
     ----------
+    YY : array_like
+        impulse-response data from which the StateSpace model is estimated,
+        1D or 3D array.
     data : TimeResponseData
         impulse-response data from which the StateSpace model is estimated.
     r : integer
@@ -398,11 +412,16 @@ def era(data, r, m=None, n=None, dt=True):
         It can be used to scale the StateSpace model in order to match
         the impulse response of this library.
         Default values is True.
+    transpose : bool, optional
+        Assume that input data is transposed relative to the standard
+        :ref:`time-series-convention`. For TimeResponseData this parameter 
+        is ignored.
+        Default value is False.
 
     Returns
     -------
     sys : StateSpace
-        A reduced order model sys=StateSpace(Ar,Br,Cr,Dr,dt)
+        A reduced order model sys=StateSpace(Ar,Br,Cr,Dr,dt).
     S : array
         Singular values of Hankel matrix.
         Can be used to choose a good r value.
@@ -416,6 +435,10 @@ def era(data, r, m=None, n=None, dt=True):
 
     Examples
     --------
+    >>> T = np.linspace(0, 10, 100)
+    >>> _, YY = ct.impulse_response(ct.tf([1], [1, 0.5], True), T)
+    >>> sysd, _ = ct.era(YY, r=1)
+
     >>> T = np.linspace(0, 10, 100)
     >>> response = ct.impulse_response(ct.tf([1], [1, 0.5], True), T)
     >>> sysd, _ = ct.era(response, r=1)
@@ -434,10 +457,16 @@ def era(data, r, m=None, n=None, dt=True):
                 
         return H
     
-    Y = np.array(data.outputs, ndmin=3)
-    if data.transpose:
-        Y = np.transpose(Y)
-    q, p, l = Y.shape
+    if isinstance(arg, TimeResponseData):
+        YY = np.array(arg.outputs, ndmin=3)
+        if arg.transpose:
+            YY = np.transpose(YY)
+    else:
+        YY = np.array(arg, ndmin=3)
+        if transpose:
+            YY = np.transpose(YY)
+    
+    q, p, l = YY.shape
 
     if m is None:
         m = 2*r
@@ -450,7 +479,7 @@ def era(data, r, m=None, n=None, dt=True):
     if (l-1) < m+n:
         raise ValueError("Not enough data for requested number of parameters")
     
-    H = block_hankel_matrix(Y[:,:,1:], m, n+1) # Hankel matrix (q*m, p*(n+1))
+    H = block_hankel_matrix(YY[:,:,1:], m, n+1) # Hankel matrix (q*m, p*(n+1))
     Hf = H[:,:-p] # first p*n columns of H
     Hl = H[:,p:] # last p*n columns of H
     
@@ -463,7 +492,7 @@ def era(data, r, m=None, n=None, dt=True):
     Ar = Sigma_inv @ Ur.T @ Hl @ Vhr.T @ Sigma_inv
     Br = Sigma_inv @ Ur.T @ Hf[:,0:p]*dt
     Cr = Hf[0:q,:] @ Vhr.T @ Sigma_inv
-    Dr = Y[:,:,0]
+    Dr = YY[:,:,0]
 
     return StateSpace(Ar,Br,Cr,Dr,dt), S
 
--- a/control/modelsimp.py
+++ b/control/modelsimp.py
@@ -50,7 +50,7 @@ from .statesp import StateSpace
 from .statefbk import gram
 from .timeresp import TimeResponseData
 
-__all__ = ['hsvd', 'balred', 'modred', 'era', 'markov', 'minreal']
+__all__ = ['hsvd', 'balred', 'modred', 'eigensys_realization', 'markov', 'minreal', 'era']
 
 
 # Hankel Singular Value Decomposition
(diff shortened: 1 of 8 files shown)


control/modelsimp.py
```diff
@@ -368,38 +369,132 @@ def minreal(sys, tol=None, verbose=True):
...
-        Number of columns in Hankel matrix
-    nin: integer
-        Number of input variables
-    nout: integer
-        Number of output variables
-    r: integer
-        Order of model
+    YY : array_like
+        impulse-response data from which the StateSpace model is estimated,
+        1D or 3D array.
```
murrayrm, line 398: impulse → Impulse (start sentence with a capital letter) and try to fit on one line (if less than 75 characters).

control/modelsimp.py
```diff
@@ -368,38 +369,132 @@ def minreal(sys, tol=None, verbose=True):
...
+    data : TimeResponseData
+        impulse-response data from which the StateSpace model is estimated.
```
murrayrm, line 400: impose → Impulse

control/modelsimp.py
```diff
@@ -368,38 +369,132 @@ def minreal(sys, tol=None, verbose=True):
...
+    r : integer
+        Order of model.
+    m : integer, optional
+        Number of rows in Hankel matrix.
+        Default is 2*r.
```
murrayrm, line 405: Put on a single line (here and below).

control/modelsimp.py
```diff
@@ -368,38 +369,132 @@ def minreal(sys, tol=None, verbose=True):
...
+    n : integer, optional
+        Number of columns in Hankel matrix.
+        Default is 2*r.
+    dt : True or float, optional
+        True indicates discrete time with unspecified sampling time,
+        positive number is discrete time with specified sampling time.
+        It can be used to scale the StateSpace model in order to match
+        the impulse response of this library.
+        Default values is True.
```
murrayrm, line 414: Can probably have fewer lines by going out to column 75 before wrapping.  Default value sentence doesn't need to start on a new line.

control/modelsimp.py
```diff
@@ -368,38 +369,132 @@ def minreal(sys, tol=None, verbose=True):
     return sysr
 
 
-def era(YY, m, n, nin, nout, r):
-    """Calculate an ERA model of order `r` based on the impulse-response data
-    `YY`.
+def era(arg, r, m=None, n=None, dt=True, transpose=False):
+    r"""era(YY, r)
```
murrayrm, line 373: Optional: to be consistent with the longer naming style we are using in python-control, I suggest we call the function `eigensys_realization` and then define `era = eigensys_realization` at the bottom of the file (see `freqplot.py` for some example).  We should also (but not here) redefine the other functions in this file (`minimal_realization`, `balanced_realization`, etc).

murrayrm (changes requested):
Overall this looks good.  I found a few small things that you might want to update.

KybernetikJo:
I get a warning

```bash
../python-control/doc/generated/control.era.rst: WARNING: document isn't included in any toctree
```

I could not exclude the alias name era from autosummary. 

__all__=[eigensys_realization, era]

How to exclude era from autosummary?

control/modelsimp.py
```diff
@@ -368,38 +369,128 @@ def minreal(sys, tol=None, verbose=True):
...
+    Hl = H[:,p:] # last p*n columns of H
+    
+    U,S,Vh = np.linalg.svd(Hf, True)
+    Ur =U[:,0:r]
+    Vhr =Vh[0:r,:]
+
+    # balanced realizations
+    Sigma_inv = np.diag(1./np.sqrt(S[0:r]))
+    Ar = Sigma_inv @ Ur.T @ Hl @ Vhr.T @ Sigma_inv
+    Br = Sigma_inv @ Ur.T @ Hf[:,0:p]*dt
```
roryyorke, line 489: It looks like the reference here is "Algorithm 1" of the arxiv paper.

Why does `dt` makes an appearance here?  I don't see any mention of sample rate or sample period in the referenced paper.  It must be needed since you have dt=0.1 in the spring-damper model unit test -- I guess I'm missing something obvious?

roryyorke (commented):
some minor suggestions, and a question.

control/modelsimp.py
```diff
@@ -368,38 +369,128 @@ def minreal(sys, tol=None, verbose=True):
...
+        Order of model.
+    m : integer, optional
+        Number of rows in Hankel matrix. Default is 2*r.
+    n : integer, optional
+        Number of columns in Hankel matrix. Default is 2*r.
+    dt : True or float, optional
+        True indicates discrete time with unspecified sampling time,
+        positive number is discrete time with specified sampling time. It
+        can be used to scale the StateSpace model in order to match the
+        impulse response of this library. Default is True.
```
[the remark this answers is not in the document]
KybernetikJo: This library "python-control" uses a unit area impulse "delta = 1/dt" for dt > 0 and a unit impulse "delta = 1" for dt = True.

The `dt` parameter is used to create a discrete time state space system (A,B,C,D,dt) to match an impulse response generated by control.impulse_response.

control/modelsimp.py
KybernetikJo: ERA in the paper assumes a unit impulse response "delta = 1". There is no dt in the paper.

However, python control generates a unit-area impulse response "delta = 1/dt". See also: #812, #977.

In other words, the impulse responses of StateSpace(A,B,C,D,dt=True) and StateSpace(A,B,C,D,dt=0.1) differ by a scaling.
The dt parameter in era can be used to create the right StateSpace model.

Alternative:
We could also rescale the output data by Y=Y*dt in order to get the unit impulse response instead of the unit-area impulse response. 

Overall, my goal is that impulse_response, era, markov and StateSpace should work well together.

control/modelsimp.py
```diff
@@ -368,38 +369,128 @@ def minreal(sys, tol=None, verbose=True):
...
+
+    for a given impulse-response data (see [1]_).
+
+    The function can be called with 2 arguments:
+
+    * ``sysd, S = eigensys_realization(data, r)``
+    * ``sysd, S = eigensys_realization(YY, r)``
+
+    where `response` is an `TimeResponseData` object, and `YY` is 1D or 3D
+    array and r is an integer.
```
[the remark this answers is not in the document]
KybernetikJo: `response` or `data`? Not sure. 

For me, this function [time_response_plo](https://python-control.readthedocs.io/en/0.10.0/generated/control.time_response_plot.html#control.time_response_plot) which uses `data` for time_response, is my blueprint.

The plotting functions use `data` for input.
[bode_plot](https://python-control.readthedocs.io/en/0.10.0/generated/control.bode_plot.html#control.bode_plot)

The response data types use `response` for input.
[NyquistResponseData](https://python-control.readthedocs.io/en/0.10.0/generated/control.NyquistResponseData.html#control-nyquistresponsedata)

control/modelsimp.py
```diff
@@ -368,38 +369,128 @@ def minreal(sys, tol=None, verbose=True):
...
 
     Returns
     -------
-    sys: StateSpace
-        A reduced order model sys=ss(Ar,Br,Cr,Dr)
+    sys : StateSpace
+        A reduced order model sys=StateSpace(Ar,Br,Cr,Dr,dt).
+    S : array
+        Singular values of Hankel matrix. Can be used to choose a good r
+        value.
```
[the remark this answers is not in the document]
KybernetikJo: I was thinking about that, but I will not add this functionality, at this point in time.

I don't know the literature on this topic well enough, but I know that there are papers that study the choice of r.

control/modelsimp.py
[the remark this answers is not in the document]
KybernetikJo: Done.

control/modelsimp.py
```diff
@@ -368,38 +369,128 @@ def minreal(sys, tol=None, verbose=True):
...
+        Impulse-response from which the StateSpace model is estimated.
+    r : integer
+        Order of model.
+    m : integer, optional
+        Number of rows in Hankel matrix. Default is 2*r.
+    n : integer, optional
+        Number of columns in Hankel matrix. Default is 2*r.
+    dt : True or float, optional
+        True indicates discrete time with unspecified sampling time,
+        positive number is discrete time with specified sampling time. It
```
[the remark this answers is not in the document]
KybernetikJo: Done.

control/modelsimp.py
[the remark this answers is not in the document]
KybernetikJo: Done!

I change the text to  " It can be used to scale the StateSpace model in order to match the unit-area impulse response of python-control. " 

Is there a better way to explain that behavior. Should we add something to Library conventions in order to explain the unit-area impulse response of python-control?

control/modelsimp.py
```diff
@@ -368,38 +369,128 @@ def minreal(sys, tol=None, verbose=True):
...
+    >>> T = np.linspace(0, 10, 100)
+    >>> _, YY = ct.impulse_response(ct.tf([1], [1, 0.5], True), T)
+    >>> sysd, _ = ct.eigensys_realization(YY, r=1)
 
+    >>> T = np.linspace(0, 10, 100)
+    >>> response = ct.impulse_response(ct.tf([1], [1, 0.5], True), T)
+    >>> sysd, _ = ct.eigensys_realization(response, r=1)
     """
-    raise NotImplementedError('This function is not implemented yet.')
+    def block_hankel_matrix(Y, m, n):
```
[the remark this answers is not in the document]
KybernetikJo: Done!

control/modelsimp.py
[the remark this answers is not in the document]
KybernetikJo: Postponed. Possible improvement in the future.

control/modelsimp.py
KybernetikJo: Done. I added a little comment to indicate the intentional use of Br*dt.

Is there a more elegant way to achieve this?

control/modelsimp.py
KybernetikJo: Done!

control/modelsimp.py
KybernetikJo: Done! See #1028

control/modelsimp.py
KybernetikJo: Done.

control/modelsimp.py
KybernetikJo: Done.

control/modelsimp.py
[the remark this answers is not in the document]
KybernetikJo: OK, in my PR #1022 I initially used `response` instead of `data`.

@roryyorke @murrayrm what should we use?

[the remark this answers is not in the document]
roryyorke: I prefer `response` as it's more specific, but the main thing is to be consistent within the ERA docstring(s).

control/modelsimp.py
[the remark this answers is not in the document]
roryyorke: Ah, thanks, now I understand.  

One possibly-not-obvious side-effect of using `True` is that it has a value of 1, so one gets debatably natural discrete-time behaviour in that case.

The unit-area convention is documented in `timeresp.impulse_resp` under `Notes`, I suggest a reference to that.

roryyorke (approved):
LGTM.  If you think it's useful, add a reference to `timeresp.impulse_response`.

PR #1027, opened 2024-07-11
Add FrequencyResponseList to explicit exports __all__

KybernetikJo:
Is used in sphinx doc, but not exported to module __all__.


KybernetikJo:
Is addressed by #1034.

PR #1036, opened 2024-08-08
Update OS/BLAS test matrix workflow

murrayrm:
This PR updates the OS/BLAS test matrix workflow that is used to test releases against various combinations of operating systems and BLAS implementations.  Changes:

* Found and fixed an issue in `nyquist_test.py` in which some calls with a manually specified frequency response would generate a user warning and others wouldn't.  Both cases are not handled properly (and a redundant check has been removed).
* Fixed some issues with paths for OpenBLAS on macos that were causing compiler errors.
* Updated pip-based tests to check out current branch instead of main (mainly useful when using a development branch, like this one).

No changes to the actual control toolbox itself.

### DIFF
--- a/control/tests/nyquist_test.py
+++ b/control/tests/nyquist_test.py
@@ -132,10 +132,17 @@ def test_nyquist_basic():
     # Nyquist plot with poles on imaginary axis, omega specified
     # (can miss encirclements due to the imaginary poles at +/- 1j)
     sys = ct.tf([1], [1, 3, 2]) * ct.tf([1], [1, 0, 1])
-    with pytest.warns(UserWarning, match="does not match") as records:
+    with warnings.catch_warnings(record=True) as records:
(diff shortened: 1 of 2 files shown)


control/tests/nyquist_test.py
```diff
@@ -132,18 +132,17 @@ def test_nyquist_basic():
...
-        assert _Z(sys) == count + _P(sys)
+        if len(records) == 0:
+            # No warnings (it happens) => make sure count is correct
+            assert _Z(sys) == count + _P(sys)
+        elif len(records) == 1:
+            # Expected case: make sure warning is the right one
+            assert issubclass(records[0].category, UserWarning)
+            assert "encirclements does not match" in str(records[0].message)
+        else:
+            pytest.fails("multiple warnings in nyquist_response (?)")
```
slivingston, line 145:
```suggestion
            pytest.fail("multiple warnings in nyquist_response (?)")
```
https://docs.pytest.org/en/stable/reference/reference.html#pytest-fail

PR #1038, opened 2024-08-09
Documentation updates and testing

murrayrm:
This PR provide updated docstrings for a large number of functions that had undocumented arguments along with a unit that that checks to make sure that positional arguments and keyword arguments are appropriately documented.  The unit test checks for the following:

* All named arguments should be documented in the docstring using standard Python docstring syntax (parameter name followed by a space, then a color, then another space).
* Detectable keyword arguments access via **kwargs are also checked to make sure they are documented.
* Exceptions are allowed to skip testing for entire functions or specific arguments in functions.
* Functions with unnamed positional arguments (*args) must be checked manually and an MD5 hash on the source code is used to make sure that changes will require reconfirmation (by updated the stored hash).
* Deprecated functions must contain a Sphynx deprecation message and issue a `FutureWarning`.

Implementing this unit test picked up 100+ undocumented variables that are now documented, as well as triggering a functions changes in argument processing and argument hiding (for internal arguments that the user should never call).

This PR contains no changes in functionality, just docstring updates and small code changes for consistency (eg, `DeprecationWarning` [which is ignored/hidden by default] → `FutureWarning`).


slivingston (commented):
I plan to finish review tomorrow.

slivingston (commented):
Excellent work on the docstrings checker. It seems useful enough to be offered to the [numpydoc repository](https://github.com/numpy/numpydoc) or made into a [ruff rule](https://docs.astral.sh/ruff/rules/).

I added several more small comments. After considering those, this is ready to merge.

PR #1041, opened 2024-08-15
update macos gfortran version

murrayrm:
This PR updates the version of the FORTRAN compiler that is used in the OS/BLAS test matrix.  Apparently gfortran11 is no longer available on the Github Actions macos runner.


slivingston:
For posterity, this was announced at <https://github.com/actions/runner-images/issues/10213> and is now indicated in the [upstream image readme](https://github.com/actions/runner-images/blob/4e6e715037c7824b785bfb8eb09f1eb59bdbf936/images/macos/macos-14-Readme.md).

PR #1051, opened 2024-10-17
Improve code style, markov

KybernetikJo:
Improve code style, markov:

- Add space after commas


KybernetikJo:
@murrayrm
I don't understand why a new hash was necessary. The docstring has not changed!
I have updated the hash for markov.

murrayrm:
Because this function uses `*args`, we use a hash of the docstring and the function text to ensure that the developer confirms that arguments are still documented properly (by updating the hash).

PR #1054, opened 2024-11-09
Update find_eqpt to find_operating_point, adding root_method + better docs

murrayrm:
This PR is motivated by discussion #1053 in which a situation the current implementation of `find_eqpt` does not find a good solution due to the problem being over/underconstrained.  The main functional change is this PR is to help address that by adding a `root_method` method that allows a least squares solution to be obtained.

In addition, while looking through this (fairly old) code, it seemed worth making some additional changes:

* Renamed `find_eqpt` to `find_operating_point` to be more consistent with python-control naming conventions (and the fact that an operating point is not always an equilibrium point.
* Changed the return type for `find_operating_point` from a tuple of various lengths (depending on parameter settings) to a `OperatingPoint` object, which contains the states, inputs, and outputs corresponding to the operating point, along with the result of the call to the `scipy.minimize.root` function.
* Added `root_method` and `root_kwargs` parameters to allow customization of the root finding method.
* Maintained backward compatibility so that `find_eqpt` still works, as well as `return_y` (now `return_outputs`, with legacy processing of `return_y`) and `return_result`.  In addition, the return object can be accessed as a tuple for consistency with the previous return signature.
* Updated docstrings and documentation to use the new conventions.
* Added unit tests to check all new functionality.

In the context of #1053, a solution to the issue there can be obtained by calling `find_eqpt` (now `find_operating_point`) with `root_method='lm'`.

### DIFF
--- a/control/nlsys.py
+++ b/control/nlsys.py
@@ -93,7 +93,7 @@ class NonlinearIOSystem(InputOutputSystem):
         generic name <sys[id]> is generated with a unique integer id.
 
     params : dict, optional
-       Parameter values for the system.  Passed to the evaluation functions
+        Parameter values for the system.  Passed to the evaluation functions
         for the system as default values, overriding internal defaults.
 
     See Also
@@ -1671,7 +1671,7 @@ def input_output_response(
         success=soln.success, message=message)
 
 
-class OperatingPoint(object):
+class OperatingPoint():
     """A class for representing the operating point of a nonlinear I/O system.
 
     The ``OperatingPoint`` class stores the operating point of a nonlinear
@@ -1693,18 +1693,18 @@ class OperatingPoint(object):
 
     """
     def __init__(
-            self, states, inputs=None, outputs=None, result=None,
+            self, states, inputs, outputs=None, result=None,
             return_outputs=False, return_result=False):
         self.states = states
         self.inputs = inputs
 
         if outputs is None and return_outputs and not return_result:
-            raise SystemError("return_outputs specified by no y0 value")
+            raise ValueError("return_outputs specified but no outputs value")
         self.outputs = outputs
         self.return_outputs = return_outputs
 
         if result is None and return_result:
-            raise SystemError("return_result specified by no result value")
+            raise ValueError("return_result specified but no result value")
         self.result = result
         self.return_result = return_result
 
@@ -1799,7 +1799,7 @@ def find_operating_point(
         values will be ignored in solving for an operating point.  State
         indices can be listed in any order.  By default, all updates will be
         fixed at `dx0` in searching for an operating point.
-    root_method : str, optonal
+    root_method : str, optional
         Method to find the operating point.  If specified, this parameter
         is passed to the :func:`scipy.optimize.root` function.
     root_kwargs : dict, optional
--- a/control/nlsys.py
+++ b/control/nlsys.py
@@ -1694,14 +1694,14 @@ class OperatingPoint(object):
     """
     def __init__(
             self, states, inputs=None, outputs=None, result=None,
-            return_y=False, return_result=False):
+            return_outputs=False, return_result=False):
         self.states = states
         self.inputs = inputs
 
-        if outputs is None and return_y and not return_result:
-            raise SystemError("return_y specified by no y0 value")
+        if outputs is None and return_outputs and not return_result:
+            raise SystemError("return_outputs specified by no y0 value")
         self.outputs = outputs
-        self.return_y = return_y
+        self.return_outputs = return_outputs
 
         if result is None and return_result:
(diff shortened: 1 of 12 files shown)


control/nlsys.py
```diff
@@ -1663,87 +1671,190 @@ def ivp_rhs(t, x):
...
+
+    """
+    def __init__(
+            self, states, inputs=None, outputs=None, result=None,
+            return_outputs=False, return_result=False):
+        self.states = states
+        self.inputs = inputs
+
+        if outputs is None and return_outputs and not return_result:
+            raise SystemError("return_outputs specified by no y0 value")
```
slivingston, line 1702:
```suggestion
            raise SystemError("return_outputs specified but no y0 value")
```

control/nlsys.py
```diff
@@ -1663,87 +1671,190 @@ def ivp_rhs(t, x):
...
+        self.outputs = outputs
+        self.return_outputs = return_outputs
+
+        if result is None and return_result:
+            raise SystemError("return_result specified by no result value")
```
slivingston, line 1707:
```suggestion
            raise SystemError("return_result specified but no result value")
```

control/nlsys.py
slivingston: Here and below are both non-standard uses of [SystemError, which normally is for "when the [Python] interpreter finds an internal error"](https://docs.python.org/3/library/exceptions.html#SystemError). I recommend [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError) instead, which is also used elsewhere in this module for inconsistent arguments.

examples/cruise-control.py
```diff
@@ -347,7 +347,7 @@ def cruise_plot(sys, t, y, label=None, t_hill=None, vref=20, antiwindup=False,
 
 # Compute the equilibrium throttle setting for the desired speed (solve for x
 # and u given the gear, slope, and desired output velocity)
-X0, U0, Y0 = ct.find_eqpt(
+X0, U0, Y0 = ct.find_operating_point(
     cruise_pi, [vref[0], 0], [vref[0], gear[0], theta0[0]],
     y0=[0, vref[0]], iu=[1, 2], iy=[1], return_y=True)
```
slivingston, line 352:
```suggestion
    y0=[0, vref[0]], iu=[1, 2], iy=[1], return_outputs=True)
```
Update the argument to avoid a `FutureWarning` being raised in the example.

slivingston: Similarly, `return_y=True` in [examples/cruise.ipynb](https://github.com/python-control/python-control/blob/cc0020a990bcb96ade16e338f5b152884fa4fe16/examples/cruise.ipynb) should be updated, too.

slivingston (commented):
Good improvements! I made several comments to consider before merging.

I ran [the example that originally motivated this pull request](https://github.com/python-control/python-control/discussions/1053#discussioncomment-11136440), and I confirmed the result is as expected after adding the new argument `root_method='lm'`.

Note that other files in examples/ besides cruise-control.py use `find_eqpt()`. However, it is now an alias for `find_operating_point()` without any indication of being deprecated, so OK to leave the other examples for now.

control/nlsys.py
```diff
@@ -1663,87 +1671,190 @@ def ivp_rhs(t, x):
...
+
+    # Implement (thin) len to emulate legacy return value
+    def __len__(self):
+        return len(list(self.__iter__()))
+
+
+def find_operating_point(
+        sys, x0, u0=None, y0=None, t=0, params=None, iu=None, iy=None,
+        ix=None, idx=None, dx0=None, root_method=None, root_kwargs=None,
+        return_outputs=None, return_result=None, **kwargs):
```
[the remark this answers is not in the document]
murrayrm: This needs to be `None` in order for the legacy keyword handling to work: if both `return_y` and `return_outputs` are specified, `_process_legacy_keyword` will raise an exception. Thus with this change, setting the default value of `return_outputs` to False would mean that if you call `find_operating_point` with `return_y`, you get  an exception instead of a warning.

control/nlsys.py
```diff
@@ -1663,87 +1671,190 @@ def ivp_rhs(t, x):
         success=soln.success, message=message)
 
 
-def find_eqpt(sys, x0, u0=None, y0=None, t=0, params=None,
-              iu=None, iy=None, ix=None, idx=None, dx0=None,
-              return_y=False, return_result=False):
-    """Find the equilibrium point for an input/output system.
+class OperatingPoint(object):
```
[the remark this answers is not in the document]
murrayrm: Since we are using Python 3 and that only supports new classes, I think we should move to not inheriting from `object` (which is how things are done in the official documentation).

PR #1066, opened 2024-11-23
fix phase plot lines output

murrayrm:
The `phase_plane_plot` return value was returning a sublist of lines for equilibrium points and streamlines, so the `cplt.lines` output looked like
```
[line1, line2, ..., lineN, [sep1], [sep2] ...]
```
This PR gets rid of the inner list, so you just get a list of all of the lines (or markers) on the plot, matching the documentation.


slivingston (approved):
Good catch! I did not find any code in the repository affected by this, which perhaps is why it went unnoticed for a while.

PR #1070, opened 2024-12-02
sort loci in PoleZeroData

murrayrm:
This small PR adds the option to sort loci when creating a PoleZeroData object.  This is useful if you create a PoleZeroData object in some way other than calling `root_locus_map` (which sorts the data ahead of time).

The need for this functionality came up when working through Example 5.17 of FBS2e (bicycle eigenvalue location versus speed), as shown in the following code:

https://github.com/murrayrm/fbs2e-python/blob/main/example-5.17-bicycle_stability.py

(At some point I would like to generalized the root locus capability to allow plotting an eigenvalue locus diagram against any free parameter in a nonlinear system, but that will take more time and effort.)


control/pzmap.py
```diff
@@ -79,11 +79,13 @@ class PoleZeroData:
         System name.
     sys : StateSpace or TransferFunction
         System corresponding to the data.
+    sort_loci : bool, optional)
```
slivingston, line 82:
```suggestion
    sort_loci : bool, optional
```

slivingston (approved):
There is a small misprint in the docstring. Besides that, this is ready to merge.

Not related to code: the PR description refers to Example 6.8 but links to [Example 5.17](https://github.com/murrayrm/fbs2e-python/blob/e3e4045f9a237290cd721ec11708bb5a47a3f762/example-5.17-bicycle_stability.py). This PR is clearly relevant to Example 5.17, so I guess "Example 6.8" is a misprint or perhaps example numbers changed in the book, but I want to bring attention to it in case it indicates that more actions are necessary.

PR #1074, opened 2024-12-04
update model_reduction to allow input/output selection + unstable warning (vs exception)

murrayrm:
This PR updates the `model_reduction` function to add some additional functionality.  From the docstring:

> This function produces a reduced-order model of a system by eliminating specified inputs, outputs, and/or states from the original system.  The specific states, inputs, or outputs that are eliminated can be specified by either listing the states, inputs, or outputs to be eliminated or those to be kept.

Specific updates:
* Changed the `ELIM` positional parameter to `elim_states` (positional or keyword).
* Added `keep_states` to allow specifying the list of states to keep instead of those to eliminate.
* Added `elim_inputs`, `elim_outputs`, `keep_inputs`, `keep_outputs` to allow selection of inputs and outputs.
* Specifications of states, inputs, and outputs can be done by a list of integers (original implementation) as well as slices or lists of signal names (new).
* In the case of eliminating states, inputs, or outputs from an unstable system, a warning is issued (versus an exception) and this warning can be suppressed using `warn_unstable=False`.
* Removed the copyright info from the top of `modelsimp.py` since we know just use the `LICENSE` file.
* Add unit tests and documentation updates.
* Cleaned up import order (isort -m2) and removed extraneous spaces.

This PR was motivated by an example from FBS2e (see [steering.py](https://github.com/murrayrm/fbs2e-python/blob/main/steering.py)), where I wanted to extract the dynamics of a subsystem (and the original system had an eigenvalue at the origin).  It is possible to accomplish the same result by using the old implementation (assuming the exception is changed to a warning) and then indexing the system directly to handle the inputs and outputs, but it seemed natural to include in a single function call.

### DIFF
--- a/control/modelsimp.py
+++ b/control/modelsimp.py
@@ -146,6 +146,9 @@ def model_reduction(
     other checking is done, so users to be careful not to render a system
     unobservable or unreachable.
 
+    States, inputs, and outputs can be specified using integer offers.
+    Slices can also be specified, but must use the Python ``slice()`` function.
+
     """
     if not isinstance(sys, StateSpace):
         raise TypeError("system must be a a StateSpace system")
@@ -158,8 +161,16 @@ def model_reduction(
 
     # Utility function to process keep/elim keywords
     def _process_elim_or_keep(elim, keep, labels, allow_both=False):
-        elim = [] if elim is None else np.atleast_1d(elim)
-        keep = [] if keep is None else np.atleast_1d(keep)
+        def _expand_key(key):
+            if key is None:
+                return []
+            elif isinstance(key, slice):
+                return range(len(labels))[key]
+            else:
+                return np.atleast_1d(key)
+
+        elim = _expand_key(elim)
+        keep = _expand_key(keep)
             
         if len(elim) > 0 and len(keep) > 0:
             if not allow_both:
--- a/control/modelsimp.py
+++ b/control/modelsimp.py
@@ -108,7 +108,7 @@ def hankel_singular_values(sys):
     return hsv[::-1]
 
 
-def model_reduction(sys, ELIM, method='matchdc'):
+def model_reduction(sys, ELIM, method='matchdc', warn_unstable=True):
     """Model reduction by state elimination.
     
     Model reduction of `sys` by eliminating the states in `ELIM` using a given
@@ -122,7 +122,9 @@ def model_reduction(sys, ELIM, method='matchdc'):
         Vector of states to eliminate.
     method : string
         Method of removing states in `ELIM`: either 'truncate' or
-        'matchdc'.
+        'matchdc' (default).
+    warn_unstable: bool, option
+        If `False`, don't warn if system is unstable.
 
     Returns
     -------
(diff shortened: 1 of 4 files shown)


control/modelsimp.py
```diff
@@ -108,89 +76,148 @@ def hankel_singular_values(sys):
...
-    >>> Gr = ct.modred(G, [0, 2], method='matchdc')
+    >>> Gr = ct.model_reduction(G, [0, 2], method='matchdc')
     >>> Gr.nstates
     2
 
-    """
+    See Also
+    --------
+    balanced_reduction : Eliminate states using Hankel singular values.
+    minimal_realization : Eliminate unreachable or unobseravble states.
```
slivingston, line 139:
```suggestion
    minimal_realization : Eliminate unreachable or unobservable states.
```

control/modelsimp.py
```diff
@@ -108,89 +76,148 @@ def hankel_singular_values(sys):
...
-
-    if method == 'matchdc':
-        # if matchdc, residualize
+    if warn_unstable:
+        if isctime(sys) and np.any(np.linalg.eigvals(sys.A).real >= 0.0) or \
+           isdtime(sys) and np.any(np.abs(np.linalg.eigvals(sys.A)) >= 1):
+            warnings.warn("System is unstable; reduction may be meaningless")
+
+    # Utility function to process keep/elim keywords
+    def _process_elim_or_keep(elim, keep, labels, allow_both=False):
```
slivingston, line 164: Any reason to keep `allow_both`? _process_elim_or_keep() is a function defined only in the scope of model_reduction(), and all calls to it have the default `allow_both` value (`allow_both=False`).

control/modelsimp.py
```diff
@@ -108,89 +76,148 @@ def hankel_singular_values(sys):
...
+    elim_inputs, elim_outputs, elim_states : array of int or str, optional
+        Vector of inputs, outputs, or states to eliminate.  Can be specified
+        either as an offset into the appropriate vector or as a signal name.
+    keep_inputs, keep_outputs, keep_states : array, optional
+        Vector of inputs, outputs, or states to keep.  Can be specified
+        either as an offset into the appropriate vector or as a signal name.
     method : string
-        Method of removing states in `ELIM`: either 'truncate' or
-        'matchdc'.
+        Method of removing states in `elim`: either 'truncate' or
```
slivingston, line 107:
```suggestion
        Method of removing inputs, outputs, and/or states: either 'truncate' or
```
This is confusing because there is no `elim` parameter and because inputs and outputs can be removed as well, in addition to states.

slivingston (commented):
Good improvement! I made several stylistic and misprint-related comments. Else, ready to merge.

control/modelsimp.py
```diff
@@ -108,89 +76,148 @@ def hankel_singular_values(sys):
...
-        dico = 'C'
-    else:
-        raise NotImplementedError("Function not implemented in discrete time")
+    States, inputs, and outputs can be specified using integer offers or
+    using signal names.  Slices can also be specified, but must use the
+    Python ``slice()`` function.
+
+    """
+    if not isinstance(sys, StateSpace):
+        raise TypeError("system must be a a StateSpace system")
```
slivingston, line 155:
```suggestion
        raise TypeError("system must be a StateSpace system")
```

control/modelsimp.py
murrayrm: Will update, but this only applies to the way states are removed.

slivingston (approved):

Issue #1075, opened 2024-12-05
Multiplying a MIMO `TransferFunction` by `1.0` multiplies by a matrix of ones, not identity


sdahdah:
Multiplying a MIMO `TransferFunction` object by the Python float `1.0` appears to multiply it by a matrix of ones, rather than the identity matrix. The same goes for any scalar. Is this the intended behaviour? I checked in the documentation and didn't see anything.

Example:
```python
import control

G = control.TransferFunction(
    [
        [[87.8], [-86.4]],
        [[108.2], [-109.6]],
    ],
    [
        [[1, 1], [1, 1]],
        [[1, 1], [1, 1]],
    ],
)

I = control.TransferFunction(
    [
        [[1], [1]],
        [[1], [1]],
    ],
    [
        [[1], [1]],
        [[1], [1]],
    ],
)

print(G)
print(1 * G)
print(I * G)
```

PR #1078, opened 2024-12-07
fix issue with multiplying MIMO LTI system by scalar

murrayrm:
This PR addresses #1075 and also fixes a small bug with constant FRD systems with small numbers of frequency points.

Changes:
* Multiplying (*), adding (+), or subtracting (-) a constant from any (MIMO) LTI object now acts element-wise (same as ndarray's).  This fixes a bug where multiplying a MIMO LTI system by a constant was multiplying by a matrix filled with the constant rather than a diagonal matrix (scaled identity).
* Specifying an FRD system with fewer than 4 frequency points was generating an error because the default settings try to set up a smooth (interpolating) response and the default degree of the fit was 3.  The SciPy interpolation function requires k+1 data points for a degree k fit, so an exception was generated. 
* Added unit tests that catch the bug in #1075 but also check across all LTI system representations (frd, ss, tf).

### DIFF
--- a/control/xferfcn.py
+++ b/control/xferfcn.py
@@ -634,11 +634,11 @@ class TransferFunction(LTI):
         from .statesp import StateSpace
 
         # Convert the second argument to a transfer function.
-        if isinstance(other, StateSpace):
+        if isinstance(other, (StateSpace, np.ndarray)):
             other = _convert_to_transfer_function(other)
-        elif isinstance(other, (int, float, complex, np.number, np.ndarray)):
-            other = _convert_to_transfer_function(other, inputs=self.ninputs,
-                                                  outputs=self.noutputs)
+        elif isinstance(other, (int, float, complex, np.number)):
+            # Multiply by a scaled identify matrix (transfer function)
+            other = _convert_to_transfer_function(np.eye(self.ninputs) * other)
         if not isinstance(other, TransferFunction):
             return NotImplemented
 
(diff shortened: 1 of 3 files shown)


control/xferfcn.py
```diff
@@ -681,8 +681,8 @@ def __rmul__(self, other):
 
         # Convert the second argument to a transfer function.
         if isinstance(other, (int, float, complex, np.number)):
-            other = _convert_to_transfer_function(other, inputs=self.ninputs,
-                                                  outputs=self.ninputs)
+            # Multiply by a scaled identify matrix (transfer function)
```
slivingston, line 684:
```suggestion
            # Multiply by a scaled identity matrix (transfer function)
```

control/xferfcn.py
```diff
@@ -723,9 +723,8 @@ def __truediv__(self, other):
         """Divide two LTI objects."""
 
         if isinstance(other, (int, float, complex, np.number)):
-            other = _convert_to_transfer_function(
-                other, inputs=self.ninputs,
-                outputs=self.ninputs)
+            # Multiply by a scaled identify matrix (transfer function)
```
slivingston, line 726:
```suggestion
            # Multiply by a scaled identity matrix (transfer function)
```

control/xferfcn.py
```diff
@@ -634,11 +634,11 @@ def __mul__(self, other):
...
 
         # Convert the second argument to a transfer function.
-        if isinstance(other, StateSpace):
+        if isinstance(other, (StateSpace, np.ndarray)):
             other = _convert_to_transfer_function(other)
-        elif isinstance(other, (int, float, complex, np.number, np.ndarray)):
-            other = _convert_to_transfer_function(other, inputs=self.ninputs,
-                                                  outputs=self.noutputs)
+        elif isinstance(other, (int, float, complex, np.number)):
+            # Multiply by a scaled identify matrix (transfer function)
```
slivingston, line 640:
```suggestion
            # Multiply by a scaled identity matrix (transfer function)
```

PR #1079, opened 2024-12-15
update GitHub Actions to use miniforge v3 (fixes mambaforge deprecation)

murrayrm:
Mambaforge support in Miniconda is being deprecated since mamba is now included directly in miniforge, as described [here](https://conda-forge.org/news/2024/07/29/sunsetting-mambaforge/).  To "encourage users to switch to Miniforge" a set of "brownouts" has been scheduled at increasing frequency, and one of them happened today, causing GitHub Actions workflows to stop working.

This PR updates the GitHub Actions workflows to use miniconda@v3, which fixes the issue.

Note: there are still some failures in the OS/BLAS test matrix workflow for Intel-based MacOS ([example](https://github.com/murrayrm/python-control/actions/runs/12333951479/job/34423656180)).  I'm not sure these are directly related to miniforge/mamba since they appear to affect only certain configurations in the test matrix.  I'll post a separate issue on that one.

### DIFF
--- a/.github/workflows/doctest.yml
+++ b/.github/workflows/doctest.yml
@@ -13,13 +13,12 @@ jobs:
       uses: actions/checkout@v3
 
     - name: Setup Conda
-      uses: conda-incubator/setup-miniconda@v2
+      uses: conda-incubator/setup-miniconda@v3
       with:
         python-version: 3.12
         activate-environment: doctest-env
         environment-file: .github/conda-env/doctest-env.yml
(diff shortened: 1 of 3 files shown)


slivingston (approved):
I agree that the test failure is likely not related to this PR, but I am investigating it. Feel free to merge since there is an issue (https://github.com/python-control/python-control/issues/1080) to remind us to debug it.

PR #1081, opened 2024-12-19
Support for binary operations between MIMO and SISO LTI systems

### LINKED ISSUE #1076: Proposal for multiplying a NumPy array by a scalar `TransferFunction`
Currently, multiplying a SISO `TransferFunction` object by a NumPy array results in a dimension error:

```python
import control
import numpy as np

tf = control.TransferFunction([1], [1, 0])
arr = np.array([
    [1, 0],
    [2, 1],
])

print(tf * arr)
print(arr * tf)
```

```
ValueError: C = A * B: A has 1 column(s) (input(s)), but B has 2 row(s) (output(s)).
```

I think it would be convenient if output was a MIMO `TransferFunction` object of the array's dimension. So, for example,

```python
G = control.TransferFunction(
    [
        [[1], [0]],
        [[2], [1]],
    ],
    [
        [[1, 0], [1, 0]],
        [[1, 0], [1, 0]],
    ],
)
````

I would be happy to implement this and do a PR, but I want to make sure it's something the maintainers actually want. I would think all SISO systems (state-space, etc) should behave in the same way.

This appears to have been discussed in #459 , but I'm not sure what the conclusion was.

### PR DESCRIPTION
**Pull request by sdahdah**
Resolves #1076 and #939 

This PR implements binary operations (`__add__`, `__mul__`, etc.) between MIMO and SISO LTI systems.

Specifically, it

- implements `TransferFunction.append()` and `FrequencyResponseData.append()`,
- changes `control.append()` to return the type of the first argument,
- implements MIMO-SISO `__mul__`, `__rmul__`, `__truediv__` and `__rtruediv__` for MIMO and SISO LTI systems using the updated `control.append()`,
- implements MIMO-SISO `__add__`, `__radd__`, `__sub__`, and `__rsub__` using the updated multiplication and division operations,
- implements inversion of biproper state-space systems via `__pow__`, and
- fixes a bug (?) where `StateSpace.minreal()` drops the `dt`

I would recommend reviewing the changes in that order.

See also #459

### CODE REVIEW

**Comment by murrayrm** [comment]
I'll try to review this in the coming days.  In the meantime, there is a typo in a docstring ("ct.ombine_tf" should be "ct.combine_tf") that is causing a failure in the doctest check.

**Comment by sdahdah** [comment]
> I'll try to review this in the coming days. In the meantime, there is a typo in a docstring ("ct.ombine_tf" should be "ct.combine_tf") that is causing a failure in the doctest check.

Thank you! Is there a way to run the doctests locally? `make html` in the `doc/` folder I think? I have not been able to figure that out

**Comment by murrayrm** [comment]
> Thank you! Is there a way to run the doctests locally? `make html` in the `doc/` folder I think? I have not been able to figure that out

make doctest in the doc/ directory will run the tests.

File: control/bdalg.py [hunk 1]
```diff
@@ -356,14 +356,14 @@ def feedback(sys1, sys2=1, sign=-1, **kwargs):
...
-    Group LTI state space models by appending their inputs and outputs.
+    Group LTI models by appending their inputs and outputs.
 
     Forms an augmented system model, and appends the inputs and
     outputs together.
 
     Parameters
     ----------
-    sys1, sys2, ..., sysn: scalar, array, or :class:`StateSpace`
+    sys1, sys2, ..., sysn: scalar, array, or :class:`LTI`
```
murrayrm, line 366: Standard Numpydoc style is to add a space on each side of the colon, so this should be
```
sys1, sys2, ..., sysn : ...
```

control/bdalg.py
```diff
@@ -382,9 +382,10 @@ def append(*sys, **kwargs):
 
     Returns
     -------
-    out: :class:`StateSpace`
+    out: :class:`LTI`
```
murrayrm, line 385: Add space after colon following 'out'.

control/frdata.py
```diff
@@ -753,6 +768,35 @@ def feedback(self, other=1, sign=-1):
...
+        # TODO: handle omega re-mapping
+
+        new_fresp = np.zeros(
+            (
+                self.noutputs + other.noutputs,
+                self.ninputs + other.ninputs,
+                self.omega.shape[-1],
+            ),
+            dtype=complex,
+        )
```
murrayrm, line 788: This is OK, but a different style than elsewhere in the codebase.  Normally we just wrap things in a way that uses less than 78 characters/line and uses as few lines as possible.  So something like:
```
        new_fresp = np.zeros(
            (self.noutputs + other.noutputs, self.ninputs + other.ninputs,
             self.omega.shape[-1]), dtype=complex)

File: control/tests/statesp_test.py [hunk 4]
```diff
@@ -320,6 +320,339 @@ def test_multiply_ss(self, sys222, sys322):
...
+    @slycotonly
+    def test_pow(self, sys222, sys322):
+        """Test state space powers."""
+        for sys in [sys222, sys322]:
+            # Power of 0
+            result = sys**0
+            expected = StateSpace([], [], [], np.eye(2), dt=0)
+            assert _tf_close_coeff(
+                ss2tf(expected).minreal(),
+                ss2tf(result).minreal(),
```
murrayrm, line 574: Probably OK, but could you have a situation here (or below) where the transfer functions matched but the state space system was incorrect?

murrayrm (commented):
Preliminary comments.  I haven't gone through to make sure the everything is correct, but wanted to pass on some higher level comments first.  OK to ignore the code style ones, but I would fix up some of the Numpydoc ones (which I eventually need to check in `tests/docstrings_tests.py`.

sdahdah:
Thanks for taking a look. Would you like me to fix these before you continue the review? I might not be able to until the new year unfortunately.

murrayrm (commented):
As currently written, the new systems that are being created are non-minimal.  For example:
```
s = ct.tf('s')
ct.config.defaults['xferfcn.display_format'] = 'zpk'  # simpler output format
siso = (s+1) / ((s+2) * (s+3))
mimo = ct.combine_tf([
  [1/s,         1/(s+4)], 
  [1/(s+5), (s+6)/(s**2 + 2*s + 2)]
])
print(siso * mimo)

<TransferFunction>: sys[219]
Inputs (2): ['u[0]', 'u[1]']
Outputs (2): ['y[0]', 'y[1]']

Input 1 to output 1:
      (s + 1) (s + 5)
---------------------------
(s) (s + 2) (s + 3) (s + 5)

Input 1 to output 2:
        (s) (s + 1)
---------------------------
(s) (s + 2) (s + 3) (s + 5)

Input 2 to output 1:
        (s + 1) (s + (1-1j)) (s + (1+1j))
-------------------------------------------------
(s + (1-1j)) (s + (1+1j)) (s + 2) (s + 3) (s + 4)

Input 2 to output 2:
             (s + 1) (s + 4) (s + 6)
-------------------------------------------------
(s + (1-1j)) (s + (1+1j)) (s + 2) (s + 3) (s + 4)
```
Note the various pole/zero cancellations that appear.  It seems to me that these are spurious and shouldn't show up (eg, (s+4) in the [2, 2] transfer function).

This same type of situation occurs in state space and is consistent with the results that occur if you convert the SISO system to a matrix system by hand.  I'm not sure (yet) what is gong on, but this seems like something we should fix, either for this specific case or for the more general case of multiplying MIMO systems.

I need to think a bit more about what is causing this, but wanted to flag it here in case someone else has some insights.  It seems to me that perfect pole/zero cancellations (and the equivalent in state space) should get sorted out ahead of time, either by not creating them or by removing them before we send things back to the user.

control/tests/statesp_test.py
sdahdah: In the places where I could, I replaced these with state-space matrix comparisons.

When I am trying to show that `G * G**-1` is identity (`ss([], [], [], eye(n))`), I'm not able to guarantee these will have the same state-space realization. I don't think I can compare in a canonical form for MIMO systems.

sdahdah:
> As currently written, the new systems that are being created are non-minimal. For example:
> 
> ```
> s = ct.tf('s')
> ct.config.defaults['xferfcn.display_format'] = 'zpk'  # simpler output format
> siso = (s+1) / ((s+2) * (s+3))
> mimo = ct.combine_tf([
>   [1/s,         1/(s+4)], 
>   [1/(s+5), (s+6)/(s**2 + 2*s + 2)]
> ])
> print(siso * mimo)
> 
> <TransferFunction>: sys[219]
> Inputs (2): ['u[0]', 'u[1]']
> Outputs (2): ['y[0]', 'y[1]']
> 
> Input 1 to output 1:
>       (s + 1) (s + 5)
> ---------------------------
> (s) (s + 2) (s + 3) (s + 5)
> 
> Input 1 to output 2:
>         (s) (s + 1)
> ---------------------------
> (s) (s + 2) (s + 3) (s + 5)
> 
> Input 2 to output 1:
>         (s + 1) (s + (1-1j)) (s + (1+1j))
> -------------------------------------------------
> (s + (1-1j)) (s + (1+1j)) (s + 2) (s + 3) (s + 4)
> 
> Input 2 to output 2:
>              (s + 1) (s + 4) (s + 6)
> -------------------------------------------------
> (s + (1-1j)) (s + (1+1j)) (s + 2) (s + 3) (s + 4)
> ```
> 
> Note the various pole/zero cancellations that appear. It seems to me that these are spurious and shouldn't show up (eg, (s+4) in the [2, 2] transfer function).
> 
> This same type of situation occurs in state space and is consistent with the results that occur if you convert the SISO system to a matrix system by hand. I'm not sure (yet) what is gong on, but this seems like something we should fix, either for this specific case or for the more general case of multiplying MIMO systems.
> 
> I need to think a bit more about what is causing this, but wanted to flag it here in case someone else has some insights. It seems to me that perfect pole/zero cancellations (and the equivalent in state space) should get sorted out ahead of time, either by not creating them or by removing them before we send things back to the user.

I took a quick look but I'm not sure. It's definitely happening inside `xferfcn.TransferFunction.__mul__`. I think roots need to be cancelled in `xferfcn._add_siso()`, but I'm not sure whether `minreal()` should just be used?

sdahdah:
Happy new year @murrayrm . I fixed some of your initial comments (and also tried to look for other instances of nonstandard docstrings and indentation along the way). I'm not 100% sure what's going on with the spurious poles and zeros, buy maybe another issue should be raised to track that?

sdahdah:
Not sure why the `Py3.12; conda Slycot; conda Pandas; conda CVXOPT ; QtAgg` check is failing during setup:

```
error    libmamba Error when extracting package: Could not stat include/event2/keyvalq_struct.h
libevent-2.1.12-hf998b51_1.conda extraction failed
Found incorrect download: libevent. Aborting
```

I don't think it's related to the PR

Issue #1087, opened 2024-12-30
control.timeresp.TimeResponseData.to_pandas() failing


joaoantoniocardoso:
Hi, today I was using some step responses and noticed that the `.to_pandas()` is not actually working.

I managed to workaround it by creating my own function to translate the response into a dataframe.

Example of code not working:

```Python
import control as ct
import numpy as np

model = ct.rss(states=['x0', 'x1'], outputs=['y0', 'y1'], inputs=['u0', 'u1'], name='My Model')

T = np.linspace(0, 10, 100, endpoint=False)
X0 = np.zeros(model.nstates)

res = ct.step_response(model, T=T, X0=X0, input=0)

df = res.to_pandas()
```

Error:

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[140], line 6
      3 T = np.linspace(0, 10, 100, endpoint=False)
      4 X0 = np.zeros(model.nstates)
----> 6 res = ct.step_response(model, T=T, X0=X0, input=0).to_pandas()

File ~.env/lib/python3.10/site-packages/control/timeresp.py:723, in TimeResponseData.to_pandas(self)
    719 if self.nstates > 0:
    720     data.update(
    721         {name: self.x[i] for i, name in enumerate(self.state_labels)})
--> 723 return pandas.DataFrame(data)

File ~.env/lib/python3.10/site-packages/pandas/core/frame.py:778, in DataFrame.__init__(self, data, index, columns, dtype, copy)
    772     mgr = self._init_mgr(
    773         data, axes={"index": index, "columns": columns}, dtype=dtype, copy=copy
    774     )
    776 elif isinstance(data, dict):
    777     # GH#38939 de facto copy defaults to False only in non-dict cases
--> 778     mgr = dict_to_mgr(data, index, columns, dtype=dtype, copy=copy, typ=manager)
    779 elif isinstance(data, ma.MaskedArray):
    780     from numpy.ma import mrecords

File ~.env/lib/python3.10/site-packages/pandas/core/internals/construction.py:503, in dict_to_mgr(data, index, columns, dtype, typ, copy)
    499     else:
    500         # dtype check to exclude e.g. range objects, scalars
    501         arrays = [x.copy() if hasattr(x, "dtype") else x for x in arrays]
--> 503 return arrays_to_mgr(arrays, columns, index, dtype=dtype, typ=typ, consolidate=copy)

[...]

PR #1088, opened 2024-12-31
Fix TimeResponseData.to_pandas() for multi-trace responses

murrayrm:
This PR addresses issue #1087.  Multi-trace data (e.g., the output from a MIMO step response) was not being processed correctly.  This PR creates a new column 'trace' for multi-trace responses and reshapes the data appropriately.  Unit tests and updated documentation are included.

Thanks to @joaoantoniocardoso for finding this bug and providing a fix (modified only slightly here).

@joaoantoniocardoso: if you have time to try this out and confirm that it works for your use case, that would be great.  I think the only difference is that I used 'trace' for the column label instead of 'trace_label'.  I also had to do things a bit differently for the case where there were multiple traces (your example was "multi-trace" but with sys.ntrace=1 since you selected an input).

### DIFF
--- a/control/tests/timeresp_test.py
+++ b/control/tests/timeresp_test.py
@@ -1266,16 +1266,28 @@ def test_to_pandas():
     np.testing.assert_equal(df['u[0]'], resp.inputs)
     np.testing.assert_equal(df['y[0]'], resp.inputs * 5)
 
+    # Multi-trace data
     # https://github.com/python-control/python-control/issues/1087
     model = ct.rss(
         states=['x0', 'x1'], outputs=['y0', 'y1'],
         inputs=['u0', 'u1'], name='My Model')
     T = np.linspace(0, 10, 100, endpoint=False)
     X0 = np.zeros(model.nstates)
-    res = ct.step_response(model, T=T, X0=X0, input=0)
+
+    res = ct.step_response(model, T=T, X0=X0, input=0)  # extract single trace
     df = res.to_pandas()
-    np.testing.assert_equal(df['time'], res.time)
-    np.testing.assert_equal(df['y1'], res.outputs['y1'])
+    np.testing.assert_equal(
+        df[df['trace'] == 'From u0']['time'], res.time)
+    np.testing.assert_equal(
+        df[df['trace'] == 'From u0']['y1'], res.outputs['y1', 0])
+
+    res = ct.step_response(model, T=T, X0=X0)           # all traces
+    df = res.to_pandas()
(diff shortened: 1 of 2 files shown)


control/tests/timeresp_test.py
```diff
@@ -1265,6 +1266,29 @@ def test_to_pandas():
...
+        inputs=['u0', 'u1'], name='My Model')
+    T = np.linspace(0, 10, 100, endpoint=False)
+    X0 = np.zeros(model.nstates)
+
+    res = ct.step_response(model, T=T, X0=X0, input=0)  # extract single trace
+    df = res.to_pandas()
+    np.testing.assert_equal(
+        df[df['trace'] == 'From u0']['time'], res.time)
+    np.testing.assert_equal(
+        df[df['trace'] == 'From u0']['y1'], res.outputs['y1', 0])
```
slivingston, line 1282:
```suggestion
    np.testing.assert_equal(
        df[df['trace'] == 'From u0']['time'], res.time)
    np.testing.assert_equal(
        df[df['trace'] == 'From u0']['u0'], res.inputs['u0', 0])
    np.testing.assert_equal(
        df[df['trace'] == 'From u0']['y1'], res.outputs['y1', 0])
```
Include comparison of `inputs` here to be symmetric with comparisons in the for-loop below.

control/tests/timeresp_test.py
```diff
@@ -1265,6 +1266,29 @@ def test_to_pandas():
...
+
+    res = ct.step_response(model, T=T, X0=X0)           # all traces
+    df = res.to_pandas()
+    for i, label in enumerate(res.trace_labels):
+        np.testing.assert_equal(
+            df[df['trace'] == label]['time'], res.time)
+        np.testing.assert_equal(
+            df[df['trace'] == label]['u0'], res.inputs['u0', i])
```
slivingston, line 1290:
```suggestion
        np.testing.assert_equal(
            df[df['trace'] == label]['time'], res.time)
        np.testing.assert_equal(
            df[df['trace'] == label]['u0'], res.inputs['u0', i])
        np.testing.assert_equal(
            df[df['trace'] == label]['y1'], res.outputs['y1', i])
```
Include comparison of `outputs` here to be symmetric with comparisons for the case of single-trace above.

slivingston (approved):

joaoantoniocardoso (approved):
Woah, that one was quick!

I've just tested it here, and it worked as a charm!

Thank you so much for keeping this lib, awesome work! And happy New Year! =D

Issue #1105, opened 2025-01-29
phase_crossover_frequencies generates errors with poles at origin


murrayrm:
### Discussed in https://github.com/python-control/python-control/discussions/1089

<div type='discussions-op-text'>

<sup>Originally posted by **alittlepinetree** December 31, 2024</sup>
When I used this function to calculate the phase crossover frequencies, it didn't work.
`import control
import matplotlib.pyplot as plt

G = control.tf([200.0], [1.0, 21.0, 20.0, 0.0])
x_omega, x_gain = control.phase_crossover_frequencies(G)
print(x_omega)`</div>

murrayrm:
Fixed in #1106

PR #1106, opened 2025-01-29
Turn off warn_infinite for phase_crossover_frequencies() calculation

murrayrm:
This PR addresses #1105 where warning messages were being generated if a system had poles at the origin (which generates a phase crossover at infinity).  The fix was simple: warning messages should be turned off when computing the crossover gains (as they are in `stability_margins`).

### DIFF
--- a/control/tests/margin_test.py
+++ b/control/tests/margin_test.py
@@ -12,11 +12,9 @@ import pytest
 from numpy import inf, nan
 from numpy.testing import assert_allclose
 
-from ..frdata import FrequencyResponseData
-from ..margins import margin, phase_crossover_frequencies, stability_margins
-from ..statesp import StateSpace
-from ..xferfcn import TransferFunction
-from ..exception import ControlMIMONotImplemented
+from control import ControlMIMONotImplemented, FrequencyResponseData, \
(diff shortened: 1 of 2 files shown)


control/tests/margin_test.py
```diff
@@ -119,6 +118,8 @@ def test_margin_3input(tsys):
      (([2], [1, 3, 3, 1]), [1.732, 0.], [-0.25, 2.]),
      ((np.array([3, 11, 3]) * 1e-4, [1., -2.7145, 2.4562, -0.7408], .1),
       [1.6235, 0.], [-0.28598, 1.88889]),
+     (([200.0], [1.0, 21.0, 20.0, 0.0]),
+      [4.47213595, 0], [-0.47619048, inf]),
```
slivingston, line 122: Warnings do not cause the test to fail, so this case does not provide a regression test. To fix this, add the decorator `@pytest.mark.filterwarnings("error")` as done [in another test in margin_test.py](https://github.com/python-control/python-control/blob/2f59e954830fbb98cdde8ca4743cf171c82d4111/control/tests/margin_test.py#L360).

slivingston (commented):
Good work! Besides my comment about the regression test, this is ready to merge.

slivingston (approved):

PR #1108, opened 2025-02-01
scipy-based implementation of ss2tf

murrayrm:
This PR allows `ss2tf` to work using `method=scipy` for MIMO systems, allow MIMO ss2tf without requiring `slycot`.  Also re-enabled unit tests that now work without `slycot`.


bnavigator (approved):
LGTM

Issue #1101, opened 2025-01-19
pyflakes (pylint, ruff, ....) clean-up, CI?


roryyorke:
Running pyflakes on main HEAD (0ff045263f) reveals lots of possible issues; see end.

1. Is there interest in getting and keeping the tree lint-clean?

    This would require some changes to style, e.g., dropping `from foo import *` (see output below), and some patience when the linter spits out an incomprehensible message.
 
2. Are there strong preferences for a particular linter?

    I use pyflakes often: it's simple, has a low false-positive rate, but is not configurable.

    The other two I'm aware of:
      - pylint, more extensive checks, configurable
      - ruff; my crude, probably unfair thinking of it is "pylint written in Rust", but I think it may do more

pyflakes output:

```
control/modelsimp.py:343:9: local variable 'order' is assigned to but never used
control/pzmap.py:14:1: 'math.pi' imported but unused
control/pzmap.py:18:1: 'numpy.cos' imported but unused
control/pzmap.py:18:1: 'numpy.exp' imported but unused
control/pzmap.py:18:1: 'numpy.linspace' imported but unused
control/pzmap.py:18:1: 'numpy.sin' imported but unused
control/pzmap.py:18:1: 'numpy.sqrt' imported but unused
control/pzmap.py:25:1: '.freqplot._freqplot_defaults' imported but unused
control/pzmap.py:28:1: '.lti.LTI' imported but unused
control/stochsys.py:25:1: '.iosys.InputOutputSystem' imported but unused
control/stochsys.py:25:1: '.iosys._process_indices' imported but unused
control/stochsys.py:29:1: redefinition of unused 'StateSpace' from line 23
control/stochsys.py:173:9: local variable 'NN' is assigned to but never used
control/stochsys.py:292:9: local variable 'NN' is assigned to but never used
control/stochsys.py:449:13: undefined name 'warnings'
control/stochsys.py:463:9: 'raise NotImplemented' should be 'raise NotImplementedError'
control/iosys.py:11:1: 'warnings.warn' imported but unused
control/nlsys.py:21:1: 'copy' imported but unused
control/nlsys.py:28:1: '.iosys._process_signal_list' imported but unused
control/nlsys.py:214:9: local variable 'dt' is assigned to but never used
control/nlsys.py:246:9: local variable 'dt' is assigned to but never used
control/nlsys.py:814:16: f-string is missing placeholders
control/nlsys.py:834:16: f-string is missing placeholders
[...]

PR #1118, opened 2025-02-02
Lint library code only with `ruff check`

roryyorke:
Make library code `ruff check` clean for pyflakes ("F" code) warnings, and checks this in a Github Action.

"Library code" is all non-test code in `control/`, so excludes tests, examples, and benchmarks.  Making all of those ruff-check-clean too will take some time, and be a big change, so it seems worth getting this in first.  This also gives an opportunity for reviewing the ruff configuration, and the overall approach, e.g., disabling import checks in `__init__.py`.

Besides making the other parts of the code pass lint, `ruff` can do more, and we can look at adding more checks later.

Partially addresses gh-1101.

You can see an example of a failing Github Action ruff check [here](https://github.com/roryyorke/python-control/actions/runs/13088422392).

### DIFF
--- a/.github/workflows/ruff-check.yml
+++ b/.github/workflows/ruff-check.yml
@@ -1,0 +1,29 @@
+# run ruff check on library source
+# TODO: extend to tests, examples, benchmarks
+
+name: ruff-check
+
+on: [push, pull_request]
+
+jobs:
+  ruff-check-linux:
+    # ruff *shouldn't* be sensitive to platform
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout python-control
+      uses: actions/checkout@v3
+
+    - name: Setup environment
+      uses: actions/setup-python@v4
+      with:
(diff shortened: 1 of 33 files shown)


roryyorke:
failure in "", in step `Install optional dependencies`.  Seem to be getting these often.

possibly relevant parts there:

```
warning  libmamba Could not parse mod/etag header
```

```
error    libmamba Error when extracting package: Could not stat info/recipe/parent/install-libstdc++-devel.sh
libgfortran-ng-14.2.0-h69a702a_1.conda extraction failed
Found incorrect download: libgfortran-ng. Aborting
```

murrayrm:
Thanks for putting this together @roryyorke!

These changes look good to me.  I'll hold on merging in case others want to take a look, but will plan on merging in the next day or two.

I'll create an issue to discuss the sporadically failing CI tests. Very annoying to have to re-run them and it seems to be increasing in frequency.

PR #1127, opened 2025-02-16
Lint tests

roryyorke:
This extends ruff testing to control/tests/*.py

I've tried to order the commits from obvious to subtle; on the whole:

  - remove unused imports
  - removed unused variables, or sometimes prefix with '_' where that seemed to make things clearer
  - tweaked two `eval` calls

The remaining errors I'm less sure of.  It would be simple to silence them with `# noqa: F<relevant code>`, and defer fixing these to later; then at least no new ruff errors will sneak in in the future.

More detail on remaining failures:

## `control/tests/kwargs_test.py`

Here the dict literal has repeated keys for `'StateSpace.__init__'` and `'StateSpace.__sample'`:

```
    'StateSpace.__init__':
        interconnect_test.test_interconnect_exceptions,
    'StateSpace.sample': test_unrecognized_kwargs,
    'NonlinearIOSystem.__init__':
        interconnect_test.test_interconnect_exceptions,
    'StateSpace.__init__': test_unrecognized_kwargs,
    'StateSpace.initial_response': timeresp_test.test_timeresp_aliases,
    'StateSpace.sample': test_unrecognized_kwargs,
```

[Last-value-wins in dict literals](https://stackoverflow.com/a/34539878/1008142), so the first two entries have no effect.  It looks like this is just a check that *something* exists, so I guess we can just erase the first two entries?

## `control/tests/modelsimp_test.py`

The variable `mrk` is assigned to: `mrk = markov(outp, inp, 1, transpose=False)`.  I didn't investigate issue #395 referenced above -- is it enough that this call doesn't raise?

## control/tests/optimal_test.py

`curved_seg_length` is calculated, but not used.  It's not obviously copy-and-pasted from another test, so I'm not sure if the test should be changed here.  It's obviously easy enough to just remove this variable, and `straight_seg_length` just above it.


murrayrm:
> control/tests/kwargs_test.py

Correct that there just needs to be some test that verifies that unknown keywords generate an error.  Whichever of the tests that has that check should be kept and if both do then keeping either is fine.

> control/tests/modelsimp_test.py

Correct that the intent of this test was just to make sure an exception wasn't raised.  We could either get rid of the assignment to `mrk` or check the output to make sure it does something sensible.  For example:
```
np.testing.assert_almost_equal(mrk, 2.)
```
> control/tests/optimal_test.py

I'm pretty sure I wrote this example and I can't see any reason to keep `straight_seg_length` or `curved_seg_length`.

roryyorke:
Oops - had a commit there I'd intended for another PR.  Reverted.

If this PR isn't merged now, it'll be worth merging main into the PR before merging the PR to main, to make sure the lint checks don't causes failures on main after the merge of the PR.  (I hope that sentence is less confusing than I think it is!)

PR #1135, opened 2025-03-02
Lint fixes on benchmarks and examples/*.py

roryyorke:
I don't think these are too controversial.

# CI failures

I see two failures on CI , but I can't see that these changes caused them:

first, in `doctest-linux`: https://github.com/roryyorke/python-control/actions/runs/13612156165/job/38050709282

```
Traceback
=========

      File "/home/runner/miniconda3/envs/doctest-env/lib/python3.12/site-packages/sphinx/events.py", line 415, in emit
        raise ExtensionError(
    sphinx.errors.ExtensionError: Handler <function html_collect_pages at 0x7f36a35016c0> for event 'html-collect-pages' threw an exception (exception: module 'sphinx.util' has no attribute 'console')
```

which seems to be https://github.com/sphinx-doc/sphinx/issues/13352.

Then in `install-examples`: https://github.com/roryyorke/python-control/actions/runs/13612156159/job/38050709274

```
ModuleNotFoundError: No module named 'autoreload'
```

I'll install jupyterlab in my dev environment and see if I can reproduce this.  I can't see that changing *.py files in examples could cause this.

# tfvis.py and  ~~Pwm~~ Pmw

When I ran `tfvis.py` under Python 3.12, I got

```
AttributeError: module 'collections' has no attribute 'Callable'
```

in ~~Pwm~~ Pmw code.

`Callable` was moved to `collections.abc` in Python 3.10 (October 2021).  The hack below lets the script run, but unless ~~pwm~~ Pmw is updated we should consider retiring this example.

```diff
modified   examples/tfvis.py
@@ -48,6 +48,9 @@ from matplotlib.backends.backend_tkagg import FigureCanvasTkAgg
 from control.matlab import logspace
 from numpy import array, conj, polymul, real, size
 
+import collections.abc
+collections.Callable = collections.abc.Callable
+
 
 def make_poly(facts):
     """ Create polynomial from factors """
```

# Warnings in examples

Some examples raise control deprecation warnings, e.g., `genswitch.py` uses `phase_plot` instead of `phase_plane_plot`.  I started looking into this, but it got a bit involved.

# Lint on notebooks

`ruff` can also run on Jupyter notebooks.  Some fixes  are obvious:

```
examples/bode-and-nyquist-plots.ipynb:cell 2:2:17: F401 [*] `scipy` imported but unused
  |
1 | import numpy as np
2 | import scipy as sp
  |                 ^^ F401
3 | import matplotlib.pyplot as plt
4 | import control as ct
  |
  = help: Remove unused import: `scipy`
```

but I'm less sure about changing these `*` imports; they are convenient, and make things more familiar for Matlab users:

```
examples/pvtol-lqr-nested.ipynb:cell 3:2:1: F403 `from matplotlib.pyplot import *` used; unable to detect undefined names
  |
1 | from numpy import *             # Grab all of the NumPy functions
2 | from matplotlib.pyplot import * # Grab MATLAB plotting functions
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ F403
3 | from control.matlab import *    # MATLAB-like functions
4 | %matplotlib inline
  |
```

### DIFF
--- a/benchmarks/flatsys_bench.py
+++ b/benchmarks/flatsys_bench.py
@@ -7,7 +7,6 @@
 
 import numpy as np
 import math
-import control as ct
 import control.flatsys as flat
 import control.optimal as opt
 
--- a/benchmarks/optestim_bench.py
+++ b/benchmarks/optestim_bench.py
@@ -6,7 +6,6 @@
 # used for optimization-based estimation.
 
 import numpy as np
-import math
 import control as ct
 import control.optimal as opt
 
--- a/benchmarks/optimal_bench.py
+++ b/benchmarks/optimal_bench.py
@@ -6,7 +6,6 @@
 # performance of the functions used for optimization-base control.
 
 import numpy as np
-import math
 import control as ct
 import control.flatsys as fs
 import control.optimal as opt
@@ -20,7 +19,6 @@ integrator_table = {
     'default': (None, {}),
     'RK23': ('RK23', {}),
     'RK23_sloppy': ('RK23', {'atol': 1e-4, 'rtol': 1e-2}),
-    'RK45': ('RK45', {}),
     'RK45': ('RK45', {}),
     'RK45_sloppy': ('RK45', {'atol': 1e-4, 'rtol': 1e-2}),
     'LSODA': ('LSODA', {}),
@@ -129,9 +127,6 @@ def time_optimal_lq_methods(integrator_name, minimizer_name, method):
     Tf = 10
     timepts = np.linspace(0, Tf, 20)
 
-    # Create the basis function to use
-    basis = get_basis('poly', 12, Tf)
-
     res = opt.solve_ocp(
         sys, timepts, x0, traj_cost, constraints, terminal_cost=term_cost,
         solve_ivp_method=integrator[0], solve_ivp_kwargs=integrator[1],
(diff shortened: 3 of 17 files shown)


murrayrm:
The CI errors are also showing up in #1133, so probably something changed on the GitHub Actions end of things.  I'll set up a new issue to track that.

roryyorke:
> The CI errors are also showing up in https://github.com/python-control/python-control/pull/1133, so probably something changed on the GitHub Actions end of things.

Thanks.  I can't reproduce the notebook conversion failure locally.  I've attached the output of `mamba list` to record versions of everything installed.

I see there's already a bug on Pmw for `collections.Callable`, filed in 2020: https://sourceforge.net/p/pmw/bugs/52/

[packages.txt](https://github.com/user-attachments/files/19041895/packages.txt)

murrayrm:
Sorry for the delay in reviewing this.  All of the changes look good to me.

Some comments on your comments:

> tfvis.py and Pmw
>
> Callable was moved to collections.abc in Python 3.10 (October 2021). The hack below lets the script run, but unless pwm Pmw is updated we should consider retiring this example.

I think it would be fine to move this to `examples/legacy`.  To a large extent the functionality demonstrated in the script is captured  in `sisotool`.

> Warnings in examples
> 
> Some examples raise control deprecation warnings, e.g., genswitch.py uses phase_plot instead of phase_plane_plot. I started looking into this, but it got a bit involved.

Perhaps flag some of these as separate issues?  For the specific case of `genswitch.py` (and perhaps others), I likely have new versions that I did for FBS2e (I'm slowly converting all examples and figures to use `python-control`).

> Lint on notebooks
>
> ... I'm less sure about changing these * imports; they are convenient, and make things more familiar for Matlab users ...

It would be nice to run lint on the notebooks.  I agree that for MATLAB users, it will be convenient to use `from control.matlab import *` and perhaps we should just flag that one as OK wherever it appears.  But I believe that the main control toolbox should always use a prefix (preferably `ct` for consistency).

murrayrm (approved):

Issue #1136, opened 2025-03-02
CI failures in doctest, examples


murrayrm:
From #1135 (also in #1133):

> I see two failures on CI , but I can't see that these changes caused them:
> 
> first, in doctest-linux: [roryyorke/python-control/actions/runs/13612156165/job/38050709282](https://github.com/roryyorke/python-control/actions/runs/13612156165/job/38050709282)
> ```
> Traceback
> =========
> 
>       File "/home/runner/miniconda3/envs/doctest-env/lib/python3.12/site-packages/sphinx/events.py", line 415, in emit
>         raise ExtensionError(
>     sphinx.errors.ExtensionError: Handler <function html_collect_pages at 0x7f36a35016c0> for event 'html-collect-pages' threw an exception (exception: module 'sphinx.util' has no attribute 'console')
> which seems to be https://github.com/sphinx-doc/sphinx/issues/13352.
> ```
> Then in install-examples: [roryyorke/python-control/actions/runs/13612156159/job/38050709274](https://github.com/roryyorke/python-control/actions/runs/13612156159/job/38050709274)
> ```
> ModuleNotFoundError: No module named 'autoreload'
> ```

roryyorke:
I think the relevant error is the one you showed in #1133; Internet search on that leads to

https://github.com/ipython/ipython/issues/14808

and

https://github.com/ipython/ipython/pull/14807

Maybe for both problems we need to limit the version of the relevant package (sphinx-doc, IPython) used in the install?  I think Ben's previously used '!=' magic to prevent specific package versions being used, rather than '<' or '<=', but I guess whatever works.  See e.g. https://github.com/fpsim/fpsim/pull/493/files which fixes the sphinx-doc error for that project.

PR #1137, opened 2025-03-02
Fix CI doctest, install-examples

roryyorke:
Fixes gh-1136.

I tried '!=8.2' for sphinx-doc, which didn't work, so I used '<8.2'. I think '!=8.2.*' might, but didn't try it.  See https://packaging.python.org/en/latest/specifications/version-specifiers/#id5 (not sure this applies to Conda/Mamba - it should).

### DIFF
--- a/.github/conda-env/doctest-env.yml
+++ b/.github/conda-env/doctest-env.yml
@@ -7,7 +7,7 @@ dependencies:
   - numpy
   - matplotlib
   - scipy
-  - sphinx
+  - sphinx<8.2
   - sphinx_rtd_theme
   - ipykernel
   - nbsphinx
--- a/.github/workflows/install_examples.yml
+++ b/.github/workflows/install_examples.yml
@@ -20,7 +20,8 @@ jobs:
            --quiet --yes \
            python=3.12 pip \
            numpy matplotlib scipy \
-           slycot pmw jupyter
+           slycot pmw jupyter \
+           ipython!=9.0
 
     - name: Install from source
       run: |


murrayrm:
It would be nice to block out just the version that is failing, if possible, so that we will test against the latest new release when it comes out (and hopefully the bug is fixed).

Also, should we put something in `pyproject.toml` to flag this issue?  If not, then a user could install with the latest version of iPython (or sphinx) and things might not work.  I'm not quite sure where that would go, since python-control is not *dependent* on iPython.

roryyorke:
For IPython, it's reasonably likely that an upcoming release, say 9.0.1, will fix this (https://github.com/ipython/ipython/pull/14807 claims to fix it).  If it's later, 9.0.2 or whatever, we can update the relevant config file(s) to exclude all the broken version.  So here I think the CI fix is OK.

For sphinx, the situation is trickier.  From https://github.com/sphinx-doc/sphinx/issues/13364 and https://github.com/sphinx-doc/sphinx/issues/13352 , sphinx have made a change that breaks nbsphinx, and nbsphinx has a PR for necessary updates at https://github.com/spatialaudio/nbsphinx/pull/826, but it's not clear when that will be merged and released.

I don't think there's a way to say "use nbsphinx >= 0.9.7, or failing that, use sphinx < 8.2.*", so I think the sphinx version condition I've added is a stop-gap until there's a released nbsphinx fix.  I know this sort of stop-gap can easily last months (or years), and I don't know what to do about that.  I guess a TODO in the `doctest-env.yml`, and open an issue now.

I've pushed a prototype of `sphinx!=8.2.*`, see https://github.com/roryyorke/python-control/actions/runs/13617760094/job/38063032976 , but I now think it's just as much of a stop-gap as `<8.2`.

As for pyproject.toml: I don't know if we can or should do anything about IPython.  It's a clear bug (install IPython, run `load_ext autoreload`, and it's triggered) that doesn't affect only control-python, and that should be fixed soon(ish).  The  same reasoning applies to sphinx and nbsphinx, but in addition I imagine non-dev users rarely install these.

murrayrm:
Thanks @roryyorke for the detail comments and the fix.  I'll let this sit for a bit (at least rest of the day) before merging, in case others have comments.  But I think this is otherwise good to go (and we need to get something in place soon to fix the CI tests for current and upcoming PRs).

PR #1148, opened 2025-05-11
Add support for continuous delay systems

MythasNauveili:
This PR introduces initial support for linear time-invariant (LTI) systems with time delays in `python-control`, inspired by  the capabilities of [ControlSystems.jl](https://github.com/JuliaControl/ControlSystems.jl) package, and the `InternalDelay.py` script from [Skogestad-Python](https://github.com/alchemyst/Skogestad-Python/tree/master). 

**Key features**
* **Creation of delay sytems** Systems with delays can now be intuitively constructed by multiplying a `TransferFunction` object with a `delay(tau)` object or an `exp(-tau*s)` expression:
    ```python
    tf_delay = tf([1], [1,1]) * delay(2)
    # or
    s = tf("s")
    tf_delay = tf([1], [1,1]) * exp(-2*s)
    ```
*   **Underlying Representation:** The approach utilizes Linear Fractional Transformations (LFT) with a `PartitionedStateSpace` representation (implemented in `partitionedssp.py`). The `DelayLTI` class (in `delaylti.py`) serves as the core object for these systems.
*   **Operations:** Standard operations such as addition, multiplication, and feedback are supported for `DelayLTI` objects.
*   **Time and Frequency Responses:** Time responses for delayed systems are computed by solving the underlying Delay Differential Equations (DDEs). Frequency responses are also supported.
*   **Retrocompatibility:** Maintain retrocompatibility with existing functionalities.

**Implementation Details:**
*   **DDE Solver:** Time responses for delayed systems are computed by solving DDEs. After exploring alternatives, a custom DDE solver based on the Method of Steps, wrapping SciPy's `solve_ivp`, was implemented in `dde.py`.
    *   **Accuracy Note:** While this solver performs well for many cases, for stiff problems, it may require smaller step sizes for convergence compared to specialized DDE solvers (like those in ControlSystems.jl). In tested scenarios, the error typically remains at least three orders of magnitude below the system response with reasonable step sizes. (An error estimation algorithm could be a valuable future addition).
*   **Validation:** A Julia script (`control/julia/compute_tests.jl`) using ControlSystems.jl has been added to generate reference results, which are exported to a JSON file. These results are used in the test suite (`control/tests/delay_lti_test.py` and `control/tests/dde_test.py`) to benchmark the accuracy of the `DelayLTI` implementation and DDE solver.

**Current Scope and Limitations:**
*   This implementation currently focuses on **continuous-time** delay systems. Discrete-time delay systems are not yet covered.
*   The DDE solver may benefit from further enhancements for stiff problems or adaptive step sizing.

**Discussion Points:**
This PR represents the initial groundwork for this feature. I'm submitting this draft to share these first steps and would appreciate feedback on:
*   The overall approach and class design (`DelayLTI`, `PartitionedStateSpace`).
*   The DDE solver implementation and its current performance.
*   Any potential integration issues or API considerations.

I'm aware there is still work to be done for this feature to be ready for a full release, but I believe this provides a solid foundation.


control/dde.py
```diff
@@ -0,0 +1,374 @@
...
+    )
+
+    # Test if U has correct shape and type
+    legal_shapes = (
+        [(n_steps,), (1, n_steps)] if n_inputs == 1 else [(n_inputs, n_steps)]
+    )
+    U = _check_convert_array(
+        U, legal_shapes, "Parameter `U`: ", squeeze=False, transpose=transpose
+    )
+    print("U shape: ", U.shape)
```
murrayrm, line 89: This looks like a debugging statement that was left in the code?

control/dde.py
```diff
@@ -0,0 +1,374 @@
...
+
+    # TODO: handle discontinuity propagation
+    discontinuity_times = set(tau_list)
+    while current_t < tf:
+        t_stop = min(discontinuity_times) if discontinuity_times else tf
+        if not np.isclose(t_stop, tf):
+            discontinuity_times.remove(t_stop)
+        local_t_eval = [t for t in T if current_t < t <= t_stop]
+
+        print("Integrate bewtween ", current_t, " and ", t_stop)
```
murrayrm, line 331: Debugging statement?

control/__init__.py
```diff
@@ -89,6 +89,7 @@
 
 # Allow access to phase_plane functions as ct.phaseplot.fcn or ct.pp.fcn
 from . import phaseplot as phaseplot
+
```
murrayrm, line 92: Was there a reason to add an extraneous blank line here?

control/dde.py
```diff
@@ -0,0 +1,374 @@
...
+        plot_inputs=True,
+        title="Forced response for " + delay_sys.name,
+        trace_types=["forced"],
+        transpose=transpose,
+        return_x=return_x,
+        squeeze=squeeze,
+    )
+
+
+def pchip_interp_u(T, U):
```
murrayrm, line 113: This looks like it is probably an internal function.  If so, add underscore to start of function name.

control/dde.py
```diff
@@ -0,0 +1,374 @@
...
+        return np.array([negative_wrapper(PchipInterpolator(T, U))])
+    elif np.ndim(U) == 0:
+        # Scalar input, assumed constant.
+        return U
+    else:
+        # Multiple input signals, U.shape is (n_inputs, n_steps)
+        return np.array([negative_wrapper(PchipInterpolator(T, ui)) for ui in U])
+
+
+class DdeHistory:
```
murrayrm, line 150: This might be better named `DDEHistory` (since DDE is a standard abbreviation, like LTI)?

Also, if this is an internal class, begin the class name with an underscore.

control/dde.py
```diff
@@ -0,0 +1,374 @@
...
+        else:
+            for segment in self.segments:
+                if segment.t[0] <= t <= segment.t[-1]:
+                    return segment.sol(t)
+            # Fallback: should ideally not be reached if t is within (t0, last_valid_time]
+            # and segments cover this range.
+            return np.zeros_like(self.last_state)  # Deal with first call
+
+
+def dde_wrapper(t, x, A, B1, B2, C2, D21, tau_list, u_func, history_x):
```
murrayrm, line 226: Internal function?

control/dde.py
```diff
@@ -0,0 +1,374 @@
...
+        z = C2 @ history_x(t - tau) + D21 @ u_delayed
+        z_delayed.append(z[i])
+    z_delayed = np.array(z_delayed).flatten()
+
+    u_current = np.array([u_func[i](t) for i in range(len(u_func))])
+    dxdt = A @ x + B1 @ u_current + B2 @ z_delayed
+    return dxdt.flatten()
+
+
+def solve_dde(delay_sys, T, U, X0, dt):
```
murrayrm, line 271: Internal function?

control/dde.py
```diff
@@ -0,0 +1,374 @@
+import numpy as np
+
+from scipy.integrate import solve_ivp, OdeSolution
+from scipy.interpolate import PchipInterpolator
+from typing import List
+
+
+def dde_response(
```
murrayrm, line 8: It would be nice to update `input_output_response` so that it calls `dde_response` if it is passed a `DelayLTI` system, so that there is a single interface.  Similarly, the LTI response functions (`step_response`, etc) should be updated to handle the `DelayLTI` case.  (This might already be handled by the same code that handles nonlinear I/O systems passed to the LTI response functions).

control/delaylti.py
```diff
@@ -0,0 +1,842 @@
+import numpy as np
```
murrayrm, line 1: Update to use standard template (see above).

control/delaylti.py
```diff
@@ -0,0 +1,842 @@
...
+    __call__(x, squeeze=False, warn_infinite=True)
+        Evaluate the system's frequency response at complex frequencies.
+    to_ss()
+        Convert to StateSpace (not implemented for DelayLTI).
+    to_tf()
+        Convert to TransferFunction (not implemented for DelayLTI).
+
+    Notes
+    -----
+    The way to create a DelayLTI object is by multiplying a transfert
```
murrayrm, line 89: Typo: transfert → transfer

control/delaylti.py
```diff
@@ -0,0 +1,842 @@
...
+
+    >>> ct.tf([1], [1,1]) * delay(1.5)
+
+    Or
+
+    >>> s = ct.tf('s')
+    >>> ct.tf([1], [1,1]) * exp(-1.5*s)
+
+    It's possible to MIMO delayed systems from arrays of SISO delayed systems,
+    see function mimo_delay
```
murrayrm, line 100: It's possible to MIMO → It is possible to create

Also missing a period at the end.

control/delaylti.py
```diff
@@ -0,0 +1,842 @@
...
+            raise TypeError("Input must be a TransferFunction")
+        return DelayLTI.from_ss(tf2ss(sys), tau)
+
+    def size(self):
+        """Return the number of outputs and inputs."""
+        return (self.noutputs, self.ninputs)
+
+    # Poles and zeros functions for DelayLTI are not supported by julia ControlSystems.jl
+    # might not be accurate for delayLTI, to be discussed
+    # Here the poles and zeros computed are the ones of the system without taking into accoutn the delay
```
murrayrm, line 181: Typo + may be longer than 78 char (PEP8).

control/delaylti.py
```diff
@@ -0,0 +1,842 @@
...
+                    K = K.reshape(self.ninputs, 1)
+                elif K.ndim != 2:
+                    raise ValueError("Feedback gain must be scalar, 1D, or 2D array.")
+            else:
+                raise TypeError(f"Unsupported type for static feedback: {type(other)}")
+
+            # Check dimensions of K
+            if K.shape != (self.nu, self.ny):
+                raise ValueError(
+                    f"Feedback gain K has incompatible shape. Expected ({self.nu}, {self.ny}), got {K.shape}."
```
murrayrm, line 488: Update to avoid line with more than 78 characters and put closing parenthesis at end of string, for consistency with current code base.

control/delaylti.py
```diff
@@ -0,0 +1,842 @@
...
+            )
+
+            clsys_ss = StateSpace(A_new, B_new, C_new, D_new, self.dt)
+            clsys_part = PartitionedStateSpace(clsys_ss, nu1=n_r, ny1=self.ny)
+            return DelayLTI(clsys_part, taus)
+
+    def issiso(self):
+        """Check if the system is single-input, single-output."""
+        # Based on EXTERNAL dimensions
+        return self.nu == 1 and self.ny == 1
```
murrayrm, line 552: This should be handled by the parent class (`InputOutputSystem`).

murrayrm (changes requested):
Initial round of comments.  I haven't yet checked functionality, but found a few style issues that should be updated.

murrayrm:
This would be a great contribution, @MythasNauveili.  Thanks for working on it!

There are some issues with unit tests.  In addition to the failed ruff and doctest checks, it looks like some of the standard tests might be hanging.  Not sure if that is due to the new code or if something is messed up in the GitHub actions.  I'll try to come back and take a look at that later.

In the meantime, some style things to adjust in your next pass.

MythasNauveili:
Thanks for the first feedback @murrayrm. I tried to fix the most in the two last commits. For the tests hanging, it might be because of debug plots that I did not disable in `dde_test.py` (this should be done now).

control/dde.py
MythasNauveili: removed in last commit

control/dde.py
MythasNauveili: removed in last commit

control/__init__.py
MythasNauveili: no, blank line is removed in the last commit

control/dde.py
MythasNauveili: underscore added

control/dde.py
MythasNauveili: underscore added

control/dde.py
MythasNauveili: underscore added

control/dde.py
MythasNauveili: I implemented the delayLTI case and added a test for the `input_output_response` that compares the results to forced response (it should be the same)
For the LTI responses functions, actually using them from a `delayLTI` system should work and call `dde_response`since the distiction is done directly in `step_response`, `forced_response`, etc ...

control/delaylti.py
MythasNauveili: I added headers to new files

control/delaylti.py
MythasNauveili: This is now moved to `InputOutputSystem`

murrayrm:
@MythasNauveili There are still two unit tests failing.  Click on the links above to see the issues.  The ruff checks should be straightforward.  For the doctest, it looks like something is printing out spurious information; not sure where that is happening.

MythasNauveili:
@murrayrm it might be ok now, I just removed a forgotten debug print in `timeresp.py` that was the source of the `dxdt = [...]`

PR #1181, opened 2025-09-09
BugFix: `place_acker` output should be 2D array

guptavaibhav0:
`place_acker` function's output is a 1D array, while it should be a 2D array as described in its doc string.


murrayrm:
Alternative fix in #1195 with unit tests to catch the bug.

PR #1184, opened 2025-11-04
Robust hinfsyn: add suboptimal synth

wuschel-brompf:
to obtain numerically more robust controllers, usually a slightly suboptimal controller is synthesized after running the first optimization.

### DIFF
--- a/control/robust.py
+++ b/control/robust.py
@@ -85,7 +85,7 @@ def h2syn(P, nmeas, ncon):
     return K
 
 
-def hinfsyn(P, nmeas, ncon, gamTry):
+def hinfsyn(P, nmeas, ncon, gamTry=None):
     # TODO: document significance of rcond
     """H-infinity control synthesis for plant P.
 


ilayn:
Note that this relaxation works typically OK with LMI based design since you introduce slackness but for Riccati based solvers there are no guarantees that this improves the conditioning of the controllers. This is typically due to the bilinear nature of the problem. It works OK for smooth-enough problems though.

control/robust.py
```diff
@@ -85,7 +85,7 @@ def h2syn(P, nmeas, ncon):
     return K
 
 
-def hinfsyn(P, nmeas, ncon):
+def hinfsyn(P, nmeas, ncon, gamTry=None):
```
murrayrm: This parameter name doesn't not follow python-control naming conventions.  From the [developer notes](https://python-control.readthedocs.io/en/0.10.2/develop.html#parameter-names):

> Use longer description parameter names that describe the action or role (e.g., trajectory_constraints and print_summary in [optimal.solve_optimal_trajectory](https://python-control.readthedocs.io/en/0.10.2/generated/control.optimal.solve_optimal_trajectory.html#control.optimal.solve_optimal_trajectory).

I suggest something like `target_gamma` for this parameter name.

murrayrm (changes requested):
Thanks for this contribution, @wuschel-brompf.  I had a few things that I think should be updated to be consistent with our coding conventions and to provide a bit more information to the user.

Also, we need a unit test that covers the new code.

murrayrm:
@wuschel-brompf: Will you be able to update the PR to address the earlier comments?

PR #1212, opened 2026-03-13
Fix bandwidth function for DT systems

aWZHY0yQH81uOYvH:
Frequency needs to be evaluated differently for discrete-time systems.

### DIFF
--- a/control/lti.py
+++ b/control/lti.py
@@ -211,8 +211,14 @@ class LTI(InputOutputSystem):
             # solve for the bandwidth, use scipy.optimize.root_scalar() to
             # solve using bisection
             import scipy
+
+            if self.isdtime(strict=True):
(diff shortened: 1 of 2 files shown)


sawyerbfuller:
Thanks for the submission. Looks like this was a bug, and your submission appears to correctly avoid problems that could arise in the optimization routine because the frequency response of discrete-time systems is periodic. 

Can you add a unit test? I think that all that would be needed would be a confirmation of correct behavior on a first-order system.

sawyerbfuller:
I think the failing unit tests are unrelated. Sometimes they go away on a re-run

aWZHY0yQH81uOYvH:
Ok, unit test added.

sawyerbfuller:
Looks good to me

PR #1216, opened 2026-05-15
Fix negative real handling and improve pole matching logic

### LINKED ISSUE #1204: RuntimeWarning in impulse_response for discrete-time systems with negative real poles
## Description

When computing impulse responses for discrete-time systems with negative real poles, `ct.impulse_response()` generates a `RuntimeWarning: divide by zero encountered in divide`. This occurs even though the system is valid and the results are correct.

## Minimal Reproducible Example
```python
import control as ct
import numpy as np

# Discrete-time system with poles at -1 and -2
b, a = [1, 3, 0], [1, 3, 2]
sys = ct.TransferFunction(b, a, dt=True)

# This triggers the warning
t, y = ct.impulse_response(sys, 5)
```

**Output:**
```
RuntimeWarning: divide by zero encountered in divide
  t_emp = np.max(log_decay_percent / np.abs((np.log(p_nr)/dt).real))
```

## Root Cause

The warning originates from the settling time calculation in the time response functions. For negative real poles in discrete-time systems:

- `np.log(-1) = 0 + πj` (real part is exactly 0)
- `np.log(-2) = 0.693 + πj` (real part is small)

The code attempts to compute settling time using:
```python
t_emp = np.max(log_decay_percent / np.abs((np.log(p_nr)/dt).real))
```

When the real part of `log(p_nr)` is zero or near-zero, this causes division by zero.

For discrete-time systems, stability and settling behavior should be based on pole magnitudes (`|pole| < 1`), not logarithms of negative numbers.

UbeenII:
2 fixes. 1st fix: the m_nr branch which is meant to handle negative reals takes the -1 root and uses the p_nr case causes t_emp to become infinite due to ln|p|=0 causing a division by zero and a RuntimeWarning. FIX exclude |p|=1  so they use the m_w case which uses cycle counting formula which is appropriate for unit circle poles.

2nd fix:  The m_int case used p.real -1 <sqrt_eps instead of np.abs (p.real-1)<sqrt_eps. Without abs any pole with Re(p) < 1 were matched rather than just poles near +1 causing stable poles like 0.95 to be discarded as discrete integrators. I updated the corresponding test to have the expected value to match the correct result.

Relaxed the T[-1] tolerance from 0.5 *ideal_dt to ideal_dt because discrete-time vectors must land on integer multiples of dt, so T[-1] can overshoot ideal_tfinal by up to one full sample period. 

Fixes #1204.

### DIFF
--- a/control/tests/timeresp_test.py
+++ b/control/tests/timeresp_test.py
@@ -2,10 +2,10 @@
 
 from copy import copy
 from math import isclose
+import warnings
 
 import numpy as np
 import pytest
-
 
 import control as ct
 from control import StateSpace, TransferFunction, c2d, isctime, ss2tf, tf2ss
@@ -14,7 +14,7 @@ from control.timeresp import _default_time_vector, _ideal_tfinal_and_dt, \
     forced_response, impulse_response, initial_response, step_info, \
     step_response
 
-import warnings
+
 
 
 class TSys:
--- a/control/tests/timeresp_test.py
+++ b/control/tests/timeresp_test.py
@@ -820,7 +820,7 @@ class TestTimeresp:
 
     @pytest.mark.parametrize(
         "tfsys, tfinal",
-        [(TransferFunction(1, [1, .5]), 25),        #  pole at 0.5
+        [(TransferFunction(1, [1, .5]), 13.81551),        #  pole at 0.5
          (TransferFunction(1, [1, .5]).sample(.1), 25),  # discrete pole at 0.5
          (TransferFunction(1, [1, .5, 0]), 25)])         # poles at 0.5 and 0
     def test_auto_generated_time_vector_tfinal(self, tfsys, tfinal):
--- a/control/tests/timeresp_test.py
+++ b/control/tests/timeresp_test.py
@@ -832,7 +832,7 @@ class TestTimeresp:
     
     def test_discrete_time_negative_one_settling(self):
         #system with -1 pole
-        TF = TransferFunction([1,3,0],[1,3,2], dt =True)
+        TF = TransferFunction([1,3,0],[1,3,2], dt=True)
         with warnings.catch_warnings():
             warnings.simplefilter("error")
             impulse_response(TF)
--- a/control/tests/timeresp_test.py
+++ b/control/tests/timeresp_test.py
@@ -6,12 +6,15 @@ from math import isclose
 import numpy as np
 import pytest
 
+
 import control as ct
 from control import StateSpace, TransferFunction, c2d, isctime, ss2tf, tf2ss
 from control.exception import pandas_check
 from control.timeresp import _default_time_vector, _ideal_tfinal_and_dt, \
     forced_response, impulse_response, initial_response, step_info, \
     step_response
+
+import warnings
 
 
 class TSys:
@@ -817,7 +820,7 @@ class TestTimeresp:
 
     @pytest.mark.parametrize(
         "tfsys, tfinal",
-        [(TransferFunction(1, [1, .5]), 13.81551),        #  pole at 0.5
+        [(TransferFunction(1, [1, .5]), 25),        #  pole at 0.5
          (TransferFunction(1, [1, .5]).sample(.1), 25),  # discrete pole at 0.5
          (TransferFunction(1, [1, .5, 0]), 25)])         # poles at 0.5 and 0
     def test_auto_generated_time_vector_tfinal(self, tfsys, tfinal):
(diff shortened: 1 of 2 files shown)


slivingston (changes requested):
Initial comments:

* Why are the commits associated with a GitHub user different from the user who is opening this PR? (Neither user account has a real name, so authorship is unclear.)
* A change that results in `_ideal_tfinal_and_dt()` returning a different tfinal is not ideal because it will change behavior of several public functions, e.g., `impule_response` when `timepts` arg is not given.
* Ideally, when fixing a bug, you have a test that reproduces it; i.e., a regression test.

UbeenII:
Sorry this is my first time contributing. I have fixed the dual author problem and I think GitHub now reflects the correct account now. I added an additional test to this case: `test_discrete_time_negative_one_settling` for #1204. The test creates a system with a pole at z=-1 and verifies that `impulse_response` no longer raises a `RuntimeWarning`. I have also reverted the `m_int` change to what it was originally.

UbeenII:
I realised the previous failure on the CI was due to an accidental change I made to one of the tests ( changing 13.8 to 25) I've restored it to the original main branch value (13.81551) and verified that the entire suite passes cleanly locally now. Ready for review whenever you get a chance

slivingston:
Thanks; I will review it tomorrow.

control/tests/timeresp_test.py
```diff
@@ -6,13 +6,16 @@
...
 
+
 import control as ct
 from control import StateSpace, TransferFunction, c2d, isctime, ss2tf, tf2ss
 from control.exception import pandas_check
 from control.timeresp import _default_time_vector, _ideal_tfinal_and_dt, \
     forced_response, impulse_response, initial_response, step_info, \
     step_response
 
+import warnings
```
slivingston: Move this `warnings` import above, so it is near other standard library imports. The pattern is to put imports from the standard library together, then imports of other packages below that, and finally imports of the current package. E.g., follow the pattern of https://github.com/python-control/python-control/blob/17baf62da1ba3902a615c322546c3e40d801452f/control/tests/ctrlplot_test.py#L4-L6

control/tests/timeresp_test.py
```diff
@@ -6,13 +6,16 @@
 import numpy as np
 import pytest
 
+
```
slivingston: Do not add a blank line here. Preferred style is 1 blank line between groups of imports.

control/tests/timeresp_test.py
```diff
@@ -826,6 +829,14 @@ def test_auto_generated_time_vector_tfinal(self, tfsys, tfinal):
         np.testing.assert_allclose(ideal_tfinal, tfinal, rtol=1e-4)
         T = _default_time_vector(tfsys)
         np.testing.assert_allclose(T[-1], tfinal, atol=0.5*ideal_dt)
+    
+    def test_discrete_time_negative_one_settling(self):
+        #system with -1 pole
+        TF = TransferFunction([1,3,0],[1,3,2], dt =True)
```
slivingston:
```suggestion
        TF = TransferFunction([1,3,0],[1,3,2], dt=True)
```

control/timeresp.py
```diff
@@ -2172,12 +2172,13 @@ def _ideal_tfinal_and_dt(sys, is_step=True):
         m_z = np.abs(p) < sqrt_eps
         p = p[~m_z]
         # Negative reals- treated as oscillatory mode
-        m_nr = (p.real < 0) & (np.abs(p.imag) < sqrt_eps)
+        m_nr = (p.real < 0) & (np.abs(p.imag) < sqrt_eps) & (np.abs(p.real+1)>sqrt_eps)
```
slivingston:
```suggestion
        m_nr = (p.real < 0) & (np.abs(p.imag) < sqrt_eps) & (np.abs(p.real+1) > sqrt_eps)
```

control/timeresp.py
```diff
@@ -2172,12 +2172,13 @@ def _ideal_tfinal_and_dt(sys, is_step=True):
...
         p_nr, p = p[m_nr], p[~m_nr]
         if p_nr.size > 0:
             t_emp = np.max(log_decay_percent / np.abs((np.log(p_nr)/dt).real))
             tfinal = max(tfinal, t_emp)
         # discrete integrators
+
```
slivingston: Do not insert a blank line here.

slivingston (changes requested):
Thanks! I will merge this after you consider my formatting comments.

slivingston (approved):

control/tests/timeresp_test.py
```diff
@@ -14,6 +15,8 @@
     step_response
 
 
+
+
```
slivingston, line 19: I will remove these extra blank lines before merging. (Normally, only up to 2 blank lines are used to separate top-level blocks in a Python file.)

slivingston:
This pull request was merged via https://github.com/python-control/python-control/commit/7d5019e0b31f7f9f1099808d15cedfbaf01a0a56. Thanks!

PR #1219, opened 2026-06-05
docs: clarify lqe and dlqe error covariance wording

### LINKED ISSUE #1183: Inaccurate wording in documentation of lqe and dlqe
In the documentation of lqe and dlqe, it states:

> produces a state estimate x_e that minimizes the expected squared error using the sensor measurements y.

From my understanding of theory, this is quite inaccurate/imprecise.
A better wording would be:
> produces a state estimate x_e that minimizes the generalized variance of x - x_e using the sensor measurements y.

Source: 
https://en.wikipedia.org/wiki/Generalized_variance
https://arxiv.org/pdf/2103.07275

Another wording would be that it minimizes the trace of the covariance matrix x - x_e,
but thinking in a generalized variance is more catchy.

marko1olo:
Closes #1183.

This updates the `lqe` and `dlqe` docstrings to describe the steady-state estimation error covariance minimized by the estimator, instead of saying only that the state estimate minimizes expected squared error.

I avoided adding a runtime behavior change or changing generated docs directly.

Validation:
- `python -m ruff check control/stochsys.py`
- `python -m pytest control/tests/stochsys_test.py -q` -> 25 passed, 2 skipped
- `python -m numpydoc render control.stochsys.lqe`
- `python -m numpydoc render control.stochsys.dlqe`
- `$env:PYTHONPATH='..'; python -m pytest test_sphinxdocs.py::test_config_defaults -q` -> 1 passed, 4 warnings

Known unrelated local docs issues:
- `python -m numpydoc validate control.stochsys.lqe/control.stochsys.dlqe` reports pre-existing docstring style findings.
- The full Sphinx docs test file needs generated docs/CWD setup that is not present in this fresh clone.

---


murrayrm:
I don't think that the new wording is accurate.  In particular, it uses the term "minimum covariance", but the covariance is (in general) a matrix and so the minimum depends on the norm that you use (eg, determinant or trace).  As noted in #1183, I think that something like "minimizes the mean squared error" or perhaps "minimizes the mean of x - x_e" would be more accurate.

marko1olo:
Thanks, updated this to avoid describing covariance as the scalar objective.

`lqe` and `dlqe` now say the estimator minimizes the mean squared estimation error (`x - x_e` / `x[n] - x_e[n]`) using the measurements `y`.

Local checks:
- `uv run --no-project --with ruff ruff check control/stochsys.py`
- `python -m compileall -q control\stochsys.py`
- `git diff --check`

marko1olo:
Thanks, updated this to avoid describing covariance as the scalar objective.

`lqe` and `dlqe` now say the estimator minimizes the mean squared estimation error (`x - x_e` / `x[n] - x_e[n]`) using the measurements `y`.

Local checks:
- `uv run --no-project --with ruff ruff check control/stochsys.py`
- `python -m compileall -q control\stochsys.py`
- `git diff --check`

murrayrm:
Comit 6666024 looks fine, but commit 90764ad  seems to have introduced a lot of extraneous changes that should not be part of this PR.

Can you revert 90764ad.  You may also need to wait until PR #1243 is merged into main and then rebase if unit tests don't pass.

PR #1225, opened 2026-06-07
fix(timeresp): avoid warning for discrete -1 poles

### LINKED ISSUE #1204: RuntimeWarning in impulse_response for discrete-time systems with negative real poles

marko1olo:
Fixes python-control/python-control#1204.

`_ideal_tfinal_and_dt()` already treats negative real discrete poles separately, but a pole exactly on the unit circle at `-1` was still sent through the decaying negative-real path. That path divides by `real(log(pole))`, which is zero for `-1`, so `impulse_response()` emitted a `RuntimeWarning` even though the response itself was valid.

This change keeps decaying negative real poles such as `-0.5` on the existing decay-based path, keeps positive real discrete poles on the existing padded path, and lets unit-circle negative real poles flow to the pure oscillatory-mode estimate.

Verification:

- Reproduced the issue script with `warnings.simplefilter("error", RuntimeWarning)` before the fix.
- `python -m pytest control\tests\timeresp_test.py::TestTimeresp::test_auto_generated_time_vector_tfinal control\tests\timeresp_test.py::TestTimeresp::test_discrete_negative_real_pole_tfinal control\tests\timeresp_test.py::TestTimeresp::test_discrete_time_impulse_negative_poles_no_warning -q`
- `python -m pytest control\tests\timeresp_test.py -q`
- `python -m ruff check --no-cache control\timeresp.py control\tests\timeresp_test.py`
- `PYTHONPYCACHEPREFIX=C:\tmp\pycache-python-control-1204 python -m compileall -q control\timeresp.py control\tests\timeresp_test.py`
- `git diff --check`

AI-assisted contribution: implemented with OpenAI Codex and manually verified with the commands above.

### DIFF
--- a/control/tests/timeresp_test.py
+++ b/control/tests/timeresp_test.py
@@ -1,5 +1,6 @@
 """timeresp_test.py - test time response functions"""
 
+import warnings
 from copy import copy
 from math import isclose
 
@@ -582,6 +583,17 @@ class TestTimeresp:
 
         np.testing.assert_array_equal(response.inputs,Uexpected)
 
+    def test_discrete_time_impulse_negative_poles_no_warning(self):
+        # A negative pole on the unit circle should be treated as oscillatory.
+        sysd = ct.TransferFunction([1, 3, 0], [1, 3, 2], dt=True)
+
+        with warnings.catch_warnings():
+            warnings.simplefilter("error", RuntimeWarning)
+            response = ct.impulse_response(sysd, 5)
+
+        np.testing.assert_allclose(response.time, np.arange(6))
+        assert np.all(np.isfinite(response.outputs))
+
     @pytest.mark.parametrize("tsystem", ["siso_ss1"], indirect=True)
     def test_impulse_response_warnD(self, tsystem):
         """Test warning about direct feedthrough"""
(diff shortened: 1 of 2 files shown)


slivingston:
@marko1olo There is already an open PR that is supposed to fix #1204 : https://github.com/python-control/python-control/pull/1216

I will re-open this PR if the other one is not sufficient.

PR #1227, opened 2026-06-07
Stabilize LQE covariance symmetry checks

### LINKED ISSUE #1174: better symmetry check test
Hi,

While porting our MATLAB code to Python using the control library, we noticed that the current symmetry check:

```
def _is_symmetric(M):
    M = np.atleast_2d(M)
    if isinstance(M[0, 0], inexact):
        eps = finfo(M.dtype).eps
        return ((M - M.T) < eps).all()
    else:
        return (M == M.T).all()
```

has some limitations:

1. It performs elementwise comparison without using abs(), so small negative differences (e.g., from roundoff) can cause false failures. 
2. It uses a fixed eps, which is stricter than MATLAB's behavior and doesn’t scale with matrix magnitude.

Would it be possible to adopt a more numerically robust check, similar to what SciPy uses?

```
for ind, mat in enumerate((q, r)):
    if norm(mat - mat.conj().T, 1) > np.spacing(norm(mat, 1)) * 100:
        raise ValueError(f"Matrix {'qr'[ind]} should be symmetric/hermitian.")
```

This approach allows for small, scale-aware numerical asymmetries and works well for both real symmetric and complex Hermitian matrices.

Prithvi

marko1olo:
## Summary

Fixes #1174.

This makes the Riccati symmetry checks more robust and avoids false `QN must be a symmetric matrix` errors when `lqe()` / `dlqe()` form the projected process covariance `G @ QN @ G.T`.

## Details

- Replace the local floating-point symmetry check with SciPy's `issymmetric` / `ishermitian` using a small scale-aware tolerance.
- Validate user-supplied `QN` as a symmetric covariance matrix before projection.
- Symmetrize only the projected covariance `G @ QN @ G.T`, which is mathematically symmetric but can pick up roundoff-level asymmetry.
- Add deterministic regressions for both `lqe(..., method="scipy")` and `dlqe(..., method="scipy")`.
- Add a guard test showing clearly asymmetric user `QN` is still rejected instead of silently repaired.

## Local validation

- Red proof: using the regression matrices without the projected-covariance symmetrization makes `care()` and `dare()` reject `G @ QN @ G.T` with `ControlArgument: QN must be a symmetric matrix`; the measured projected asymmetry was `4.44e-16`.
- `python -m pytest -p no:cacheprovider control\tests\mateqn_test.py::TestMatrixEquations::test_symmetric_shape_check_tolerance -q`
- `python -m pytest -p no:cacheprovider control\tests\stochsys_test.py::test_lqe_symmetrizes_projected_process_covariance control\tests\stochsys_test.py::test_dlqe_symmetrizes_projected_process_covariance control\tests\stochsys_test.py::test_lqe_rejects_user_supplied_asymmetric_covariance -q`
- `python -m pytest -p no:cacheprovider control\tests\mateqn_test.py -q` -> 11 passed, 13 skipped (`slycot` not installed)
- `python -m pytest -p no:cacheprovider control\tests\stochsys_test.py -q` -> 29 passed, 2 skipped (`slycot` not installed)
- `python -m ruff check --no-cache control\mateqn.py control\stochsys.py control\tests\mateqn_test.py control\tests\stochsys_test.py`
- `PYTHONPYCACHEPREFIX=C:\tmp\pycache-python-control-1174 python -m compileall -q control\mateqn.py control\stochsys.py control\tests\mateqn_test.py control\tests\stochsys_test.py`
- `git diff --cached --check`

AI-assisted contribution disclosure: implementation prepared with OpenAI Codex and reviewed/tested locally before submission.

---

### DIFF
--- a/control/mateqn.py
+++ b/control/mateqn.py
@@ -653,7 +653,9 @@ def _check_shape(M, n, m, square=False, symmetric=False, name="??"):
 def _is_symmetric(M):
     M = np.atleast_2d(M)
     if isinstance(M[0, 0], inexact):
-        eps = finfo(M.dtype).eps
-        return ((M - M.T) < eps).all()
+        eps = finfo(M.real.dtype if np.iscomplexobj(M) else M.dtype).eps
+        if np.iscomplexobj(M):
+            return sp.linalg.ishermitian(M, atol=100 * eps, rtol=100 * eps)
+        return sp.linalg.issymmetric(M, atol=100 * eps, rtol=100 * eps)
     else:
         return (M == M.T).all()
--- a/control/mateqn.py
+++ b/control/mateqn.py
@@ -788,7 +788,9 @@ def _check_shape(M, n, m, square=False, symmetric=False, name="??"):
 def _is_symmetric(M):
     M = np.atleast_2d(M)
     if isinstance(M[0, 0], inexact):
-        eps = finfo(M.dtype).eps
-        return ((M - M.T) < eps).all()
+        eps = finfo(M.real.dtype if np.iscomplexobj(M) else M.dtype).eps
+        if np.iscomplexobj(M):
+            return sp.linalg.ishermitian(M, atol=100 * eps, rtol=100 * eps)
+        return sp.linalg.issymmetric(M, atol=100 * eps, rtol=100 * eps)
     else:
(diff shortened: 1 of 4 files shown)


murrayrm:

PR #1229, opened 2026-06-07
Fix forced_response time inference for transposed inputs

### LINKED ISSUE #931: forced_response transpose bug
If `transpose `is true, then `U` should be transposed before getting `n_steps `from  `U.shape[1]`

marko1olo:
## Summary

Fixes #931.

When `forced_response()` is called for a discrete-time system with `timepts=None`, it infers the generated time vector length from `inputs`. With `transpose=True`, 2D input data is supplied in `(time, input)` order, but the current code still reads `U.shape[1]`, treating the input count as the number of time steps.

This changes the inference to use `U.shape[0]` for 2D inputs when `transpose=True`, matching the later `_check_convert_array(..., transpose=True)` conversion.

## Validation

- Red repro before the fix: `forced_response(sys, inputs=U, transpose=True)` with `U.shape == (5, 2)` failed with `ValueError: Parameter U: Wrong shape ... Expected: (2, 2)`.
- `python -m pytest control/tests/timeresp_test.py -k forced_response_transpose_without_time_vector -q` passed.
- `python -m pytest control/tests/timeresp_test.py -q -p no:cacheprovider` passed: 253 passed, 46 skipped.
- `python -m ruff check --no-cache control\timeresp.py control\tests\timeresp_test.py` passed.
- `PYTHONPYCACHEPREFIX=C:\tmp\pycache-python-control-931 python -m compileall -q control\timeresp.py control\tests\timeresp_test.py` passed.
- `git diff --check` passed.

---

### DIFF
--- a/control/timeresp.py
+++ b/control/timeresp.py
@@ -1112,7 +1112,7 @@ def forced_response(
                 raise ValueError('Parameters `T` and `U` can\'t both be '
                                  'zero for discrete-time simulation')
             # Set T to equally spaced samples with same length as U
-            if U.ndim == 1:
+            if U.ndim == 1 or transpose:
                 n_steps = U.shape[0]
             else:
                 n_steps = U.shape[1]
--- a/control/tests/timeresp_test.py
+++ b/control/tests/timeresp_test.py
@@ -1245,6 +1245,17 @@ class TestTimeresp:
         assert y.shape == ysh_no
         assert x.shape == (T.size, sys.nstates)
 
+    def test_forced_response_transpose_without_time_vector(self):
+        sys = ct.ss(
+            [[0.5]], [[1.0, -0.25]], [[1.0]], [[0.0, 0.0]], dt=True)
+        U = np.column_stack((np.arange(5.0), np.arange(5.0) + 10))
+
+        response = ct.forced_response(sys, inputs=U, transpose=True)
+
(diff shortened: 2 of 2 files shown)


control/timeresp.py
```diff
@@ -1114,6 +1114,8 @@ def forced_response(
             # Set T to equally spaced samples with same length as U
             if U.ndim == 1:
                 n_steps = U.shape[0]
+            elif transpose:
```
slivingston: Why add another branch here instead of `or` with the above? I.e.,
```python
if U.ndim == 1 or transpose:
```

marko1olo: Thanks, good point. I simplified this in `d12d4896` to use `if U.ndim == 1 or transpose:` since both cases use `U.shape[0]`.

Rechecked locally:

- `python -m pytest control\tests\timeresp_test.py::TestTimeresp::test_forced_response_transpose_without_time_vector -q`
- `python -m pytest control\tests\timeresp_test.py -q`
- `python -m ruff check control\timeresp.py control\tests\timeresp_test.py`
- `python -m compileall -q control\timeresp.py control\tests\timeresp_test.py`
- `git diff --check`

murrayrm:

control/timeresp.py
marko1olo: Good catch — you're right, the `or` is cleaner. Updated.

Issue #1057, opened 2024-11-12
Slycot license is GPL whereas python-control license is BSD.


ubaldot:
It looks like that python-control (BSD license) relies on slycot (GPL license).  Shouldn't python-control be released under GPL license as well?

ubaldot:
Ok, I may self-answer my question: slycot is not necessary (but it is desirable). Python-control can run without slycot for many use-cases. If your work includes such use-cases, then you can consider your work bsd license compliant. 
If you are using python-control functionalities that require slycot, then your work must adhere to gpl licensing. Make sense?

slivingston:
Your [self-answer](https://github.com/python-control/python-control/issues/1057#issuecomment-2471621199) is correct, but I am not a lawyer.

Note the section [about license in the Slycot README](https://github.com/python-control/Slycot/blob/c47b63fdc8b3e7d05f8d28a3207ddb857b846489/README.rst#license): it is possible to switch to BSD-3-Clause if we get explicit permission from all authors.

slivingston:
Sadly, interpreting the relevance/consequences of GPL can be complicated. Notes:

* Slycot is under the GPLv2 (not the current GPLv3). The Linux kernel also has license GPLv2 and is famous/infamous for being lax regarding licenses of kernel modules.
* Running the software as a service (a.k.a. users over a network) does not imply the need to distribute changes. For that requirement, you need a license like the GNU Affero GPL.
* If Slycot with GPL is a significant difficulty in your work, consider the task of getting permissions so we can switch Slycot to BSD-3-Clause. If you do this, be sure to confirm how to do it correctly (legally). There are other projects that have changed licenses before, so there is precedent.

ubaldot:
Ok, I see, thanks. 

The point is that private organizations prefer to stay on the safe side and generally the utilization of open-source software is forbidden. However, some permissive companies allow the solely utilization of MIT-like licenses open source software, and typically utilization of GPL license software is blacklisted. In python-control specific case, the situation is a bit on the grey zone, since the statements around slycot could be further clarified. Such uncertainty would be enough to forbid employees to use python-control package. That would be a pity IMO. 

Perhaps it could be good to add some clearer and sharp notes explaining when the gpl constraints apply and how to avoid fall into that use-case in the README file? 

I am thinking to further stress this aspect through a statement like the following: "If you are not allowed to use GPL licensed software just don't install slycot.
[...]

Issue #36, opened 2014-08-17
Extend dare to handle S != 0, E != I case correctly (using scipy?)


murrayrm:
From commit 7cf9630146a961c28c38817d48794525d22c6287:

> Note: the scipy implementation handles only the case S = 0, E = I, the default values.  If S and E are specified, the old routine (using slycot) is called.  This passes the existing tests, but the tests include only one simple case, so it would be good to test this more extensively.

@robertobucher suggests the following fix:

``` python
# Check dimensions for consistency
    nstates = B.shape[0];
    ninputs = B.shape[1];
    if (A.shape[0] != nstates or A.shape[1] != nstates):
        raise ControlDimension("inconsistent system dimensions")

    elif (Q.shape[0] != nstates or Q.shape[1] != nstates or
          R.shape[0] != ninputs or R.shape[1] != ninputs) :
        raise ControlDimension("incorrect weighting matrix dimensions")

    X,rcond,w,S,T = \
        sb02od(nstates, ninputs, A, B, Q, R, 'D');

    return X
```

This doesn't seem to directly deal with the S and E arguments, so need to figure that part out.

ilayn:
Just a reminder: SciPy v0.19+ can handle S and E terms with its [care ](http://scipy.github.io/devdocs/generated/scipy.linalg.solve_continuous_are.html#scipy.linalg.solve_continuous_are) and [dare ](http://scipy.github.io/devdocs/generated/scipy.linalg.solve_discrete_are.html#scipy.linalg.solve_discrete_are) functions with consistency checks.

PR #1234, opened 2026-06-15
Add scipy fallbacks for generalized Lyapunov and discrete Sylvester equations

kwlee2025cpp:
## Summary

`lyap` and `dlyap` currently raise `ControlArgument("method='scipy' not
valid for ...")` for three problems, so solving them requires the optional
`slycot` package:

- the generalized continuous Lyapunov equation `A X Eᵀ + E X Aᵀ + Q = 0`,
- the generalized discrete Lyapunov equation `A X Aᵀ − E X Eᵀ + Q = 0`,
- the discrete-time Sylvester equation `A X Qᵀ − X + C = 0`.

This PR implements pure scipy/numpy fallbacks for all three, so they work
when `slycot` is not installed. When `slycot` *is* installed it remains the
default (`method='slycot'`), and the new tests cross-validate the scipy
results against it.

## What it adds

**Generalized Lyapunov (continuous and discrete).** The equation is reduced
to a standard Lyapunov equation by a congruence transform with `inv(E)`
(`Ã = E⁻¹A`, `Q̃ = E⁻¹Q E⁻ᵀ`, leaving `X` unchanged), then solved with
`scipy.linalg.solve_{continuous,discrete}_lyapunov`. This requires `E` to be
nonsingular. SLICOT `SG03AD` (`method='slycot'`), based on the generalized
Schur method of Penzl, additionally handles singular `E`; that case raises a
clear `ControlArgument` pointing the user to `method='slycot'`.

**Discrete-time Sylvester.** Solved by the Bartels–Stewart method: complex
Schur factorizations of `A` and `Qᵀ`, then a column-by-column back
substitution. This matches the `O(n³ + m³)` complexity of the
Hessenberg–Schur algorithm in SLICOT `SB04QD` (`method='slycot'`).
Singularity (a pair of eigenvalues with `λᵢ·λⱼ ≈ 1`) is detected and raises
a clear `ControlArgument`.

**Ill-conditioned `E`.** Because the generalized-Lyapunov fallback inverts
`E`, accuracy degrades when `E` is poorly conditioned. A `UserWarning` is
now issued when `cond(E) > 1/√eps`, recommending `method='slycot'`. (This
also surfaces a case in the continuous path that was previously silently
inaccurate.)

## Fixes included

- **Import-alias bug** in the optional-`slycot` guards: the `except
  ImportError` branches bound the sentinels under the wrong names
  (`sb0qmd` for `sb04qd`, `sb04ad` for `sg03ad`), so on a slycot-less
  install the intended `None` fallbacks were never set under the names the
  code references.
- **Stale `dlyap` docstring** note (implied a Kronecker-product method);
  corrected to Bartels–Stewart, with Golub–Nash–Van Loan and
  Bartels–Stewart references added to the `Notes`/`References`.

## Testing

- scipy↔slycot cross-validation added to `test_lyap_g`, `test_dlyap_g`, and
  `test_dlyap_sylvester` (scipy result matches slycot to `~1e-9…1e-16`).
- New `test_lyap_g_ill_conditioned_E_scipy` (parametrized over `lyap`/`dlyap`)
  asserts the ill-conditioned-`E` `UserWarning`.
- New `test_dlyap_sylvester_singular_scipy` asserts the singular-pencil
  `ControlArgument`.
- Green both with and without slycot: `19 passed, 12 skipped` (no slycot) /
  `31 passed` (slycot).

Only `control/mateqn.py` and `control/tests/mateqn_test.py` are touched.

## Scope

This is an intentionally small, self-contained slice. It is the first of a
short series factoring a larger "scipy fallbacks for slycot-gated routines"
effort into independently reviewable PRs; the matrix-equation solvers go
first because they are foundational and easy to validate against SLICOT.
`gram` is the planned next slice.

## Alternatives considered

The same `inv(E)` reduction is used by pyMOR's dense scipy Lyapunov
backend. Factoring the pencil `(A, E)` with `scipy.linalg.qz` instead of
inverting `E` (as `harold` does) is more robust for ill-conditioned `E`
and is a natural follow-up. Truly singular `E` (descriptor systems with
infinite generalized eigenvalues) is out of scope here and continues to
require `method='slycot'`; a pure-scipy path for that case would need
spectral-projector deflation of the infinite spectrum.

### DIFF
--- a/control/mateqn.py
+++ b/control/mateqn.py
@@ -58,21 +58,23 @@ __all__ = ['lyap', 'dlyap', 'dare', 'care']
 
 
 def _warn_ill_conditioned_E(E):
-    """Warn that the inv(E) congruence transform will lose accuracy.
+    """Warn that an ill-conditioned E costs accuracy.
 
     The scipy generalized-Lyapunov fallback reduces the problem to a
     standard Lyapunov equation by inverting E, so a poorly conditioned E
     costs accuracy (continuous and discrete paths alike, regardless of
-    whether the underlying scipy solve happens to warn).  SLICOT sg03ad
-    (method='slycot') avoids inverting E and is preferable in that case.
+    whether the underlying scipy solve happens to warn).  The generalized
+    Lyapunov problem is itself ill-conditioned (about cond(E)**2) when E
+    is, so method='slycot', though it does not form inv(E) explicitly, is
+    not measurably more accurate in that regime.
     """
     condE = np.linalg.cond(E)
     if condE > 1.0 / np.sqrt(finfo(float).eps):
         warnings.warn(
-            f"E is ill-conditioned (cond(E) = {condE:.2g}); the "
-            "method='scipy' generalized Lyapunov solution may have reduced "
-            "accuracy.  Use method='slycot' (SLICOT sg03ad) for a more "
-            "robust solution.", UserWarning, stacklevel=3)
+            f"E is ill-conditioned (cond(E) = {condE:.2g}); the generalized "
+            "Lyapunov solution may have reduced accuracy.  The problem itself "
+            "is ill-conditioned for such E, so method='slycot' is not "
+            "measurably more accurate.", UserWarning, stacklevel=3)
 
 #
 # Lyapunov equation solvers lyap and dlyap
@@ -129,8 +131,10 @@ def lyap(A, Q, C=None, E=None, method=None):
     With method='scipy', the equation is transformed to a standard
     Lyapunov equation by inverting E, which requires E to be nonsingular
     and loses accuracy when E is ill-conditioned (a UserWarning is then
-    issued); method='slycot' does not invert E and is preferable in that
-    case.  Both methods require E nonsingular; a truly singular
+    issued).  The generalized Lyapunov problem is itself ill-conditioned
+    (about cond(E)**2) when E is, so method='slycot', though it does not
+    invert E, is not measurably more accurate in that case.  Both methods
+    require E nonsingular; a truly singular
     (descriptor) E is not currently handled by either.
 
     References
@@ -289,8 +293,10 @@ def dlyap(A, Q, C=None, E=None, method=None):
     With method='scipy', the equation is transformed to a standard
     Lyapunov equation by inverting E, which requires E to be nonsingular
     and loses accuracy when E is ill-conditioned (a UserWarning is then
-    issued); method='slycot' does not invert E and is preferable in that
-    case.  Both methods require E nonsingular; a truly singular
+    issued).  The generalized Lyapunov problem is itself ill-conditioned
+    (about cond(E)**2) when E is, so method='slycot', though it does not
+    invert E, is not measurably more accurate in that case.  Both methods
+    require E nonsingular; a truly singular
     (descriptor) E is not currently handled by either.
 
     For the Sylvester equation, method='slycot' uses the
--- a/control/mateqn.py
+++ b/control/mateqn.py
@@ -125,11 +125,13 @@ def lyap(A, Q, C=None, E=None, method=None):
     -----
     For the generalized Lyapunov equation, method='slycot' uses the
     SLICOT routine SG03AD, based on the generalized Schur method of
-    Penzl [1]_, which also handles singular E.  With method='scipy', the
-    equation is transformed to a standard Lyapunov equation by inverting
-    E, which requires E to be nonsingular and loses accuracy when E is
-    ill-conditioned (a UserWarning is then issued); method='slycot' does
-    not invert E and is preferable in that case.
+    Penzl [1]_, which factors the matrix pencil without inverting E.
+    With method='scipy', the equation is transformed to a standard
+    Lyapunov equation by inverting E, which requires E to be nonsingular
+    and loses accuracy when E is ill-conditioned (a UserWarning is then
+    issued); method='slycot' does not invert E and is preferable in that
+    case.  Both methods require E nonsingular; a truly singular
+    (descriptor) E is not currently handled by either.
 
     References
     ----------
@@ -200,16 +202,18 @@ def lyap(A, Q, C=None, E=None, method=None):
             #
             #     (E^-1 A) X + X (E^-1 A)^T + E^-1 Q E^-T = 0
             #
-            # This requires E to be nonsingular; the SLICOT routine
-            # SG03AD used by method='slycot' (based on the generalized
-            # Schur method of Penzl (1998)) also handles singular E.
+            # This requires E to be nonsingular.  SG03AD (method='slycot',
+            # Penzl's generalized Schur method) factors the pencil without
+            # inverting E, but a truly singular E is not handled by either
+            # method.
             try:
                 At = solve(E, A)
                 Qt = solve(E, solve(E, Q).T).T
             except np.linalg.LinAlgError:
                 raise ControlArgument(
                     "method='scipy' requires E to be nonsingular; "
-                    "use method='slycot' (SLICOT sg03ad) for singular E")
+                    "a truly singular E (descriptor system) is not "
+                    "supported by either method")
             _warn_ill_conditioned_E(E)
             return sp.linalg.solve_continuous_lyapunov(At, -Qt)
 
@@ -281,11 +285,13 @@ def dlyap(A, Q, C=None, E=None, method=None):
     -----
     For the generalized Lyapunov equation, method='slycot' uses the
     SLICOT routine SG03AD, based on the generalized Schur method of
-    Penzl [1]_, which also handles singular E.  With method='scipy', the
-    equation is transformed to a standard Lyapunov equation by inverting
-    E, which requires E to be nonsingular and loses accuracy when E is
-    ill-conditioned (a UserWarning is then issued); method='slycot' does
-    not invert E and is preferable in that case.
+    Penzl [1]_, which factors the matrix pencil without inverting E.
+    With method='scipy', the equation is transformed to a standard
+    Lyapunov equation by inverting E, which requires E to be nonsingular
+    and loses accuracy when E is ill-conditioned (a UserWarning is then
+    issued); method='slycot' does not invert E and is preferable in that
+    case.  Both methods require E nonsingular; a truly singular
+    (descriptor) E is not currently handled by either.
 
     For the Sylvester equation, method='slycot' uses the
     Hessenberg-Schur method of the SLICOT routine SB04QD [2]_ and
@@ -403,16 +409,18 @@ def dlyap(A, Q, C=None, E=None, method=None):
             #
             #     (E^-1 A) X (E^-1 A)^T - X + E^-1 Q E^-T = 0
             #
-            # This requires E to be nonsingular; the SLICOT routine
-            # SG03AD used by method='slycot' (based on the generalized
-            # Schur method of Penzl (1998)) also handles singular E.
+            # This requires E to be nonsingular.  SG03AD (method='slycot',
+            # Penzl's generalized Schur method) factors the pencil without
+            # inverting E, but a truly singular E is not handled by either
+            # method.
             try:
                 At = solve(E, A)
                 Qt = solve(E, solve(E, Q).T).T
             except np.linalg.LinAlgError:
                 raise ControlArgument(
                     "method='scipy' requires E to be nonsingular; "
-                    "use method='slycot' (SLICOT sg03ad) for singular E")
+                    "a truly singular E (descriptor system) is not "
+                    "supported by either method")
             _warn_ill_conditioned_E(E)
             return sp.linalg.solve_discrete_lyapunov(At, Qt)
 
--- a/control/mateqn.py
+++ b/control/mateqn.py
@@ -47,14 +47,32 @@ except ImportError:
 try:
     from slycot import sb04qd
 except ImportError:
-    sb0qmd = None
+    sb04qd = None
 
 try:
     from slycot import sg03ad
 except ImportError:
-    sb04ad = None
+    sg03ad = None
 
 __all__ = ['lyap', 'dlyap', 'dare', 'care']
+
+
+def _warn_ill_conditioned_E(E):
+    """Warn that the inv(E) congruence transform will lose accuracy.
+
+    The scipy generalized-Lyapunov fallback reduces the problem to a
+    standard Lyapunov equation by inverting E, so a poorly conditioned E
+    costs accuracy (continuous and discrete paths alike, regardless of
+    whether the underlying scipy solve happens to warn).  SLICOT sg03ad
+    (method='slycot') avoids inverting E and is preferable in that case.
+    """
+    condE = np.linalg.cond(E)
+    if condE > 1.0 / np.sqrt(finfo(float).eps):
+        warnings.warn(
+            f"E is ill-conditioned (cond(E) = {condE:.2g}); the "
+            "method='scipy' generalized Lyapunov solution may have reduced "
+            "accuracy.  Use method='slycot' (SLICOT sg03ad) for a more "
+            "robust solution.", UserWarning, stacklevel=3)
 
 #
 # Lyapunov equation solvers lyap and dlyap
@@ -102,6 +120,21 @@ def lyap(A, Q, C=None, E=None, method=None):
     -------
     X : 2D array
         Solution to the Lyapunov or Sylvester equation.
+
+    Notes
+    -----
+    For the generalized Lyapunov equation, method='slycot' uses the
+    SLICOT routine SG03AD, based on the generalized Schur method of
+    Penzl [1]_, which also handles singular E.  With method='scipy', the
+    equation is transformed to a standard Lyapunov equation by inverting
+    E, which requires E to be nonsingular and loses accuracy when E is
+    ill-conditioned (a UserWarning is then issued); method='slycot' does
+    not invert E and is preferable in that case.
+
+    References
+    ----------
+    .. [1] Penzl, T., "Numerical solution of generalized Lyapunov
+       equations", Advances in Computational Mathematics, 8:33-48, 1998.
 
     """
     # Decide what method to use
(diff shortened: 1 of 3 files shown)


slivingston:
@kwlee2025cpp Thanks for proposing these changes. Can you answer a few questions to help with review and consideration for merging?

1. Is there a prior [issue](https://github.com/python-control/python-control/issues) or [discussion](https://github.com/python-control/python-control/discussions) that provides context and motivation for this addition?
2. If the feature is already provided when `slycot` is installed, is it worth the mantenance effort of an alternative implementation that does not require it? For example, how does the performance compare between this PR (the "fallbacks") and those requiring `slycot`?
3. Are you using this in your own work and now want to share the result as open source?

I am a little concerned that this PR is just a Claude-generated thing that no one is asking for. I am also concerned we are about to get many more Claude-generated PRs as your description suggestions ("...It is the first of a
short series factoring a larger 'scipy fallbacks for slycot-gated routines' effort into independently reviewable PRs..."). I am happy to be mistaken about that, so please let me know.

bnavigator:
Approving the workflows because the code on first glance looks good and obviously not malicious. No assessment about scientific merit, I fully support @slivingston's inquiry.

kwlee2025cpp:
Thanks @slivingston, and @bnavigator — these are the right questions, and I would rather answer them squarely than let this sit as an unmotivated patch.

**1. Prior issue / motivation.** It is the project's own direction. In #27 ("Move
away from SLICOT?") @murrayrm's 2014 plan was to "replace functions that require SLICOT with python versions … and if there is some advanced function that really requires SLICOT, we just leave that in as something that requires an optional package." That issue is closed-completed, but the pattern is in the tree today — `lqr` / `dlqr` / `dare` already carry this `method=` slycot-or-scipy dispatch — and the coverage is still openly requested (#36 asks the same for `dare`; #1057 raises the GPL/BSD question). So this extends an accepted line, not a new one.

One more disclosure, a personal thread: in 2017 (as @beachdweller) I was on #27 thinking about translating the SLICOT Fortran myself — the same year #133 added a no-Slycot test run. That effort and #27's pointed the same way but never met; I wish I had carried it through then instead of going quiet — belated apologies. The viable road was not translation but these scipy fallbacks.

**2. Worth it? — let me be straight, because a survey moved under me.**

- *Installation is no longer the argument.* slycot 0.7.0 now ships pip wheels across platforms — manylinux/musllinux on x86_64 and aarch64, macOS, Windows. "Hard to install" is essentially solved; I will not pretend otherwise.
- *The honest core is completeness.* Without slycot, generalized-E Lyapunov and discrete Sylvester currently *raise* — so for the many who never install the optional package, these basic operations do not exist. The scipy path makes them work in a stock `pip install control`: the function is there instead of erroring. That is independent of wheels and of license.
- *License is a bounded secondary.* slycot's wrapper is GPLv2 (the underlying SLICOT has been BSD-3-Clause since 2020; only the wrapper relicensing is stuck, on all-author consent). A scipy path gives BSD-clean coverage for GPL-averse distributors — but it is distribution-only and does not change python-control's own BSD license, so I will not overstate it.
- *Precedent:* @murrayrm merged the same move in #1108 (2025), "allow MIMO `ss2tf` without requiring `slycot`."
- *Performance:* compiled SLICOT should win at large n; the scipy path is LAPACK-backed so it is not slow, but I am not claiming it beats SLICOT — happy to benchmark.
- One correction I owe you: a docstring note here overclaims that SLICOT `SG03AD` "also handles singular E." It does not — a truly singular E is out of scope for *both* paths — and I have pushed a commit fixing the wording.

**3. Am I using it.** Not in a production project that calls these *specific* functions today, and I will retract part of my own pitch: I would have cited Fortran-free robotics-grader containers (RoboRacer / F1TENTH on ROS 2), but 0.7.0's
wheels just narrowed that to a niche. The honest drivers are the project's direction (#27 / #36) and completeness — and, further out, a descriptor-system (singular-E) case that slycot itself cannot do, which is where the real new value
lies.

**On the "Claude-generated" concern.** Claude assisted — implementation from one model, review from another, and the commit trailers say so; I am not hiding it. But the motivation, the mathematics, and the direction are mine: I have been through every line — the E ≠ I congruence, the Bartels–Stewart back-substitution and its λ_i·μ_j ≠ 1 solvability condition, the stability argument — and can defend each. I stand behind every line. If a reviewer finds one I cannot account for, that is a bug and I want to hear it.

**On "many more PRs."** It is a finite, bounded effort, sliced small to make review easier, not to flood you — I am glad to pace it, hold the rest until this one is judged, or stop. And knowing the honest and modest case for it, if you would still rather not carry it, I am genuinely happy to withdraw — I do not want to add maintenance you did not ask for. No hard feelings either way.

slivingston (changes requested):
@kwlee2025cpp Thanks for your reply. Some of it is unclear, but I think the only part critical for review is

> Performance: compiled SLICOT should win at large n; the scipy path is
LAPACK-backed so it is not slow, but I am not claiming it beats SLICOT — happy to
benchmark.

Yes, please benchmark. This will also be useful for the other PRs that you plan to make/submit.

kwlee2025cpp:
Thanks @slivingston — benchmark (scipy vs slycot across `n`, with residuals) coming shortly.

In the meantime, so I fix the unclear parts rather than guess at them: which sections read as unclear to you? I will tighten or rewrite them — I know the reply ran long.

kwlee2025cpp:
**Benchmark — `scipy` vs `slycot`.** Apple M-series, macOS, Python 3.13, OpenBLAS, slycot 0.7.0. Each problem is built from a known solution; time is the median over seeds (5 for n ≤ 200, 3 at 400, 1 at 800), ≥ 2–3 runs each. Ratio < 1 ⇒ scipy faster.

### generalized continuous Lyapunov — `lyap(A, Q, E=E)`
| n | scipy (ms) | slycot (ms) | scipy/slycot |
|--:|--:|--:|--:|
| 10 | 0.09 | 0.07 | 1.27 |
| 20 | 0.19 | 0.26 | 0.75 |
| 50 | 1.0 | 4.6 | 0.22 |
| 100 | 5.8 | 22.8 | 0.26 |
| 200 | 20.9 | 161 | 0.13 |
| 400 | 90.7 | 1003 | 0.09 |
| 800 | 797 | 6780 | 0.12 |

### generalized discrete Lyapunov — `dlyap(A, Q, E=E)`
| n | scipy (ms) | slycot (ms) | scipy/slycot |
|--:|--:|--:|--:|
| 10 | 0.15 | 0.09 | 1.72 |
| 20 | 0.24 | 0.32 | 0.75 |
| 50 | 1.6 | 4.5 | 0.36 |
| 100 | 6.7 | 24.8 | 0.27 |
| 200 | 23.5 | 181 | 0.13 |
| 400 | 100 | 1109 | 0.09 |
| 800 | 908 | 7215 | 0.13 |

### discrete Sylvester — `dlyap(A, Q, C)`
| n | scipy (ms) | slycot (ms) | scipy/slycot |
|--:|--:|--:|--:|
| 10 | 0.27 | 0.03 | 9.78 |
| 20 | 0.68 | 0.10 | 6.68 |
| 50 | 8.6 | 1.7 | 5.14 |
| 100 | 45.2 | 10.2 | 4.42 |
| 200 | 116 | 56.9 | 2.05 |
| 400 | 598 | 352 | 1.70 |
| 800 | 5187 | 2728 | 1.90 |

Residuals are at machine precision (~1e-15…1e-14) for both methods on all three routines; scipy and slycot agree to ~1e-14 (≤ 1e-9 at n = 800, reflecting the solution's sensitivity at that size).

**Headline — I was wrong that SLICOT wins at large n; it depends on the routine:**
- **Generalized Lyapunov (continuous and discrete): scipy is faster, ~8–11× at n ≥ 400.** The scipy path reduces via `inv(E)` to a *standard* Lyapunov equation (LAPACK Bartels–Stewart); `SG03AD` solves the *generalized* equation via generalized Schur, which costs more.
- **Discrete Sylvester: slycot is faster, ~1.7–10×** (widest at small n). My Bartels–Stewart Schur-factors both `A` and `Qᵀ` plus a Python column loop; `SB04QD`'s Hessenberg–Schur in Fortran is leaner.

**Accuracy vs cond(E)** — `lyap`, n = 100, median over 5 seeds, err = ‖X̂ − X_true‖ / ‖X_true‖:

| cond(E) | scipy err | slycot err | slycot / scipy |
|--:|--:|--:|--:|
| 1e0 | 7e-15 | 7e-15 | 0.99 |
| 1e2 | 1.1e-13 | 8.5e-14 | 0.80 |
| 1e4 | 2.6e-10 | 3.4e-10 | 1.31 |
| 1e6 | 1.2e-6 | 1.7e-6 | 1.61 |
| 1e8 | 7.6e-3 | 1.5e-2 | 1.95 |
| 1e10 | 67 | 110 | 1.71 |
| 1e12 | 5.2e5 | 9.2e5 | 1.82 |

This one corrected me on my *own* docstring. I had written that `method='slycot'` is "preferable" for ill-conditioned E. It isn't: **both methods degrade together** — the generalized Lyapunov problem is itself ill-conditioned (about cond(E)²) when E is — and slycot is in fact 1.3–2× *less* accurate for cond(E) ≥ 1e4, never better. I've pushed a commit fixing the docstrings and the runtime warning accordingly.

**Net:** the fallback is correct everywhere and as accurate as slycot across the whole conditioning range; for generalized Lyapunov it's also a speedup, while for discrete Sylvester slycot remains the faster path when it's installed. (Happy to add the benchmark script to the PR or a gist if useful.)

slivingston:
@kwlee2025cpp Thanks for comparing performance. In this PR, can you add the code for replicating the experiment? I think the best path is in [benchmarks/](https://github.com/python-control/python-control/tree/main/benchmarks). Give the file a name like `scipy_fallback_bench.py`. In the future (after this PR is merged), we can expand the file for comparing other fallback routines with SLICOT.

kwlee2025cpp:
Done — added `benchmarks/scipy_fallback_bench.py`.

- **`MatrixEquationTiming`** times each `(routine, n, method)` for the three routines that gained `method=` (generalized continuous/discrete Lyapunov and discrete Sylvester). `setup()` builds each problem from a known solution and asserts the chosen method actually solves it (relerr < 1e-6) before timing.
- **`GenLyapAccuracy`** is a `track_` benchmark recording the generalized-Lyapunov relative error vs `cond(E)` — the evidence behind the docstring/warning correction in this PR (both paths degrade about `cond(E)**2`; slycot is not more accurate for ill-conditioned E).
- `method='slycot'` cases raise `NotImplementedError` in `setup` when slycot isn't installed, so the suite runs with or without slycot.

Run with:

```
PYTHONPATH=`pwd` asv run --python=python --bench scipy_fallback
```

or, since these are plain classes, call the `setup`/`time_`/`track_` methods directly to reproduce the numbers without asv.

One note on reproducibility: the tables I posted earlier were medians over several seeds, whereas the committed benchmark fixes a single seed per problem so runs are deterministic and comparable across commits (asv's repeated sampling averages the timing). Same generators, same ratios.

I ran the suite locally with asv (existing-environment mode) and it reproduces both the timing and the accuracy-vs-cond(E) tables. Happy to adjust the size grid or fold in more routines as the fallbacks grow.

slivingston:
Thanks @kwlee2025cpp I will review it tomorrow.

kwlee2025cpp:
Hi @slivingston — no rush at all, just a friendly check-in on this one.

The benchmark is in as `benchmarks/scipy_fallback_bench.py` and reproduces both the timing and the accuracy-vs-`cond(E)` tables.

In case the red CI is what's giving pause: those failures are pre-existing and unrelated to this PR — `test_pole_zero_subplots` and `test_nyquist_basic` fail only on Py3.12 from newer matplotlib/numpy (`main` was last green on 2026-06-09 and hasn't re-run since). The `mateqn` changes here pass across the matrix, with and without slycot.

Happy to rebase or adjust whenever you have a chance — thanks for taking a look.

slivingston:
> Hi @slivingston — no rush at all, just a friendly check-in on this one.

Thanks for checking in. Sorry I have been slow to reply. I will be done soon.

slivingston:
@kwlee2025cpp Almost done with review. Quick question: Can you tell me the relationship between your account (@kwlee2025cpp ) and @kangwonlee ?

I ask because the accounts have similar names and one has forks of the other's repositories.

kwlee2025cpp:
Both accounts are mine.

[@kangwonlee](https://github.com/kangwonlee) is my long-standing personal account, and it owns the fork this PR is branched from. [@kwlee2025cpp](https://github.com/kwlee2025cpp) is the account I use for work where I am getting help from an LLM (Claude). I set it up to be upfront about that, rather than let such work blend into my personal account's history. The commits here are authored under my own name and email, and the pushes use @kwlee2025cpp's SSH key, which has write access to that fork — that is why one account appears to hold forks of the other's repositories.

That is the same disclosure I made earlier in this thread: I use Claude's help, I verify the mathematics myself, and I stand behind every line.

Happy to answer anything further. I would rather not re-open this PR from the other account, since that would lose your review thread, but I can consolidate for future contributions if you prefer.

murrayrm:
FYI: CI failures have been addressed in https://github.com/python-control/python-control/pull/1243. Once that has been merged into main, you should be able to rebase and clear the errors.

sawyerbfuller:
lgtm, after a rebase

kwlee2025cpp:
Rebased onto main, which now includes #1243 — thanks @murrayrm for the CI fix and @sawyerbfuller for the review.

The four commits replay unchanged; the patch is byte-identical to the pre-rebase one. Locally on numpy 2.4.4 / scipy 1.17.1 / matplotlib 3.10.8, `mateqn_test.py` (31 passed), `kwargs_test.py` (66 passed), and both tests that were red before #1243 — `ctrlplot_test.py::test_pole_zero_subplots` and `nyquist_test.py::test_nyquist_basic` — all pass.

The workflow runs on the new head are showing as `action_required`, so they may need an approval click before CI reports.

murrayrm:
@slivingston Do you have any pending comments on this PR?  If not, I'll merge as soon as CI checks clear.

slivingston:
Yes, I will wrap it up tonight and merge.

control/mateqn.py
```diff
@@ -47,15 +47,35 @@ def sb03md(n, C, A, U, dico, job='X', fact='N', trana='N', ldwork=None):
...
+    Lyapunov problem is itself ill-conditioned (about cond(E)**2) when E
+    is, so method='slycot', though it does not form inv(E) explicitly, is
+    not measurably more accurate in that regime.
+    """
+    condE = np.linalg.cond(E)
+    if condE > 1.0 / np.sqrt(finfo(float).eps):
+        warnings.warn(
+            f"E is ill-conditioned (cond(E) = {condE:.2g}); the generalized "
+            "Lyapunov solution may have reduced accuracy.  The problem itself "
+            "is ill-conditioned for such E, so method='slycot' is not "
```
slivingston, line 76: A warning like this has too much scope. The warning can only appear when `method='scipy'`, but it has a claim about `method='slycot'`. If the underlying SLICOT method is improved and this warning message is no longer true, will we remember to update this warning? Probably not because it is only on `method='scipy'` code paths. I recommend that either the claim is removed (i.e., the warning does not refer to `slycot`) or a qualifying word is used, like "method='slycot' may not be more accurate" (so, "may" instead of "is").

control/mateqn.py
```diff
@@ -162,8 +201,25 @@ def lyap(A, Q, C=None, E=None, method=None):
...
+            # inverting E, but a truly singular E is not handled by either
+            # method.
+            try:
+                At = solve(E, A)
+                Qt = solve(E, solve(E, Q).T).T
+            except np.linalg.LinAlgError:
+                raise ControlArgument(
+                    "method='scipy' requires E to be nonsingular; "
+                    "a truly singular E (descriptor system) is not "
+                    "supported by either method")
```
slivingston, line 220: My comment about _warn_ill_conditioned_E() applies to this warning as well. Furthermore, "either method" requires that there are only 2 methods. In the future, there may be more than `scipy` and `slycot` methods, at which time this warning message text would need to change, whether or not a `method='scipy'` code path is affected.

control/mateqn.py
```diff
@@ -279,8 +365,40 @@ def dlyap(A, Q, C=None, E=None, method=None):
...
+            if np.min(np.abs(np.outer(np.diag(Tq), np.diag(Ta)) - 1.)) \
+                    < finfo(float).eps * max(
+                        1., np.abs(np.diag(Ta)).max()
+                        * np.abs(np.diag(Tq)).max()):
+                raise ControlArgument(
+                    "A and Q have a pair of eigenvalues whose product "
+                    "is (almost) equal to 1; the discrete-time "
+                    "Sylvester equation is singular")
+            Y = np.empty((n, m), dtype=complex)
+            TaY = np.empty((n, m), dtype=complex)   # running Ta @ Y
```
slivingston, line 394: What does `#running Ta @ Y` indicate? If it is only a reminder that `TaY = np.empty()` is to allocate the space before filling it in the loop, then delete the comment because it does not help with reading the code. This purpose is obvious, and a similar call is made for `Y` but no such comment is there.

slivingston (approved):
@kwlee2025cpp Thank you very much for careful work. Sorry for my slow pace of review. I only have feedback about warning messages and code comments, which can be handled in a future PR. I will merge now.

Regarding my earlier comment about https://github.com/python-control/python-control/pull/1234#issuecomment-4809264345 having some unclear parts, I read through it again, and only 2 parts are unclear to me:

1. Is @beachdweller one of your previous GitHub accounts? The text "One more disclosure, a personal thread: in 2017..." implies this, but maybe not.

2. What is "robotics-grader" in "I would have cited Fortran-free robotics-grader containers (RoboRacer / F1TENTH on ROS 2), but 0.7.0's wheels just narrowed that to a niche."?

There is nothing wrong with having multiple GitHub accounts. I only asked because I was confused about authorship.

kwlee2025cpp:
Both answered, briefly:

1. Yes — @beachdweller is an earlier account of mine. That 2017 comment on #27 was me, back when I was trying to get SLICOT working through f2py/f2c and did not get far. It predates the @kangwonlee / @kwlee2025cpp split and has nothing to do with it.

2. "robotics-grader" was my own shorthand, sorry — it is not a package. I teach RoboRacer / F1TENTH labs and grade them with autograder containers running ROS 2 on ARM (Jetson-class) hardware, so a Fortran-free dependency chain used to matter for the image build. slycot 0.7.0's wheels largely removed that motivation, which is why I called it a niche rather than a selling point. To be precise: those labs use `ct.dlqr`, so they motivate the general goal rather than this PR's specific functions.

Thank you for the review and for merging. The feedback on warning messages and code comments is welcome — happy to prepare that as a small follow-up whenever suits you.

PR #1239, opened 2026-07-09
Fix typos in comments, docstrings, and error messages

maxtaran2010:
Fix several typos found by codespell in source files.

Changes:
- `control/robust.py`: `continous` -> `continuous` (2 occurrences in TODO comments)
- `control/timeplot.py`: `tranpose` -> `transpose` (comment)
- `control/nlsys.py`: `inpout` -> `input/output` (comment), `connecton` -> `connection` (comment)
- `control/optimal.py`: `covert` -> `convert` (comment)
- `control/phaseplot.py`: `Containter` -> `Container` (docstring)
- `control/statesp.py`: `lamba` -> `lambda` (comment)
- `control/flatsys/bspline.py`: `convertable` -> `convertible` (error message string)

PR #1242, opened 2026-07-25
DOC: follow NumPy policy for contributions that use "AI"

slivingston:
I propose that we follow the ["AI policy" for contributions to NumPy](https://numpy.org/doc/stable/dev/ai_policy.html). Recently we have had several pull requests that I suspect were written by bots that have little or no human management (e.g., https://github.com/python-control/python-control/pull/1239). I believe leveraging tools for code writing is worthwhile; however, we need to establish guidelines to ensure accountability and to avoid wasting time.

I have read through a variety of policies that different projects have established for acceptable use of "AI tools," and I think NumPy's is the most complete and practical, so I propose that we use it.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -177,3 +177,7 @@ Your contributions are welcome!  Simply fork the GitHub repository and send a
 Please see the `Developer's Wiki`_ for detailed instructions.
 
 .. _Developer's Wiki: https://github.com/python-control/python-control/wiki
+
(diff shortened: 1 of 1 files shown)


bnavigator (approved):

sawyerbfuller:
Sounds good to me!

sawyerbfuller (approved):

Issue #1147, opened 2025-04-26
too much white space around Root Locus with omega-damping grid


datacurse:
When I turn on the omega-damping grid for root locus plot with `grid=True` (turned on by default), it adds too much white space around the actual plot, which is very annoying. With this grid turned off, plot is tight and nice.

I would like root locus to not have this extra white space around when grid is turned on, I am pretty sure this is some sort of platform-specific bug. Would appreciate any help 🙏

grid turned on (probably bug):
![Image](https://github.com/user-attachments/assets/989e0653-7216-4e53-b6f0-a45073d50128)

grid turned off (expected behavior):
![Image](https://github.com/user-attachments/assets/517e4d6b-557b-4bb5-bac0-2becda14c949)

Libraries:
- control 0.10.1
- matplotlib 3.10.0

I'm using python control in Google Colab.

slivingston:
I cannot reproduce on the current tip of `main` branch, so this must have been fixed already. There is a new release coming soon: https://github.com/python-control/python-control/issues/1130

murrayrm:
@slivingston Did you run inside a Jupyter notebook?  What version?

I tried running this using the following versions and got the same problem as @datacurse noted:
```
IPython          : 8.31.0
ipykernel        : 6.29.5
ipywidgets       : 8.1.5
jupyter_client   : 8.6.3
jupyter_core     : 5.7.2
jupyter_server   : 2.14.2
jupyterlab       : 4.3.3
nbclient         : 0.10.1
nbconvert        : 7.16.4
nbformat         : 5.10.4
notebook         : 7.3.1
qtconsole        : not installed
traitlets        : 5.14.3
```

I also tried running this on Colab and I get the same problem using the following commands:
```
!pip install git+https://github.com/python-control/python-control.git

import control as ct
import matplotlib.pyplot as plt

sys = ct.tf([1, 2], [1, 0, 0])
cplt = ct.root_locus(sys, grid=True)
```

slivingston:
No, I ran it in IPython in a terminal, so the Matplotlib backend would be different. I just tried in Colab and reproduced it (like you), so now I think it is something new.

roryyorke:
With matplotlib 3.10.0, I can reproduce this.  (Also happens in Colab, at time of writing, and Colab is also running matplotlib 3.10.0)

With matplotlib 3.11.0, there's no extra whitespace, but the damping grid lines look wrong.

Other installed packages at end; apologies for long list, I also installed jupyterlab.

[...]

PR #1250, opened 2026-09-19
Fix s-grid rendering on Matplotlib 3.11 by disabling longitude unwrapping

aduan5567-netizen:
Refs https://github.com/python-control/python-control/issues/1147.

## Rationale

The discussion in #1147 covers two version-dependent symptoms of the
omega-damping (s-plane) grid on root-locus plots:

- With Matplotlib 3.10, enabling the grid adds extra whitespace around
  the plot — the original report. This PR does not change that behavior;
  it is a separate problem and is left untouched here.
- With Matplotlib 3.11, the damping grid lines themselves render
  incorrectly. As diagnosed in the issue discussion, this is a regression
  caused by matplotlib's axisartist refactor, and it is what this PR
  fixes.

`control/grid.py` renders the s-plane damping-ratio / natural-frequency
grid for root-locus and pole-zero diagrams using axisartist's
`GridHelperCurveLinear`. The grid only covers the left half-plane, a fixed
longitude interval of 90°–270°. To keep that interval from being unwrapped
at the 180° boundary, `ModifiedExtremeFinderCycle` subclassed
`ExtremeFinderCycle` and overrode `__call__` to apply a 360° unwrapping
threshold instead of the upstream 180°.

Matplotlib 3.11 moved the sampling/bounding-box logic into a new
`ExtremeFinderSimple._find_transformed_bbox(transform, bbox)` method,
which `GridFinder.get_grid_info()` now calls directly; `__call__` is only
a thin backward-compatible wrapper. Our override is therefore bypassed on
3.11, the upstream 180° unwrapping is applied, and the grid extremes are
computed incorrectly.

Rather than porting the override to the new interface, this PR removes
the custom subclass entirely and uses the stock `ExtremeFinderCycle`:

- Setting `lon_cycle=None` disables the periodic longitude unwrapping
  altogether, so the 90°–270° interval can never be wrapped across the
  180° boundary.
- The existing `lon_minmax=(90, 270)` setting still clips the computed
  grid bbox to the left half-plane, which is what actually enforces the
  fixed interval.

This configuration works on both Matplotlib 3.10 and 3.11, and the 3.10
behavior — including the whitespace reported in #1147 — is unchanged.

## Changes

- Delete `ModifiedExtremeFinderCycle` from `control/grid.py`.
- In `sgrid()`, construct the stock `angle_helper.ExtremeFinderCycle`
  with `lon_cycle=None` (instead of the custom subclass with
  `lon_cycle=360`); `lat_cycle=None`, `lon_minmax=(90, 270)`, and
  `lat_minmax=(0, np.inf)` are unchanged.
- Add `control/tests/grid_test.py::test_sgrid_does_not_unwrap_longitude`,
  which obtains the extreme finder from the axes built by `sgrid()`,
  asserts `finder.lon_cycle is None`, and checks that an identity
  transform over 90°–270.001° returns extremes `(90, 270, 0, 1.05)`.
  The 270.001° edge would trigger the stock 180° unwrapping
  (`270.001 - 90 > 180`), so the test guards against this regression on
  both Matplotlib versions.

The removed class is internal to `control/grid.py`; there are no public
API changes.

## Testing

- Matplotlib 3.11.2: I ran the relevant tests
  - `grid_test.py` and `rlocus_test.py` (grid and root-locus): 80
    passed, 1 skipped;
  - `pzmap_test.py` and `ctrlplot_test.py` (pole-zero and control
    plots): 126 passed, 2 skipped;
  - total: 206 passed, 3 skipped.
- Matplotlib 3.10.8: I ran `grid_test.py` and `rlocus_test.py` — 80
  passed, 1 skipped; the legacy behavior is unchanged.
- I ran `ruff check` and `git diff --check`; both are clean.

## AI-assisted development disclosure

I used AI tools while preparing this patch:

- **OpenAI Codex** drafted the code changes: removing
  `ModifiedExtremeFinderCycle`, switching `sgrid()` to the stock
  `ExtremeFinderCycle` with `lon_cycle=None`, and adding
  `control/tests/grid_test.py`.
- **Doubao (ByteDance's AI assistant)** assisted with problem analysis:
  I used it to trace the Matplotlib 3.11 refactor against the upstream
  3.11.2 source, confirming that `GridFinder.get_grid_info()` now calls
  `_find_transformed_bbox()` directly, which is why the pre-existing
  `__call__` override was bypassed.

I reviewed every AI-assisted change line by line, and I can explain what
the code does and why it is implemented this way — including why
`lon_cycle=None` prevents the 180° unwrapping, how `lon_minmax` keeps the
grid within the left half-plane, and the test's expected
`(90, 270, 0, 1.05)` bounds (the `1.05` comes from the bbox padding with
`ny = 20`, clamped by `lat_minmax`). This PR description was written by
me. I take full responsibility for the correctness of this patch.


slivingston (changes requested):
Is there a solution that does not involve duplicating lots of code?

aduan5567-netizen:
Thanks — I refactored the shared sampling, unwrapping, padding, and clipping logic into `_find_extremes()`. The legacy `__call__` path now only adapts the callable transform and tuple ordering, while the Matplotlib 3.11 path forwards `trans.transform`. This removes the duplicated algorithm while preserving compatibility with both versions. I tested it against Matplotlib 3.10.8 and 3.11.2.

control/grid.py
```diff
@@ -37,9 +37,28 @@ class ModifiedExtremeFinderCycle(angle_helper.ExtremeFinderCycle):
     https://matplotlib.org/_modules/mpl_toolkits/axisartist/angle_helper.html#ExtremeFinderCycle.__call__
     """
     def __call__(self, transform_xy, x1, y1, x2, y2):
-        x, y = np.meshgrid(
-            np.linspace(x1, x2, self.nx), np.linspace(y1, y2, self.ny))
-        lon, lat = transform_xy(np.ravel(x), np.ravel(y))
+        # Matplotlib 3.11 moved this calculation to
+        # _find_transformed_bbox(), which is also called directly by
+        # GridFinder.  Keep the legacy implementation for older releases.
+        if hasattr(super(), '_find_transformed_bbox'):
```
slivingston: Thanks for refactoring. Now that I am investigating this more, I think this approach overall is not the best. My main concerns:

1. This code checks for the presence of an intended-as-hidden superclass method (`_find_transformed_bbox`). Such inspection is bad because `_find_transformed_bbox` is not part of the public API of Matplotlib. (I might be wrong about this; please correct me if so.). Therefore, this check is fragile and could break unexpectedly.
2. Could this be a bug in Matplotlib? I doubt it, but the fact that output changed in a surprising way with the new version of Matplotlib suggests it is worth exploring the Matplotlib side more: check issue trackers, changelogs, etc.
3. Perhaps our design here is wrong and needs to be reconsidered. Is there some way to achieve the same result without making a subclass of `ExtremeFinderCycle` that overrides `__call__` ?

Concerns 2 and 3 require more investigation. Hopefully you can do it :)

aduan5567-netizen: Good point — I agree that overriding `_find_transformed_bbox` and
branching on `hasattr` would lean on a private API that was just
reorganized, and would be fragile against future matplotlib releases.

I went through matplotlib/matplotlib#27551 (the refactor that
introduced `_find_transformed_bbox` and deprecated
`GridFinder.transform_xy`/`inv_transform_xy`), the 3.11 API changes
notes, and the matplotlib issue tracker to check whether there is a
supported hook for this — there isn't one. But stepping back, the
custom subclass's only change was the unwrap threshold (180° → 360°).
Since the s-grid longitude always lies within the fixed 90°–270°
interval, that 360° threshold effectively never unwraps anything, so
it is equivalent to disabling cycle unwrapping altogether.

I've therefore removed `ModifiedExtremeFinderCycle` entirely and now
use the stock `ExtremeFinderCycle` with `lon_cycle=None`, relying on
the existing `lon_minmax=(90, 270)` to clip the grid to the left
half-plane. The new test asserts `lon_cycle is None` and that an
identity transform over 90–270.001° gives extremes
`(90, 270, 0, 1.05)` — the 270.001° edge would trigger the stock 180°
unwrapping, so it guards against this regression.

Behavior on 3.10 is unchanged (including the whitespace issue from
#1147, which is out of scope here). The relevant tests pass on both
versions: matplotlib 3.11.2 — 206 passed, 3 skipped (grid, root-locus,
pole-zero, and control-plot tests); matplotlib 3.10.8 — 80 passed,
1 skipped. `ruff check` and `git diff --check` are clean.

slivingston:
To help with review, can you generate before and after images to demonstrate that behavior is unchanged with Matplotlib 3.10 and fixed with Matplotlib 3.11 ? Please show the code used to generate the images.

(This would be a lot of effort if done manually, but you seem to be using AI tools for most/all of this, so my request seems reasonable :) )

aduan5567-netizen:
## Visual comparison

| | before (`a775479`) | after (`2a140b1`) |
| --- | --- | --- |
| Matplotlib 3.10.8 | mpl-3.10-before | mpl-3.10-after |
| Matplotlib 3.11.2 | mpl-3.11-before | mpl-3.11-after |

I rendered the same root-locus plot (`ct.tf([1, 2], [1, 2, 3])`,
`grid=True`) at `a775479` (before) and `2a140b1` (after) with
Matplotlib 3.10.8 and 3.11.2, using one fixed Agg script (same figure
size, DPI, and rcParams).

- **3.10.8:** the before/after PNGs are byte-for-byte identical — both
  have SHA-256 `28DDCB09C8C09B99EC1481E9416FA3714684658987BC89774E470F3B4E8C7D8B`.
- **3.11.2:** before draws the damping grid only in the upper
  half-plane; after renders it symmetrically, matching 3.10.

The images were generated with:

```python
"""Generate a deterministic root-locus image for PR #1250."""

import argparse

import matplotlib

matplotlib.use("Agg")

import control as ct


parser = argparse.ArgumentParser()
parser.add_argument("output")
args = parser.parse_args()

matplotlib.rcParams.update({
    "figure.dpi": 120,
    "font.size": 10,
    "savefig.dpi": 120,
})

system = ct.tf([1, 2], [1, 2, 3])
ct.root_locus_plot(system, grid=True)

figure = plt.gcf()
figure.set_size_inches(6.4, 4.8)
figure.suptitle(f"Matplotlib {matplotlib.__version__}", fontsize=11)
figure.savefig(args.output)
plt.close(figure)
```

Run `python generate_comparison.py OUTPUT.png` in each checkout.

